PR #4, opened 2023-01-20
feat: Add edge-token extractor to lock down access

chriswk:
## What
This adds EdgeToken as a type. It also adds a way to go from an HttpRequest to an EdgeToken by implementing Actix's trait FromRequest for an EdgeToken.

If you have an endpoint where you need authentication, you simply say that you want an EdgeToken and the type system will guarantee you either get an EdgeToken or you'll get a 403 error.

## Thoughts
We now have two traits, one for resolving features, and one for verifying tokens, or converting from an authorization header to a token. These are merged to a trait I've called EdgeProvider, and we now wire in an offline implementation of EdgeProvider (give bootstrap file, and a list of accepted tokens that will have read-access to the bootstrap file and it's results).

I tried to make our methods only require the traits they use, but the wiring in Actix from web::Data to method parameters wasn't able to see that something that implemented the Super trait would satisfy the Subtrait.

At least not the way I did it. Something to dig into next week.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1761,6 +1761,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "test-case"
+version = "2.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "21d6cf5a7dffb3f9dceec8e6b8ca528d9bd71d36c9f074defb548ce161f598c0"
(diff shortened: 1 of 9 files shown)


sighphyre (approved):
LGTM to me, Clippy still seems to be a bit cranky tho

PR #9, opened 2023-01-24
feat: Added cors middleware

chriswk:
Since we know we'll probably be used from browsers, setting the CORS headers are a must.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -2,7 +2,8 @@ use std::sync::Arc;
 
 use crate::cli::EdgeMode;
 use crate::offline_provider::OfflineProvider;
-use actix_web::{middleware, web, App, HttpServer};
+use actix_cors::Cors;
+use actix_web::{http, middleware, web, App, HttpServer};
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -34,8 +35,15 @@ async fn main() -> Result<(), anyhow::Error> {
...
         let client_provider_data = web::Data::from(client_provider_arc);
+
+        let cors_middleware = Cors::default()
+            .allowed_origin("*")
+            .allowed_methods(vec!["GET", "POST"])
+            .allowed_headers(vec![http::header::AUTHORIZATION, http::header::ACCEPT])
+            .allowed_header(http::header::CONTENT_TYPE);
         App::new()
             .app_data(client_provider_data)
+            .wrap(cors_middleware)
```
sighphyre, line 46: Apply the middleware to everything? Do we want that?

sighphyre: ...it's probably okay

PR #10, opened 2023-01-24
feat: use subcommands rather than valueenum

chriswk:
### What
This code moves from value enum to subcommands, keeping arguments for http setup, but allowing you to only have to worry about configuring the mode you want to launch edge in.

### Worth mentioning
The value here is harder to see, since we still only have one subcommand. Judging from our experience in the node version here, where we ended up with a rather complicated yaml config, this is pre-optimising a bit to ease the extension into more modes later.

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -1,10 +1,18 @@
 use std::path::PathBuf;
 
-use clap::{Args, Parser};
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Awesome, this is a really nice improvement

PR #11, opened 2023-01-24
docs: Flesh out README with concepts (mode)

chriswk:
To make sure we don't end up having the boring task of having to document everything just before release, I thought I'd get ahead of the curve for once and start documenting what we're thinking about modes at once. I expect the chapter for the Edge mode to become rather long, since that's the most complex mode we have. But for now, this at least gets us going, explaining the offline-mode.

I tried to find the OpenAPI docs for proxy/frontend, but I can't find it, so for now the link to the `/api/proxy | /api/frontend` endpoints will have to be void.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,59 @@
 
 Unleash Edge is the successor to the Unleash Proxy.
 
+## Running Unleash Edge
+
+Unleash edge is compiled to a single binary. We use clap to parse arguments/environment variables to configure Unleash Edge. All modes [See Concepts/Modes](#modes) share http configuration variables.
+
+```shell
+$ ./unleash-edge --help
+Options:
+  -p, --port <PORT>
+          Which port should this server listen for HTTP traffic on [env: PORT=] [default: 3063]
+  -i, --interface <INTERFACE>
+          Which interfaces should this server listen for HTTP traffic on [env: INTERFACE=] [default: 0.0.0.0]
+      --tls-enable
+          Should we bind TLS [env: TLS_ENABLE=]
+      --tls-server-key <TLS_SERVER_KEY>
+          Server key to use for TLS [env: TLS_SERVER_KEY=]
+      --tls-server-cert <TLS_SERVER_CERT>
+          Server Cert to use for TLS [env: TLS_SERVER_CERT=]
+      --tls-server-port <TLS_SERVER_PORT>
+          Port to listen for https connection on (will use the interfaces already defined) [env: TLS_SERVER_PORT=] [default: 3043]
+```
+
+## Concepts
+
+### Modes
+
+We support running in various modes, from a [local version](#offline) to a full blown [edge mode](#edge) supporting dynamic keys, metrics
+
+#### Offline
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,59 @@
 
 Unleash Edge is the successor to the Unleash Proxy.
 
+## Running Unleash Edge
+
+Unleash edge is compiled to a single binary. We use clap to parse arguments/environment variables to configure Unleash Edge. All modes [See Concepts/Modes](#modes) share http configuration variables.
```
sighphyre, line 7:
```suggestion
Unleash edge is compiled to a single binary. We use Clap to parse arguments/environment variables to configure Unleash Edge. All modes [See Concepts/Modes](#modes) share http configuration variables.
```

README.md
```diff
@@ -2,6 +2,59 @@
...
+          Server Cert to use for TLS [env: TLS_SERVER_CERT=]
+      --tls-server-port <TLS_SERVER_PORT>
+          Port to listen for https connection on (will use the interfaces already defined) [env: TLS_SERVER_PORT=] [default: 3043]
+```
+
+## Concepts
+
+### Modes
+
+We support running in various modes, from a [local version](#offline) to a full blown [edge mode](#edge) supporting dynamic keys, metrics
```
sighphyre, line 30:
```suggestion
We support running in various modes, from a [local version](#offline) to a full blown [edge mode](#edge) supporting dynamic keys, metrics.
```

README.md
```diff
@@ -2,6 +2,59 @@
...
+
+#### Offline
+
+You have a need to have full control of both the data your clients will get and which keys can be used to access the server. This mode needs a downloaded JSON dump of a result from a query against an Unleash server on the [/api/client/features](https://docs.getunleash.io/reference/api/unleash/get-client-feature) endpoint as well as a comma-separated list of keys that should be allowed to access the server.
+
+If your keys follow the Unleash API key format `[project]:[environment].<somesecret>`, Edge will filter the features dump to match the project contained in the key. 
+
+If you'd rather use a simple key like `secret-123`, any query against `/api/client/features` will receive the dump passed in on the command line. 
+
+Any query against `/api/frontend` or `/api/proxy` with a valid key will receive only enabled
```
sighphyre, line 40: Did you mean to have "toggles" here
```suggestion
Any query against `/api/frontend` or `/api/proxy` with a valid key will receive only enabled  toggles.
```

nunogois (approved):

README.md
```diff
@@ -2,6 +2,59 @@
 
 Unleash Edge is the successor to the Unleash Proxy.
 
+## Running Unleash Edge
+
+Unleash edge is compiled to a single binary. We use Clap to parse arguments/environment variables to configure Unleash Edge. All modes [See Concepts/Modes](#modes) share http configuration variables.
```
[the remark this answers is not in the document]
sighphyre: Yeah good point actually. I don't think there's a good reason, we're just excited. We can take this out in the next iteration

PR #12, opened 2023-01-24
feat: add basic proxy endpoints and related test code

sighphyre:
Adds a basic implementation of front end endpoints. The core purpose of this PR is to consume Yggdrasil and set a path forward for integration testing

### DIFF
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -83,30 +83,34 @@ mod tests {
     use actix_web::{
         http::header::ContentType,
         test,
-        web::{self, Bytes},
+        web::{self, Data},
         App,
     };
     use serde_json::json;
     use unleash_types::{
-        client_features::{ClientFeature, ClientFeatures, Variant},
+        client_features::{ClientFeature, ClientFeatures, Constraint, Operator, Strategy},
         frontend::{EvaluatedToggle, EvaluatedVariant, FrontendResult},
     };
 
-    #[derive(Clone)]
-    struct MockDataSource {}
+    #[derive(Clone, Default)]
+    struct MockDataSource {
+        features: Option<ClientFeatures>,
+    }
+
+    impl MockDataSource {
+        fn with(self, features: ClientFeatures) -> Self {
+            return MockDataSource {
+                features: Some(features),
+            };
+        }
+    }
 
     impl FeaturesProvider for MockDataSource {
         fn get_client_features(&self, _token: crate::types::EdgeToken) -> ClientFeatures {
-            return ClientFeatures {
-                version: 1,
-                features: vec![ClientFeature {
-                    name: "test".into(),
-                    enabled: true,
-                    ..ClientFeature::default()
-                }],
-                segments: None,
-                query: None,
-            };
+            self.features
+                .as_ref()
+                .expect("You need to populate the mock data for your test")
+                .clone()
         }
     }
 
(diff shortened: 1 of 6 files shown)


server/src/frontend_api.rs
```diff
@@ -1,3 +1,250 @@
...
+    let context = context.into_inner();
+
+    let evaluated_features = resolve_frontend_features(client_features, context);
+
+    Ok(Json(FrontendResult {
+        toggles: evaluated_features,
+    }))
+}
+
+#[actix_web::post("/proxy/all")]
```
nunogois, line 29: I'm assuming we can `use` post above instead of simplify this by using `post` only, which would also be more consistent with the `get` in the route above.

server/src/frontend_api.rs
```diff
@@ -1,3 +1,250 @@
...
+    #[derive(Clone, Default)]
+    struct MockDataSource {
+        features: Option<ClientFeatures>,
+    }
+
+    impl MockDataSource {
+        fn with(self, features: ClientFeatures) -> Self {
+            return MockDataSource {
+                features: Some(features),
+            };
```
nunogois, line 104: Just curious, can this be:

```rs
MockDataSource {
    features: Some(features),
}
```

instead?

server/src/frontend_api.rs
```diff
@@ -1,3 +1,169 @@
...
+                }],
+                segments: None,
+                query: None,
+            };
+        }
+    }
+
+    impl TokenProvider for MockDataSource {
+        fn get_known_tokens(&self) -> Vec<crate::types::EdgeToken> {
+            todo!()
```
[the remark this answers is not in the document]
nunogois: So, sort of like implementing an interface on our mocks on Unleash I assume. We satisfy the TS compiler by implementing the methods but they can just do w/e because it doesn't matter.

server/src/frontend_api.rs
```diff
@@ -1,3 +1,250 @@
...
+        }
+    }
+
+    impl EdgeProvider for MockDataSource {}
+
+    impl Into<Data<dyn EdgeProvider>> for MockDataSource {
+        fn into(self) -> Data<dyn EdgeProvider> {
+            let client_provider_arc: Arc<dyn EdgeProvider> = Arc::new(self.clone());
+            let provider = Data::from(client_provider_arc);
+            provider
```
nunogois, line 137: Could this also be something like `Data::from(client_provider_arc)` and return it right away instead of allocating to the variable `provider`?

server/src/frontend_api.rs
```diff
@@ -1,3 +1,250 @@
...
+                        context_name: "userId".into(),
+                        operator: Operator::In,
+                        case_insensitive: false,
+                        inverted: false,
+                        values: Some(vec!["7".into()]),
+                        value: None,
+                    }]),
+                    parameters: None,
+                }]),
+                ..ClientFeature::default()
```
nunogois, line 161: I'm assuming this is a spread in Rust?

nunogois (approved):
🦀

server/src/frontend_api.rs
sighphyre: Yep!

server/src/frontend_api.rs
sighphyre: Yeah, nice catch, updated

server/src/frontend_api.rs
sighphyre: Yeah, that's left over from me fighting with Actix to get it to respect the underlying type. Updated

server/src/frontend_api.rs
sighphyre: Yes. Yes it can. Apparently I'm still in Golang mode. Surprised Clippy didn't get cranky. Nice catch, updated

server/src/frontend_api.rs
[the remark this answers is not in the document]
sighphyre: Yeah, sort of. This is Actix playing silly buggers here, it can't dynamically resolve the types correctly so we gotta nudge it a little. Sadly not a compile error, this results in a runtime error since Actix doesn't know how to actually call the relevant functions

PR #14, opened 2023-01-25
fix: patch the way CORS headers are done

sighphyre:
Seems to be a bug in the Cors crate, according to this: https://stackoverflow.com/questions/74867363/actix-cannot-start-server-service-0

This makes the server run without crashing


chriswk (approved):
Thanks

PR #15, opened 2023-01-25
feat: internal backstage build info endpoint

chriswk:
### What
This adds build info (hash, rust, os, etc) as a new `/internal-backstage/info` endpoint, allowing us to see easy which version is deployed where.

### Next step
Add default labels on prometheus metrics for version.


chriswk:
Having talked a bit with Simon, and a bit with myself. I'll add the awc client in a separate PR. I'll force push this to not include the client stuff.

gastonfournier (approved):
Nice!

nunogois (approved):

PR #16, opened 2023-01-25
feat: add client for getting features

chriswk:
Building a client to get upstream data and post metrics upstream


sighphyre (approved):
LVGTM

PR #18, opened 2023-01-26
Feat/add enabled toggles routes

nunogois:
Adds routes for enabled-filtered toggles.
Fixes usage of `prometheus::process_collector` to be exclusive to "linux".


sighphyre (approved):
LGTM

I think Chris should take a look too, I'm too close

chriswk (approved):
I like it :)

PR #19, opened 2023-01-27
task: Build minimal docker container from scratch base

chriswk:
## What
This PR updates our Dockerfile to take advantage of our musl build, and just use scratch as BASE, which means that the only thing that exists in the container is our executable. Smaller image, smaller attack surface. :)


sighphyre (approved):
I love it

PR #20, opened 2023-01-27
task: Added prometheus metrics from shadow

chriswk:
Building on our BuildInfo endpoint, this adds a couple of nice labels to all our prometheus metrics, allowing us to filter on versions of edge what's working/not working.


sighphyre (approved):
LGTM

PR #25, opened 2023-01-31
Implement redis datasource

sighphyre:
Refactors the way we create and handle data sources i.e. we pass the parent trait around rather than passing the struct. On the backbone of that, this adds a Redis datasource and introduces test containers so that we can get useful tests around data source implementations

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -22,6 +22,7 @@ dotenv = { version = "0.15.0", features = ["clap"] }
 opentelemetry = { version = "0.18.0", features = ["trace", "rt-tokio", "metrics"] }
 opentelemetry-prometheus = "0.11.0"
 prometheus = { version = "0.13.3", features = ["process"] }
+redis = "0.22.3"
 rustls = "0.20.8"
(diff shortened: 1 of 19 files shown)


server/Cargo.toml
```diff
@@ -22,6 +22,7 @@ dotenv = { version = "0.15.0", features = ["clap"] }
 opentelemetry = { version = "0.18.0", features = ["trace", "rt-tokio", "metrics"] }
 opentelemetry-prometheus = "0.11.0"
 prometheus = { version = "0.13.3", features = ["process"] }
+redis = "0.22.3"
```
sighphyre, line 25: We may want to lock this behind a compiler flag in future so we can produce smaller binaries. Small thing

server/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub mod data_sources;
```
sighphyre, line 1: Expose the internals so we can add integration tests in a separate module

chriswk (approved):
Nice. I like the move into lib of our module declarations as well :)

PR #26, opened 2023-01-31
docs: update README

nunogois:
Small README fixes and update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,16 @@
 # Unleash Edge
 
-Unleash Edge is the successor to the Unleash Proxy.
+Unleash Edge is the successor to the [Unleash Proxy](https://docs.getunleash.io/how-to/how-to-run-the-unleash-proxy).
 
 ## Running Unleash Edge
 
-Unleash edge is compiled to a single binary. We use Clap to parse arguments/environment variables to configure Unleash Edge. All modes [See Concepts/Modes](#modes) share http configuration variables.
+Unleash Edge is compiled to a single binary. You can configure it by passing in arguments or setting environment variables.
 
 ```shell
 $ ./unleash-edge --help
+Commands:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,11 +26,13 @@ Options:
           Port to listen for https connection on (will use the interfaces already defined) [env: TLS_SERVER_PORT=] [default: 3043]
 ```
 
+All supported [modes](#modes) share common configuration variables.
```
sighphyre, line 29: I'm not sure what this means here. Different modes will definitely have different variables, some modes will share some stuff

server/src/cli.rs
```diff
@@ -4,7 +4,9 @@ use clap::{ArgGroup, Args, Parser, Subcommand};
 
 #[derive(Subcommand, Debug, Clone)]
 pub enum EdgeMode {
+    /// Run in offline mode
```
sighphyre, line 7: I don't think these comments add much, they seem to be describing what the enum variant does. If the enum is unclear, I think we should rather just rename it

sighphyre (commented):
Looks good! Left some points of discussion, feel free to tell me I'm wrong

server/src/cli.rs
sighphyre: Fair enough!

README.md
sighphyre: Oooohhh. Yeah sure, the original is not a very clear description to me. I think you're right - let's remove it and put it back in when we actually have that section stabilised. Also okay with just leaving it. Gonna slap an approve on this and leave it up to you

sighphyre (approved):
G

chriswk (approved):
LGTM as well. Simon brought up some valid points, but I see you resolved those. More eyes on documentation is always good :)

PR #30, opened 2023-02-01
feat: implement simplify tokens

gastonfournier:
## About the changes
This introduces a simplify function that reduces a set of tokens to a minimum set that still pulls the same configuration from the server

## Benchmarks
```
Running benches/benchmark.rs (target/release/deps/benchmark-d1d3ad66b6852229)
simplify_bench          time:   [1.0222 µs 1.0542 µs 1.0856 µs]
```


server/src/types.rs
```diff
@@ -50,6 +50,14 @@ impl EdgeToken {
             alias: None,
         }
     }
+
+    pub fn subsumes(&self, other: &EdgeToken) -> bool {
+        return self.token_type == other.token_type
+            && self.environment == other.environment
+            && (self.projects.contains(&"*".into())
+                || (self.projects.len() >= other.projects.len()
+                    && other.projects.iter().all(|p| self.projects.contains(p))));
```
gastonfournier, line 59: I hate doing 2 iterations on projects, especially because the length check was supposed to prevent going through the list in some situations

server/src/tokens.rs
```diff
@@ -0,0 +1,99 @@
+use crate::types::EdgeToken;
+
+pub fn simplify(tokens: &[EdgeToken]) -> Vec<EdgeToken> {
```
sighphyre, line 3: Can we make this pub(crate) instead?

server/src/tokens.rs
```diff
@@ -0,0 +1,99 @@
...
+            environment: env.map(|env| env.into()),
+            projects: projects.into_iter().map(|p| p.into()).collect(),
+            expires_at: None,
+            seen_at: None,
+            alias: None,
+        }
+    }
+
+    #[test]
+    fn test_case_1() {
```
sighphyre, line 40: Better name maybe? I would have no idea why this failed at a glance

server/src/types.rs
```diff
@@ -220,7 +228,26 @@ mod tests {
...
+    use crate::types::{EdgeToken, TokenType};
+
+    fn test_str(token: &str) -> EdgeToken {
+        EdgeToken::from_str(
+            &(token.to_owned() + ".614a75cf68bef8703aa1bd8304938a81ec871f86ea40c975468eabd6"),
+        )
+        .unwrap()
+    }
+
+    fn test_token(env: Option<&str>, projects: Vec<&str>) -> EdgeToken {
```
sighphyre, line 240: So typically what I would do here is derive default on the underlying struct, and then you can use destructuring to populate the fields you want:

```
EdgeToken{
  token_type: Some(TokenType::Client),
  ..EdgeToken::default()
}
```

server/src/tokens.rs
gastonfournier: Tryied to describe them better but I wanted to use `#[test_case]` although I couldn't find a way, so the index indicates we should refactor them later

sighphyre (approved):
LGTM

PR #32, opened 2023-02-02
feat: added workflow for uploading binaries to release

chriswk:
## What
As part of being something we'd like our users to easily be able to deploy to their infrastructure, we've discussed making binaries for the major platforms. This PR adds creation and uploading of a manifest file as well as x86_64, aarch64 and windows builds to a release tag.


nunogois (approved):
LGTM

PR #37, opened 2023-02-03
fix: allow any on CORS

nunogois:
Allows everything on the CORS middleware so we can test Edge with a frontend SDK for now.


chriswk (approved):
LGTM

PR #38, opened 2023-02-03
fix: features get refreshed.

chriswk:
Previously our spin loop slept for 15 seconds and then hit the await on the channel for a new token to validate.
This PR changes that to use tokio::select to either refresh features for known tokens each 10th second, or receive a new token to validate.

Should allow us to use more than one token and get them refreshed


gastonfournier (approved):
LG!

PR #40, opened 2023-02-03
feat: switch to backing with HashMap<token, EdgeToken>

chriswk:
There's one test failing here, because I haven't had time to fix the redisprovider defaulting to unknown for validationstatus (seems correct, we just need to mock that it is validated)


sighphyre (approved):
Looks good

PR #44, opened 2023-02-06
task: Updated to only refresh tokens of type Client

chriswk:
having thought about this over the weekend this PR includes two optimizations. One for sending on the validated_tokens from the validate token channel. One for filtering which tokens the feature_refresh_channel keeps on TokenType. Should prevent us from receiving 403's from Upstream.

Also pointed out to me that our auth middleware also needs to check the tokentype before allowing/disallowing the call.

### DIFF
--- a/server/src/http/background_refresh.rs
+++ b/server/src/http/background_refresh.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashSet, sync::Arc, time::Duration};
 
-use crate::types::{ClientFeaturesResponse, EdgeSink, EdgeToken};
+use crate::types::{ClientFeaturesResponse, EdgeSink, EdgeToken, TokenType};
 use tokio::sync::{mpsc::Receiver, mpsc::Sender, RwLock};
 use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
G

server/src/http/background_refresh.rs
```diff
@@ -15,11 +15,13 @@ pub async fn poll_for_token_status(
...
             match write_lock.validate(vec![token.clone()]).await {
                 Ok(validated_tokens) => {
-                    let sink_result = write_lock.sink_tokens(validated_tokens).await;
+                    let sink_result = write_lock.sink_tokens(validated_tokens.clone()).await;
                     if let Err(err) = sink_result {
                         warn!("Couldn't sink token result: {err:?}")
                     } else {
-                        let _ = feature_channel.send(token).await;
+                        for valid in validated_tokens {
+                            let _ = feature_channel.send(valid).await;
```
nunogois, line 23: Would it be better to only send in filtered "client" tokens to the channel instead of sending in everything and filtering on the `refresh_features` method?

sighphyre: Occurred to me as well. I suspect we're going to tap into this from more than just this place, though, so it's nice to be defensive. Probably something for a future PR when the structure is more settled here if we only call this from the one place

PR #46, opened 2023-02-07
chore: redesign source/sink architecture

sighphyre:
Reworks the source/sink architecture to be truly independent of each other. This means that stores for offline mode no longer need to implement the sink API, which makes no sense in that context


chriswk (approved):
LGTM

PR #49, opened 2023-02-08
task: implement client metrics

nunogois:
My naïve implementation of client metrics with a flatter structure - Instead of storing buckets in the metrics cache and converting back and forth, we store them as ClientMetricsEnv. Also uses a `MetricsKey` as that hash key, based on `app_name` and `toggle_name` to identify each `ClientMetricsEnv`.


sighphyre (approved):
LGTM just want to add some validation here - TO THE TEST MOBILE

PR #51, opened 2023-02-08
task: client metrics start task

nunogois:
Add `send_metrics_task` to our running tasks. It accepts a new arg: `metrics_interval_seconds`.

Rename `get_valid_tokens` to `filter_valid_tokens` to better reflect its behaviour. Add `get_valid_tokens` to return all valid tokens from our store.

Implement `background_send_metrics`.

### DIFF
--- a/server/src/data_sources/offline_provider.rs
+++ b/server/src/data_sources/offline_provider.rs
@@ -30,6 +30,15 @@ impl TokenSource for OfflineProvider {
         Ok(self.valid_tokens.values().cloned().collect())
     }
 
+    async fn get_valid_tokens(&self) -> EdgeResult<Vec<EdgeToken>> {
+        Ok(self
+            .valid_tokens
+            .values()
+            .filter(|t| t.status == TokenValidationStatus::Validated)
+            .cloned()
+            .collect())
+    }
+
     async fn get_token_validation_status(&self, secret: &str) -> EdgeResult<TokenValidationStatus> {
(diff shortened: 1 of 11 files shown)


server/src/data_sources/offline_provider.rs
```diff
@@ -30,6 +30,15 @@ impl TokenSource for OfflineProvider {
         Ok(self.valid_tokens.values().cloned().collect())
     }
 
+    async fn get_valid_tokens(&self) -> EdgeResult<Vec<EdgeToken>> {
```
sighphyre, line 33: Idle thoughts but it looks like we might end up writing this same method over and over. Might be worth implementing this as a trait on `Vec<EdgeToken>`. Not something for this PR

server/src/http/background_send_metrics.rs
```diff
@@ -1,34 +1,62 @@
...
+                    warn!("Error sending metrics: {e:?}")
+                }
             }
         }
 
         tokio::time::sleep(Duration::from_secs(send_interval)).await;
     }
 }
+
+async fn get_first_token(source: Arc<RwLock<dyn EdgeSource>>) -> EdgeResult<EdgeToken> {
```
sighphyre, line 51: Also idle thoughts, but if this feels like it wants to be an option not a result. Doesn't need to be done in this PR

sighphyre (approved):
Nice, this looks really clean!

PR #53, opened 2023-02-09
Task client metrics

sighphyre:
Implements client metrics endpoints for both standard SDK requests and bulk metrics requests from another Edge instance. Edge will currently squash metrics down to hour long bins and send them to the upstream server on a lazy schedule in the background


application.json
```diff
@@ -0,0 +1,9 @@
+{
+	"appName":"myapp",
+	"instanceId":"isbetterthanyourapp",
```
nunogois, line 3: 😎

nunogois (approved):
LGTM

PR #57, opened 2023-02-10
chore: redesign source and sinks to store features

sighphyre:
This redesigns the way we store data in the memory and Redis providers to store incoming features by environment rather the token string. This means that incoming features for an environment will be merged in the case that the come from the same environment but different tokens. To handle this in the sources, outgoing data is filtered by project

### DIFF
--- a/server/src/data_sources/mod.rs
+++ b/server/src/data_sources/mod.rs
@@ -1,4 +1,28 @@
+use unleash_types::client_features::ClientFeature;
+
+use crate::types::EdgeToken;
+
 pub mod builder;
 pub mod memory_provider;
 pub mod offline_provider;
 pub mod redis_provider;
+
+trait ProjectFilter<T> {
+    fn filter_by_projects(&self, token: &EdgeToken) -> Vec<T>;
+}
+
+impl ProjectFilter<ClientFeature> for Vec<ClientFeature> {
+    fn filter_by_projects(&self, token: &EdgeToken) -> Vec<ClientFeature> {
+        self.iter()
+            .filter(|feature| {
+                if let Some(feature_project) = &feature.project {
+                    token.projects.contains(&"*".to_string())
+                        || token.projects.contains(feature_project)
+                } else {
+                    false
+                }
+            })
+            .cloned()
+            .collect::<Vec<ClientFeature>>()
+    }
+}
--- a/server/src/data_sources/redis_provider.rs
+++ b/server/src/data_sources/redis_provider.rs
@@ -1,17 +1,19 @@
 use async_trait::async_trait;
+use redis::AsyncCommands;
 use redis::{Client, Commands, RedisError};
 use tokio::sync::{mpsc::Sender, RwLock};
-use unleash_types::client_features::ClientFeatures;
+use unleash_types::client_features::{ClientFeature, ClientFeatures};
+use unleash_types::Merge;
 
-pub const FEATURE_PREFIX: &str = "unleash-edge-feature-";
-pub const TOKENS_KEY: &str = "unleash-edge-tokens";
+pub const FEATURE_PREFIX: &str = "unleash-feature-namespace:";
+pub const TOKENS_KEY: &str = "unleash-token-namespace:";
 
 use crate::types::TokenValidationStatus;
 use crate::{
     error::EdgeError,
     types::{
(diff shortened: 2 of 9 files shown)


nunogois (approved):
LGTM

server/src/data_sources/redis_provider.rs
```diff
@@ -38,8 +40,12 @@ impl RedisProvider {
...
 impl EdgeSink for RedisProvider {}
 
-fn build_features_key(token: &String) -> String {
-    format!("{FEATURE_PREFIX}{token}")
+fn build_features_key(token: &EdgeToken) -> String {
+    token
+        .environment
+        .as_ref()
+        .map(|environment| format!("{FEATURE_PREFIX}{environment}"))
+        .expect("Tying to resolve features for a token that hasn't been validated")
```
gastonfournier, line 48:
```suggestion
        .expect("Resolving features for a validated token")
```
Just words, but when I read the `expect` (might be because I'm unfamiliar with Rust yet), it seems that you're expecting a non validated token...

server/src/data_sources/redis_provider.rs
```diff
@@ -50,14 +56,21 @@ impl FeatureSink for RedisProvider {
...
-        let serialized_features = serde_json::to_string(&features)?;
-        let _: () = lock.set(build_features_key(&token.token), serialized_features)?;
+        let mut con = lock.get_async_connection().await?;
+
+        let key = build_features_key(token);
+
+        let features_to_store =
+            if let Some(stored_features) = con.get::<&str, Option<String>>(key.as_str()).await? {
+                let stored_features = serde_json::from_str::<ClientFeatures>(&stored_features)?;
+                stored_features.merge(features.clone())
```
gastonfournier, line 66: Maybe a problem for later, but this could be an always-growing collection. We won't remove data and the TTL might get updated so stored_features will never expire...

server/src/data_sources/mod.rs
```diff
@@ -1,4 +1,28 @@
...
+        self.iter()
+            .filter(|feature| {
+                if let Some(feature_project) = &feature.project {
+                    token.projects.contains(&"*".to_string())
+                        || token.projects.contains(feature_project)
+                } else {
+                    false
+                }
+            })
+            .cloned()
```
gastonfournier, line 25: Not a big deal, but this code is pretty similar to the one in server/src/data_sources/redis_provider.rs, maybe something to extract

server/tests/redis_test.rs
```diff
@@ -1,54 +1,242 @@
...
+}
 
-//     let content = fs::read_to_string("../examples/features.json").expect("Could not read file");
+fn build_features_key(token: &EdgeToken) -> String {
+    token
+        .environment
+        .as_ref()
+        .map(|environment| format!("{FEATURE_PREFIX}{environment}"))
+        .expect("Tying to resolve features for a token that hasn't been validated")
+}
```
gastonfournier, line 29: Should we use the function from redis_provider here? Maybe it's a thing about function scopes and visibility that makes it difficult to do that

gastonfournier (approved):
Looks fine, just some comments

server/src/data_sources/redis_provider.rs
sighphyre: Yeah, that's fair. Expect should loosely read as "I expected this value and if I didn't get it, crash with the enclosed error message"

server/src/data_sources/redis_provider.rs
sighphyre: That's a good point! Yes, we should fix this but probably in a different PR

server/src/data_sources/mod.rs
sighphyre: Yeah, basically the same code. It'll be the same code for every source. Nice target for a trait impl in future

server/tests/redis_test.rs
sighphyre: I intentionally didn't do that because I don't want to expose that function. I figure it's okay because if you'll break these tests hard if the backing implementation changes. Don't mind a bit of duplication in tests to preserve sanity in the public API

server/src/data_sources/redis_provider.rs
nunogois: I also agree that `expect` is a bit confusing readability wise. It's almost like it should be a `except` instead.

PR #58, opened 2023-02-10
task: token validator

nunogois:
- Adds the token validator concept so we can properly handle token validation wherever we need to;
- Fetches data for validated tokens;
- Allows setting client keys through arg so we can pre-warm;
- Misc fixes and other improvements and refactors;


chriswk (approved):
Looks good. Closer and closer to releasable now. 😃

PR #59, opened 2023-02-11
task: Subsume keys to check

chriswk:
This collapses the keys seen. Removing keys that have been subsumed by a wider key (a key that includes the same projects or more as existing keys).

### DIFF
--- a/server/src/data_sources/memory_provider.rs
+++ b/server/src/data_sources/memory_provider.rs
@@ -51,11 +51,22 @@ impl MemoryProvider {
             self.token_store.insert(token.token.clone(), token.clone());
             if token.token_type == Some(crate::types::TokenType::Client) {
                 self.tokens_to_refresh
-                    .insert(token.token.clone(), FeatureRefresh::new(token));
+                    .entry(token.clone().token)
+                    .or_insert(FeatureRefresh::new(token.clone()));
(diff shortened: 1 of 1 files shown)


chriswk:
One thing. Quite a bit of this logic seems like it's the same no matter where we store things, so if anyone wants to have a once-over on our abstraction layer I think that could simplify adding new datasources.

server/src/data_sources/memory_provider.rs
```diff
@@ -51,11 +51,22 @@ impl MemoryProvider {
...
         }
+        self.reduce_tokens_to_refresh();
+    }
+    fn reduce_tokens_to_refresh(&mut self) {
+        let tokens: Vec<EdgeToken> = self
+            .tokens_to_refresh
+            .values()
+            .map(|r| r.token.clone())
+            .collect();
+        let minimized = crate::tokens::simplify(&tokens);
```
gastonfournier, line 66: I think it's time to remove:
https://github.com/Unleash/unleash-edge/blob/ca0a50d711f8c504f2ad9671929abc663639264b/server/src/tokens.rs#L3

gastonfournier (approved):
Looks good!

sighphyre (approved):
LG

PR #60, opened 2023-02-13
docs: add edge mode

nunogois:
- Adds documentation for **Edge mode**;
- Updates some of our existing documentation;
- Uses the term `tokens` instead of `keys` for simplification and consistency purposes;


README.md
```diff
@@ -30,48 +30,91 @@ Options:
...
+  C-->|Fetch features| E((Edge 3))
+  D-->|Fetch features| E
+  E-->|Fetch features| F((Unleash))
+```
+
+This means that, in order to start up, Edge mode needs to know where the upstream node is. This is done by passing the `--unleash-url` command line argument or setting the `UNLEASH_URL` environment variable.
+
+By default, Edge mode uses in-memory to store the features it fetches from the upstream node. However, you may want to use a more persistent storage solution. Edge supports Redis as well, and you can configure it by passing in the `--redis-url` command line argument or setting the `REDIS_URL` environment variable.
+
+Edge mode also supports dynamic tokens, meaning that when using a [client token](https://docs.getunleash.io/reference/api-tokens-and-client-keys#client-tokens) with the `/api/client/features` endpoint, even without previously knowing about it, Edge will validate upstream and fetch its respective features. After that, it gets added to the list of known tokens that gets periodically synced, making sure it is a valid token and its features are up-to-date.
```
gastonfournier, line 63: I'd add a couple of notes:
1. Edge does not need a token to start
2. It might be useful to provide an initial token in order to warm it up (and how to do it)

README.md
```diff
@@ -30,48 +30,91 @@ Options:
...
+
+[Front-end tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens) can also be used with `/api/frontend` and `/api/proxy` endpoints, however they are not allowed to fetch features upstream. In order to use these tokens correctly and make sure they return the correct information, it's important that the features they are allowed to access are already present in that Edge node's features cache. The easiest way to ensure this is by passing in at least one client token as one of the command line arguments, ensuring it has access to the same features as the front-end token you'll be using.
```
gastonfournier, line 65:
```suggestion
[Front-end tokens](https://docs.getunleash.io/reference/api-tokens-and-client-keys#front-end-tokens) can also be used with `/api/frontend` and `/api/proxy` endpoints, however they are not allowed to fetch features upstream. In order to use these tokens correctly and make sure they return the correct information: e.g. by passing in at least one client token as one of the command line arguments, ensuring it has access to the same features as the front-end token you'll be using.
```
I'd try to provide an example.

I also removed `it's important that the features they are allowed to access are already present in that Edge node's features cache` because I'm not sure how customers will be able to notice this. Perhaps the explanation above is enough

README.md
```diff
@@ -30,48 +30,91 @@ Options:
...
+
+Besides dynamic tokens, Edge mode also supports metrics and other advanced features.
```
gastonfournier, line 67:
```suggestion
```

I think we can yeet this line because we're already mentioning that here: https://github.com/Unleash/unleash-edge/pull/60/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R35

gastonfournier (commented):
Looks really nice! I appreciate the mermaid graphs. Just a few comments, let me know what you think about them

README.md
```diff
@@ -30,48 +30,91 @@ Options:
...
+Another than connecting Edge directly to your Unleash instance, it's also possible to connect to another Edge instance (_daisy chaining_). You can have as many Edge nodes as you'd like between the Edge node your clients are accessing and the Unleash server, and it's also possible for multiple nodes to connect to a single upstream one. Depending on your architecture and requirements this can be a powerful feature, offering you flexibility and scalability when planning your implementation.
 
-Any query against `/api/frontend` or `/api/proxy` with a valid key will receive only enabled toggles.
+```mermaid
+graph LR
+  A(Client 1) -->|Fetch toggles| C((Edge 1))
+  B(Client 2) -->|Fetch toggles| D((Edge 2))
+  C-->|Fetch features| E((Edge 3))
+  D-->|Fetch features| E
+  E-->|Fetch features| F((Unleash))
```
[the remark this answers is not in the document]
nunogois: I agree, but should we use `toggles` or `features`?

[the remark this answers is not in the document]
gastonfournier: I'd go with toggles, features is what the user of Unleash develops and toggles allows them to toggle them on and off

README.md
nunogois: Good points, addressed in https://github.com/Unleash/unleash-edge/pull/60/commits/970533f963717d32a29c3b3a78a8dd7b11430ad2

README.md
nunogois: Unsure about this change, I'll leave it up for discussion a bit longer if anyone else wants to chime in.

README.md
nunogois: You have a point. Does that mean we should also yeet 109 @ https://github.com/Unleash/unleash-edge/pull/60/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R109 ?

README.md
```diff
@@ -30,48 +30,93 @@ Options:
...
+Edge mode is the "standard" mode for Unleash Edge and the one you should default to in most cases. It connects to an upstream node, such as your Unleash instance, and uses that as the source of truth for feature toggles.
 
-If your keys follow the Unleash API key format `[project]:[environment].<somesecret>`, Edge will filter the features dump to match the project contained in the key.
+Another than connecting Edge directly to your Unleash instance, it's also possible to connect to another Edge instance (_daisy chaining_). You can have as many Edge nodes as you'd like between the Edge node your clients are accessing and the Unleash server, and it's also possible for multiple nodes to connect to a single upstream one. Depending on your architecture and requirements this can be a powerful feature, offering you flexibility and scalability when planning your implementation.
```
chriswk, line 48:
```suggestion
Other than connecting Edge directly to your Unleash instance, it's also possible to connect to another Edge instance (_daisy chaining_). You can have as many Edge nodes as you'd like between the Edge node your clients are accessing and the Unleash server, and it's also possible for multiple nodes to connect to a single upstream one. Depending on your architecture and requirements this can be a powerful feature, offering you flexibility and scalability when planning your implementation.
```

nunogois: Nice catch!

README.md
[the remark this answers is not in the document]
nunogois: Made me wonder if we should try to go with `toggles` instead of `features` everywhere else 🤔 maybe something to consider

chriswk (approved):
Nice! :)

sighphyre (approved):
LG

PR #61, opened 2023-02-13
Open api docs

chriswk:
## What happens here
This PR adds OpenAPI documentation for all client facing endpoints.

It does not complete description tasks and probably needs some more details in the doc for the endpoints.


server/src/client_api.rs
```diff
@@ -64,8 +86,19 @@ async fn show_applications(
...
 
+#[utoipa::path(
+    path = "/api/client/metrics",
+    responses(
+        (status = 202, description = "Accepted client metrics"),
+        (status = 403, description = "Was not allowed to access features"),
+    ),
+    request_body = ClientMetrics,
+    security(
+        ("Authorization" = [])
```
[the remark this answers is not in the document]
chriswk: As I understood it says you need an Authorization header, but it doesn't have a default value.

server/src/client_api.rs
```diff
@@ -8,8 +8,19 @@ use unleash_types::client_metrics::{
     from_bucket_app_name_and_env, ClientApplication, ClientMetrics,
 };
 
+#[utoipa::path(
+    path = "/api/client/features",
+    responses(
+        (status = 200, description = "Return feature toggles for api key used that evaluated to true", body = ClientFeatures),
```
nunogois, line 14: Maybe we should use `token` instead? Sentence also read a bit weird to me. I was also under the impression that `client/features` returned all features, not only the ones that evaluated to true.

```suggestion
        (status = 200, description = "Return feature toggles for this token", body = ClientFeatures),
```

server/src/frontend_api.rs
```diff
@@ -11,8 +11,21 @@ use unleash_yggdrasil::{Context, EngineState};
 
 use crate::types::{EdgeJsonResult, EdgeSource, EdgeToken};
 
+///
```
nunogois, line 14: Extra line perhaps?

```suggestion

```

server/src/frontend_api.rs
```diff
@@ -11,8 +11,21 @@ use unleash_yggdrasil::{Context, EngineState};
...
+/// Returns all evaluated toggles for the key used
+#[utoipa::path(
+    path = "/api/proxy/all",
+    responses(
+        (status = 200, description = "Return all known feature toggles for api key used in evaluated (true|false) state", body = FrontendResult),
```
nunogois, line 19:
```suggestion
        (status = 200, description = "Return all known feature toggles for this token in evaluated (true|false) state", body = FrontendResult),
```

server/src/frontend_api.rs
```diff
@@ -29,6 +42,18 @@ async fn get_frontend_features(
     Ok(Json(FrontendResult { toggles }))
 }
 
+#[utoipa::path(
+    path = "/api/proxy/all",
+    responses(
+        (status = 200, description = "Return all known feature toggles for api key used in evaluated (true|false) state", body = FrontendResult),
```
nunogois, line 48:
```suggestion
        (status = 200, description = "Return all known feature toggles for this token in evaluated (true|false) state", body = FrontendResult),
```

server/src/frontend_api.rs
```diff
@@ -47,6 +72,18 @@ async fn post_frontend_features(
     Ok(Json(FrontendResult { toggles }))
 }
 
+#[utoipa::path(
+    path = "/api/proxy",
+    responses(
+        (status = 200, description = "Return feature toggles for api key used that evaluated to true", body = FrontendResult),
```
nunogois, line 78:
```suggestion
        (status = 200, description = "Return feature toggles for this token that evaluated to true", body = FrontendResult),
```

server/src/frontend_api.rs
```diff
@@ -67,6 +104,18 @@ async fn get_enabled_frontend_features(
...
```
nunogois, line 110:
```suggestion
        (status = 200, description = "Return feature toggles for this token that evaluated to true", body = FrontendResult),
```

nunogois (approved):
Pretty cool, don't mind my comments too much, just stuff that caught my eye when reading through and we can always get back to this at a later stage.

server/src/frontend_api.rs
chriswk: Yeah, but we should add a proper comment here, it will be parsed and added to the description of the call.

PR #64, opened 2023-02-15
fix: an issue where client features wouldn't correctly update

sighphyre:
Client features were being read on update, correctly calculated and then subsequently discarded. This patch makes incoming features actually get written to the memory provider


gastonfournier (approved):
James Bond and Jason Bourne, what a combination :+1:

sighphyre:
> James Bond and Jason Bourne, what a combination +1

I'm glad you noticed, I several minutes looking up famous spies so I could keep our test data consistent haha

PR #67, opened 2023-02-15
fix: Make offline mode handle non-Unleash tokens as valid secrets

chriswk:
So, our FromRequest for EdgeToken was failing for offline mode now that we had changed our try_from to only allow tokens following the UnleashToken spec to actually parse successfully, thus 403'ing all requests against an unleash server in offline mode. This fixes that so that offline mode again supports users using secret123 as their token if they so desire.


sighphyre (approved):
G

PR #69, opened 2023-02-15
fix: Move to distroless with libc

chriswk:
Performance with musl images was awful. This PR changes to use distroless and makes our builds build for -gnu instead of -musl

## Future improvement
- Use non-root user, I have a template for how to do it, but this PR just moves us to distroless for now


sighphyre (approved):
LGTM

nunogois (approved):
LGTM

PR #71, opened 2023-02-20
task: Document initial benchmarks

chriswk:
A small section on benchmarking added, I feel like this needs more fleshing out, but at least it adds some initial numbers.


gastonfournier (approved):
LGTM!

PR #72, opened 2023-02-21
Chore/data store refactor

sighphyre:
# What

This adds an extra layer of indirection between our data sources and how they're used.

# Why

The datastores are a likely point of change in this system, specifically they're subject to take on many new implementations to handle different data storage technologies. As this project grows, implementing the domain logic in each of these sub systems is going to be painful to implement given its complexity. This breaks apart the data storage method from the domain logic that needs to massage that data before it can be stored. In future this means implementing a data store becomes dumb storage operations while the domain logic is handled in a single place  

# What's next?

I'm not convinced this is a perfect approach. Before this refactor we had a lot of locks floating around and ultimately this still requires hitting an external storage for all read operations even though that data is provably small enough to fit in memory. I'd like to get to the point where our data stores are wrapped by the memory provider, lazily persist their data in the background and invalidating the memory provider's cache when it can be proven a write operation changes the validity of the data. This would mean we retain the performance of close to a hashmap lookup for read operations while reducing locking on the hot path


server/src/data_sources/builder.rs
```diff
@@ -1,79 +1,98 @@
...
+    features_refresh_interval_seconds: Duration,
+) -> EdgeResult<DataProviderPair> {
     let data_source = Arc::new(RwLock::new(RedisProvider::new(&redis_url)?));
-    Ok((data_source.clone(), data_source))
+    let facade = Arc::new(DataSourceFacade {
+        token_source: data_source.clone(),
+        feature_source: data_source.clone(),
+        token_sink: data_source.clone(),
+        feature_sink: data_source,
+        features_refresh_interval: Some(features_refresh_interval_seconds),
```
chriswk, line 67: This smells like a helper function, really only two arguments needed here, would be nice to not have to type it fully out each time. Not a blocker, just an improvement

chriswk (approved):
I like it. I particularly like that we don't have locks being passed around in app_data. Seems like we might be multi-threaded again?

server/src/data_sources/builder.rs
sighphyre: Yes this is a good suggestion. I'm not sure I'm content with this refactor yet, it's good enough for now but I want to do another pass to solve the threading issues. So this code might change during that restructure

PR #73, opened 2023-02-23
task: updated workflow to use regex

chriswk:
Since we're prefixing tags with unleash-edge-, metadata-action's semver type fails to work on our released tags. This PR changes to using patterns for creating the {{ version }} and {{ major.minor }} tags.


gastonfournier (approved):
LG!

PR #74, opened 2023-02-24
chore: removal of RW locks for dashmaps

gastonfournier:
## About the changes

This swaps out the underlying locks in the memory and Redis providers. Instead we use a DashMap, which holds internal locking primitives instead. This in turn means that we remove the lock contention issues when resolving features


sighphyre (approved):
OMG I love you guys

PR #75, opened 2023-02-24
chore: remove rwlock from validator, client and builder

gastonfournier:
This removes some additional RWLocks that we no longer need


nunogois (approved):
Gotta go fast 🚀

PR #76, opened 2023-02-24
feat: Added /api/frontend endpoint to match Unleash

chriswk:
## What
This PR adds `/api/frontend` and `/api/frontend/all` for both POST and GET.

## Why
This is done to match the recommended URLs from Unleash server for frontend clients.

## Reminder
We also need to make sure we answer to the /api/frontend/client/register as well, but that's a different PR.


sighphyre (approved):
Nice!

PR #79, opened 2023-02-28
fix: Client features were not refreshing.

chriswk:
We incorrectly assumed that our merge method would be enough here, but since the merge method retained the original and deduped, it seems like we tricked ourselves. The fix reduces the action to simply replacing whatever was cached with the newly fetched features from the server

Co-authored-by: Simon Hornby <liquidwicked64@gmail.com>
Co-authored-by: Nuno Gois <nuno.gois@getunleash.io>


sighphyre (approved):
G

PR #92, opened 2023-03-07
chore: update docs for file and redis persistence

sighphyre:


README.md
```diff
@@ -77,7 +77,7 @@ graph LR
 
 This means that, in order to start up, Edge mode needs to know where the upstream node is. This is done by passing the `--upstream-url` command line argument or setting the `UPSTREAM_URL` environment variable.
 
-By default, Edge mode uses in-memory to store the features it fetches from the upstream node. However, you may want to use a more persistent storage solution. Edge supports Redis as well, and you can configure it by passing in the `--redis-url` command line argument or setting the `REDIS_URL` environment variable.
+By default, Edge mode uses in-memory to store the features it fetches from the upstream node. However, you may want to use a more persistent storage solution. Edge supports either Redis or a backup file, you can configure one of these by passing in the `--redis-url` command line argument or or `--backup_folder` respectively. On start-up, Edge check the persistent backup if one is specified, and use this to populate its internal caches. This can be useful when your Unleash server is unreachable.
```
nunogois, line 80:
```suggestion
By default, Edge mode uses an in-memory cache to store the features it fetches from the upstream node. However, you may want to use a more persistent storage solution. For this purpose, Edge supports either Redis or a backup file, which you can configure by passing in either the `--redis-url` or `--backup_folder` command line argument, respectively. On start-up, Edge checks whether the persistent backup option is specified, in which case it uses it to populate its internal caches. This can be useful when your Unleash server is unreachable.
```

README.md
```diff
@@ -97,7 +97,9 @@ Options:
   -u, --upstream-url <UPSTREAM_URL>
           Where is your upstream URL. Remember, this is the URL to your instance, without any trailing /api suffix [env: UPSTREAM_URL=]
   -r, --redis-url <REDIS_URL>
-          [env: REDIS_URL=]
+          A url to a running Redis instance. Edge will use this instance to persist feature and token data and read this back after restart. Mutually exclusive with the --backup-folder option [env: REDIS_URL=]
```
nunogois, line 100:
```suggestion
         A URL pointing to a running Redis instance. Edge will use this instance to persist feature and token data and read this back after restart. Mutually exclusive with the --backup-folder option [env: REDIS_URL=]
```

chriswk (approved):
Nice

PR #93, opened 2023-03-07
docs: Updated benchmarks

chriswk:
### What
This updates docs with benchmarks ran today (March 7th) after having merged our redesign branch that made sure that we cache Yggdrasils computed state per token, rather than rebuild the state for each request that comes in.


sighphyre (approved):
G

PR #95, opened 2023-03-07
docs: update release workflow

chriswk:
Cargo dist (what we use to build binaries) has been updated. This updates to use the recommended flow from 0.0.4 (instead of 0.0.2). Amongst other things, it should allow us to build for M1/M2 macs


README.md
```diff
@@ -45,6 +45,22 @@ Options:
...
+### Binary
+ * The binary is downloadable from our [Releases page](https://github.com/Unleash/unleash-edge/releases/latest). 
+ * We're currently building for linux x86_64, windows x86_64, darwin (OS X) x86_64 and darwin (OS X) aarch64 (M1/M2 macs)
+ * NOTE: we're not currenly building ARM binaries, but we are building ARM docker files, so if you are running on an ARM platform, feel free to use our Docker images instead.
+
+### Docker
+ * The docker image gets uploaded to dockerhub and Github Package registry.
+ * For dockerhub use the coordinates `unleashorg/unleash-edge:<version>`.
+ * For Github package registry use the coordinates `ghpr.io/unleash/unleash-edge:<version>`
+ * If you'd like to live on the edge (sic) you can use the tag `edge`. This is built from `HEAD` on each commit
```
gastonfournier, line 60: I see what you did :clap:

gastonfournier (approved):
Out of curiosity, it seems that both README files (one under server, and the other one in root) are pretty similar. Is there any reason for that?

README.md
chriswk: > Out of curiosity, it seems that both README files (one under server, and the other one in root) are pretty similar. Is there any reason for that?

So the root README is the one displayed here in the github repo, unfortunately the README for crates are copied from the server subfolder. So that's why the duplication. 

I think the better solution is to use Cargo configuration to pickup the README from the root folder. I haven't gotten around to that yet.

PR #96, opened 2023-03-07
docs: move comparison with proxy to the end

thomasheartman:
This change moves the edge/proxy comparison section towards the end of
the document (just before the development) section.

The reason is that that section isn't directly about edge and it's
capabilities. I think it makes sense to present what Edge is and how
it works first, and instead present the comparison as extra reading
for people with a special interest.


sighphyre (approved):
Amazing, love it

PR #98, opened 2023-03-08
chore: Added tests for edge and client_api

chriswk:
Also added some Tarpaulin exclusion arguments to better reflect code coverage across code that we actually want to cover.

There's still some work to do with regards to having the possibility to instantiate the entire application as a test service for full integration/e2e tests, but this at least covers more logic and also exposed a bug in how we validated keys when not using a token validator (offline mode))

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -1,7 +1,7 @@
 use crate::error::EdgeError;
 use crate::metrics::client_metrics::{ApplicationKey, MetricsCache};
 use crate::tokens::cache_key;
-use crate::types::{EdgeJsonResult, EdgeResult, EdgeToken};
+use crate::types::{EdgeJsonResult, EdgeResult, EdgeToken, ProjectFilter};
 use actix_web::web::{self, Json};
 use actix_web::{get, post, HttpResponse};
 use dashmap::DashMap;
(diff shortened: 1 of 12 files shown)


server/src/client_api.rs
```diff
@@ -28,8 +28,12 @@ pub async fn features(
     features_cache: web::Data<DashMap<String, ClientFeatures>>,
 ) -> EdgeJsonResult<ClientFeatures> {
     features_cache
-        .get(&cache_key(edge_token))
+        .get(&cache_key(edge_token.clone()))
         .map(|features| features.clone())
+        .map(|client_features| ClientFeatures {
```
sighphyre, line 33: YESSS. Let's also make sure this goes into the commit message as a fix

server/src/client_api.rs
```diff
@@ -201,4 +252,170 @@ mod tests {
         assert_eq!(found_metric.no, 0);
         assert_eq!(found_metric.no, expected.no);
     }
+
+    fn cached_client_features() -> ClientFeatures {
+        ClientFeatures {
+            version: 2,
+            features: vec![
+                ClientFeature {
```
sighphyre, line 260: I think we have a `Default` impl on ClientFeature. Entirely optional but it might make this huge blob easier to read

server/tarpaulin.toml
```diff
@@ -0,0 +1,2 @@
+[all]
+exclude-files= ["src/main.rs", ""
```
sighphyre, line 2: YEET IT. YEET IT WITH FIRE

sighphyre (approved):
This is great, sets my heart at ease

PR #99, opened 2023-03-08
fix: Uses token cache when resolving token from request

chriswk:
So, our FromRequest for EdgeToken (What Actix uses to allow us to say that our handler functions expects an EdgeToken as an argument) didn't use our token cache, which meant that our token validation (getting projects from upstream for multi project tokens) wasn't working. This PR adds another test verifying that tokens with access to the three projects we use for our production servers works against a snapshot of the hosted state.


sighphyre (approved):
LGTM

PR #100, opened 2023-03-08
fix: Move token cache resolution out of FromRequest

chriswk:
This was unfortunatley the wrong place to have it, it meant we never hit the middleware which validated our token. So unfortunately, we had to make our endpoints depend on the tokencache in order to have the up to date EdgeToken, rather than the one that comes in with the request.

I fear this has broken Offline Mode, so not quite ready to merge just yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2752,9 +2752,9 @@ dependencies = [
 
 [[package]]
 name = "unleash-types"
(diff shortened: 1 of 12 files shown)


chriswk:
Made sure offline mode also works

sighphyre (approved):
Nice one, LGTM

PR #102, opened 2023-03-09
task: Make Feature refresher register as client

chriswk:
When a new token for registration comes in, feature refresher registers as a client upstream

We've discussed this before. After talking to Gard and looking at what we were using client registrations for, it made sense for Edge to register as a client for Client tokens that it's fetching data for.

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -28,13 +28,13 @@ pub struct FeatureRefresher {
     pub persistence: Option<Arc<dyn EdgePersistence>>,
 }
 
-fn client_application_from_token(token: EdgeToken) -> ClientApplication {
+fn client_application_from_token(token: EdgeToken, refresh_interval: i64) -> ClientApplication {
     ClientApplication {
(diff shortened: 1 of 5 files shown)


gardleopard (approved):
Fantastic!

sighphyre (approved):
LGTM!

server/src/http/feature_refresher.rs
```diff
@@ -26,6 +28,19 @@ pub struct FeatureRefresher {
     pub persistence: Option<Arc<dyn EdgePersistence>>,
 }
 
+fn client_application_from_token(token: EdgeToken) -> ClientApplication {
+    ClientApplication {
+        app_name: "unleash_edge".into(),
+        connect_via: None,
+        environment: token.environment,
+        instance_id: None,
+        interval: 5,
```
sighphyre, line 37: Should we include the actual polling interval here? Might be useful

chriswk: Agreed. I'll fix

PR #103, opened 2023-03-10
chore: adds a derive for TokenValidation status

sighphyre:
Adds a derive for TokenValidation status with a #[default] on the child enum. This is to prep for the update to Rust 1.68. This is due to an update to Clippy, not Rustc, still compiles on 1.67


chriswk (approved):
LGTM

PR #104, opened 2023-03-10
chore: update toolchain to rust 1.68

sighphyre:
Updates our rust toolchain to 1.68. This won't build until #103 is merged, due to a new Clippy warning

Edit: Oh. It passes, looks like we're not failing on Clippy warns


chriswk (approved):
Nice!

PR #109, opened 2023-03-14
task: Prepare a token revalidator

chriswk:
Edge caching tokens could cause tokens to become out of sync with Unleash Upstream, adding a scheduled background task which ensures that our known tokens are still valid will go a long way towards mitigating this.

Discussion point: Should this also reach into the FeatureRefresh cache if it finds an invalid token and remove the FeatureRefresh entry for invalid tokens.

I feel like it should, or rather, I feel like the FeatureRefresher should check status codes and if it receives 403's it should probably remove the token from its own cache.

TokenValidator being aware of the FeatureRefresh and Engine state caches feels like scope creep for the TokenValidator.

### DIFF
--- a/server/src/auth/token_validator.rs
+++ b/server/src/auth/token_validator.rs
@@ -1,8 +1,9 @@
 use crate::error::EdgeError;
 use crate::http::unleash_client::UnleashClient;
 use crate::persistence::EdgePersistence;
-use crate::types::{EdgeResult, EdgeToken, ValidateTokensRequest};
+use crate::types::{EdgeResult, EdgeToken, TokenValidationStatus, ValidateTokensRequest};
 use std::sync::Arc;
+use std::time::Duration;
 
 use dashmap::DashMap;
 use unleash_types::Upsert;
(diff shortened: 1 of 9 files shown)


nunogois (approved):
> Edge caching tokens could cause tokens to become out of sync with Unleash Upstream, adding a scheduled background task which ensures that our known tokens are still valid will go a long way towards mitigating this.
> 
> Discussion point: Should this also reach into the FeatureRefresh cache if it finds an invalid token and remove the FeatureRefresh entry for invalid tokens.
> 
> I feel like it should, or rather, I feel like the FeatureRefresher should check status codes and if it receives 403's it should probably remove the token from its own cache.
> 
> TokenValidator being aware of the FeatureRefresh and Engine state caches feels like scope creep for the TokenValidator.

Looks sane to me on a first look. Agree on `FeatureRefresh` cache dropping invalid tokens, but maybe it should clean after itself like you mentioned? 

Everything you said makes sense to me :+1:

gastonfournier (approved):
Looks fine! This made me wonder if edge-tokens should have a TTL in the long run

gastonfournier:
> Looks fine! This made me wonder if edge-tokens should have a TTL in the long run

Still, tokens can be invalidated upstream for security reasons, so it's fine to have them checked regularly

chriswk:
I've now updated this PR with updates to the FeatureRefresher, so it updates the map of tokens it has to refresh. I was worried about doing double work, then I realised that the FeatureRefresher doesn't actually look at the tokens the validator or the client uses. 

It has it's own TokenRefresh struct, so removing these still leaves it up to TokenValidator to keep track of validated/invalid status of the tokens it has seen.

For now, we'll keep both invalid and valid for the lifetime of the app, but we will no longer refresh invalid tokens (tokens that have received a 403 forbidden). With Persistence this becomes an extra challenging proposition. I believe the Persistence layer probably shouldn't keep Invalid tokens. WDYT @sighphyre ?

sighphyre:
> For now, we'll keep both invalid and valid for the lifetime of the app, but we will no longer refresh invalid tokens (tokens that have received a 403 forbidden). With Persistence this becomes an extra challenging proposition. I believe the Persistence layer probably shouldn't keep Invalid tokens. WDYT @sighphyre ?

Agreed, without a way to throw these away, we risk growing the backing data store until it dies under the weight of a billion garbage tokens. Means a little extra work when we get an invalid token from a calling Client but I don't believe that's a big risk, under normal operation there really shouldn't be that many

Still looks good to me

server/src/auth/token_validator.rs
```diff
@@ -91,12 +92,54 @@ impl TokenValidator {
...
+                    let _ = self.revalidate_known_tokens().await;
+                }
+            }
+        }
+    }
+    pub async fn revalidate_known_tokens(&self) -> EdgeResult<()> {
+        let tokens_to_validate: Vec<String> = self
+            .token_cache
+            .iter()
+            .filter(|t| t.value().status == TokenValidationStatus::Validated)
```
gastonfournier, line 110: Crazy idea: maybe we can validate all known tokens. Perhaps some were forbidden to access something and then re-added (because of reasons). We might revive the token on the next request that tries to use it... but I'm not so sure

chriswk: That feels risky with how many tokens that might end up revalidating.

Since we're now persisting only validated tokens, restarting edge will give you another attempt with an earlier forbidden token.

Since the default token generation process generates random strings/secrets, the chances of you deploying an app with a not-generated token and then going to generate it are low.

PR #111, opened 2023-03-17
task: Return 511 if edge has not hydrated.

chriswk:
Our client/frontend token separation leads to us having to hydrate client features using a client token. 

If a frontend token comes in that has access to a project/environment combination that Edge has not seen a client token for, this PR now makes Edge consistently return a 511 Network Authentication Required with a body explaining which project and environment the user has to add a client token for.

Using 511 instead of 403 seems more correct, since the token is indeed allowed to access the server, but we haven't yet authorized data upstream.


sighphyre (approved):
LGTM!

Issue #119, opened 2023-03-22
Edge doesn't keep track of known bad tokens


chriswk:
### Describe the bug

When calling edge with a known bad token and a locally running Unleash as the upstream, I see that we're calling the validate endpoint for each call that comes in. This leads to a bad/slow experience for the user who has to wait while we validate. We used to store this, but it seems like we lost this capability in the process of gaining the synchronous features fetching.

### Steps to reproduce the bug

1. start Unleash locally
2. start edge with upstream pointing towards local Unleash
3. curl -H "Authorization: *:development.alreadyknowntobebadtoken" http://localhost:3063/api/client/features
4. Same
5. Observe that both calls hit upstream

### Expected behavior

Only first call with a token should hit upstream /edge/validate. after that, Edge should cache the bad state of this token, at least until next scheduled revalidation of tokens.

### Subscription type

None

### Hosting type

None

### SDK information (language and version)

Edge HEAD / 1.0.0

PR #120, opened 2023-03-23
fix: Save checked tokens even if invalid

chriswk:
To make sure we don't hammer upstream with validation requests even if edge is under heavy load, this PR allows Edge to save validation results for invalid tokens as well.

fixes: #119


sighphyre (approved):
LG

PR #122, opened 2023-03-24
task: Post appropriately sized metric batches

chriswk:
Previously we would save unacknowledged metrics until upstream accepted the post. This PR, splits into 90kB chunks, listens for http status codes to decide what to do on failure.
* 400 will cause us to drop the metrics we tried to post
* 413 would be a surprise, since we already split into chunks to avoid just this
* other status codes will be reinserted to the cache and tried again next minute.

### DIFF
--- a/server/src/metrics/client_metrics.rs
+++ b/server/src/metrics/client_metrics.rs
@@ -1,7 +1,13 @@
 use chrono::{DateTime, Utc};
 use dashmap::DashMap;
+use serde::{Deserialize, Serialize};
 use std::hash::{Hash, Hasher};
+use tracing::{debug, instrument};
 use unleash_types::client_metrics::{ClientApplication, ClientMetricsEnv};
+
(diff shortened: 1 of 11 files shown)


server/src/metrics/client_metrics.rs
```diff
@@ -59,29 +88,118 @@ pub struct MetricsCache {
...
                 .applications
                 .iter()
                 .map(|e| e.value().clone())
                 .collect(),
-            metrics: self.metrics.iter().map(|e| e.value().clone()).collect(),
+            metrics: self
+                .metrics
+                .iter()
+                .map(|e| e.value().clone())
+                .filter(|m| m.yes > 0 || m.no > 0) // Makes sure that we only return buckets that have values. We should have a test for this :P
```
sighphyre, line 156: We have a test for it!

```suggestion
                .filter(|m| m.yes > 0 || m.no > 0) // Makes sure that we only return buckets that have values
```

sighphyre (approved):
LGTM!

PR #126, opened 2023-03-27
docs: Note about Unleash 4.22

chriswk:
So. Due to bulk metrics not being added to Unleash before 4.22, this adds a note to the doc pointing out that you'll need 4.22 in order for metrics to show up.

The important change is L184-L188, the rest is IntelliJ/VSCode trying to be helpful and formatting my markdown file. I'm very open to suggestions to the phrasing here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,13 @@ Options:
   -t, --tokens <TOKENS>                  [env: TOKENS=]
 ```
 
-## Metrics
+## [Metrics](https://docs.getunleash.io/reference/api/unleash/metrics)
 
 **❗  Note:**  For Unleash to correctly register SDK usage metrics sent from Edge instances, your Unleash instance must be v4.22 or newer.
+
+Since Edge is designed to avoid overloading its upstream, Edge gathers and accumulates usage metrics from SDKs for a set interval (METRICS_INTERVAL_SECONDS) before posting a batch upstream.
+This reduces load on Unleash instances down to a single call every interval, instead of every single client posting to Unleash for updating metrics.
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
LGTM 

Guessing all the white space is Intellij

README.md
```diff
@@ -176,8 +181,14 @@ Options:
   -t, --tokens <TOKENS>                  [env: TOKENS=]
 ```
 
+## Metrics
+
+| :exclamation: | To see SDK usage metrics from Edge instances, you will need Unleash v4.22 or newer. |
+| ------------- | ----------------------------------------------------------------------------------- |
```
thomasheartman, line 187: I take it this is supposed to be an admonition, right? I would do something like this if you want to use it call it out. Using a table seems semantically very dubious. But also where do you "see metrics"? Is that in the network view? Or doesn't Unleash process it at all before 4.22? (Also, not everyone can see 💁🏼)

But also: is there nothing more to say about metrics? It looks like it's missing a description? Does edge report metrics? How does it work? I suggest putting a paragraph or two after the admonition to explain how edge works with metrics.

```suggestion
**❗  Note:**  For Unleash to correctly register SDK usage metrics sent from Edge instances, your Unleash instance must be v4.22 or newer.
```

thomasheartman (commented):
Nice! But I'd like a little more clarification on metrics in general, if you don't mind?

Also, regarding formatting tables: I don't mind whether they're formatted for reading as plaintext or to save space, as long as we avoid formatting and reformatting back and forth between two versions constantly.

README.md
chriswk: Agreed, will update. Thanks for the feedback. :)

nunogois (approved):
LGTM

PR #128, opened 2023-03-27
Proposal

gastonfournier:
To keep previous scope

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -35,7 +35,7 @@ pub async fn get_features(
         .ok_or(EdgeError::AuthorizationDenied)?;
     match req.app_data::<Data<FeatureRefresher>>() {
         Some(refresher) => refresher
-            .features_for_token(&validated_token)
+            .features_for_token(validated_token)
             .await
             .map(Json),
         None => features_cache
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -249,9 +249,11 @@ async fn post_enabled_features(
         .get(&edge_token.token)
         .map(|e| e.value().clone())
         .unwrap_or_else(|| edge_token.clone());
-    let engine = engine_cache.get(&tokens::cache_key(&edge_token)).ok_or(
-        EdgeError::FrontendNotYetHydrated(FrontendHydrationMissing::from(&edge_token)),
-    )?;
+    let engine = engine_cache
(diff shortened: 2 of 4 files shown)


server/src/client_api.rs
```diff
@@ -35,7 +35,7 @@ pub async fn get_features(
         .ok_or(EdgeError::AuthorizationDenied)?;
     match req.app_data::<Data<FeatureRefresher>>() {
         Some(refresher) => refresher
-            .features_for_token(&validated_token)
+            .features_for_token(validated_token)
```
chriswk, line 38: would this compile later on, since we reuse the validated token further down in this method? I see since you're making a PR against another PR, our workflow isn't triggering.

server/src/frontend_api.rs
```diff
@@ -249,9 +249,11 @@ async fn post_enabled_features(
...
         .map(|e| e.value().clone())
         .unwrap_or_else(|| edge_token.clone());
-    let engine = engine_cache.get(&tokens::cache_key(&edge_token)).ok_or(
-        EdgeError::FrontendNotYetHydrated(FrontendHydrationMissing::from(&edge_token)),
-    )?;
+    let engine = engine_cache
+        .get(&tokens::cache_key(&edge_token))
+        .ok_or_else(|| {
+            EdgeError::FrontendNotYetHydrated(FrontendHydrationMissing::from(&edge_token))
+        })?;
```
chriswk, line 256: Yes. Thank you, good catch

server/src/metrics/client_metrics.rs
```diff
@@ -488,7 +488,7 @@ mod test {
             cache.applications.insert(
                 ApplicationKey {
                     app_name: app.app_name.clone(),
-                    instance_id: app.instance_id.clone().unwrap_or("unknown".into()),
+                    instance_id: app.instance_id.clone().unwrap_or_else(|| "unknown".into()),
```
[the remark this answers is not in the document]
gastonfournier: Yes, it wasn't me :D

server/src/frontend_api.rs
gastonfournier: Clippy too

server/src/client_api.rs
gastonfournier: I run test locally and they passed, but I'm also not thaaat confident in my skills with Rust

PR #135, opened 2023-03-28
fix: added custom metrics handler to drop dependency

chriswk:
This PR merges the actix-web-opentelemetry library's prometheus code into edge, so we don't need to depend on it anymore.


sighphyre (approved):
LGTM

Issue #137, opened 2023-03-29
Support for /api/client/features/<<featureName>>


sebastian-bury:
### Describe the feature request

I would like the ability to call /api/client/features/<<featureName>> 
as seen in the docs: https://docs.getunleash.io/reference/api/unleash/get-client-feature

So that I can call edge for a single toggle by name and that edge supports the same APIs as if it were Unleash/proxy/frontend/etc.

PR #138, opened 2023-03-29
feat: Add client feature endpoint

chriswk:
Our client api in Unleash server also supports querying single features by name in path. This PR adds the necessary endpoint to support this.

Fixes: #137


server/src/client_api.rs
```diff
@@ -51,6 +51,51 @@ pub async fn get_features(
...
+    edge_token: EdgeToken,
+    features_cache: Data<DashMap<String, ClientFeatures>>,
+    token_cache: Data<DashMap<String, EdgeToken>>,
+    feature_name: web::Path<String>,
+    req: HttpRequest,
+) -> EdgeJsonResult<ClientFeature> {
+    let validated_token = token_cache
+        .get(&edge_token.token)
+        .map(|e| e.value().clone())
+        .ok_or(EdgeError::AuthorizationDenied)?;
```
nunogois, line 78: Do we need this check? Seems to be the same one as in `get_features`. Would be nice if it was handled through a middleware / helper method for DRY-ness.

nunogois (approved):
LGTM

server/src/client_api.rs
chriswk: I don't disagree, I'd like that. Haven't been able to work out how to do it without doing what feels like too much logic in a FromRequest implementation. In a middleware you can't put it as an argument to the function in an easy way. The one time where types are fighting us. :) The clue is to get the project(s) and environment that we've fetched from the upstream validate call.

I'm open to suggestions for a better way to do this, we basically do the same in all endpoints.

Issue #145, opened 2023-04-11
Missing metrics endpoints under /frontend and /proxy


chriswk:
### Describe the feature request

We're missing `POST /api/proxy/client/register` as well as `POST /api/frontend/client/register` and `POST /api/frontend/client/metrics`

PR #147, opened 2023-04-11
Feat/add metrics endpoints for frontend

chriswk:
After having a slack chat in the community slack with a new user missing metrics when trying to use the `/api/frontend` endpoint we realised that we were still missing a couple of endpoints for the frontend API. This PR adds the missing endpoints, duplicating what we did for /api/client/register and /api/client/metrics, just under `/api/(proxy|frontend)/client/register` and `/api/(proxy|frontend)/client/metrics` instead.

fixes: #145

### DIFF
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -9,12 +9,11 @@ use actix_web::{
 use dashmap::DashMap;
 use unleash_types::client_metrics::{ClientApplication, ConnectVia};
 use unleash_types::{
-    client_metrics::{from_bucket_app_name_and_env, ClientMetrics},
+    client_metrics::ClientMetrics,
     frontend::{EvaluatedToggle, EvaluatedVariant, FrontendResult},
 };
 use unleash_yggdrasil::{Context, EngineState, ResolvedToggle};
 
-use crate::metrics::client_metrics::ApplicationKey;
 use crate::{
     error::{EdgeError, FrontendHydrationMissing},
     metrics::client_metrics::MetricsCache,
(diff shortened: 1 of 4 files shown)


server/src/frontend_api.rs
```diff
@@ -262,11 +265,52 @@ async fn post_enabled_features(
...
+
+    let metrics = from_bucket_app_name_and_env(
+        metrics.bucket,
+        metrics.app_name,
+        edge_token.environment.unwrap(),
+    );
+
+    metrics_cache.sink_metrics(&metrics);
+
+    Ok(HttpResponse::Accepted().finish())
```
gastonfournier, line 295: Is this code the same as for /fronted/client/metrics? I don't know if it's worth having a private method with this logic or if it's just fine having this duplication. Not a big deal TBH

server/src/frontend_api.rs
```diff
@@ -281,16 +325,98 @@ async fn post_frontend_metrics(
...
+    metrics_cache.applications.insert(
+        ApplicationKey {
+            app_name: to_write.app_name.clone(),
+            instance_id: to_write
+                .instance_id
+                .clone()
+                .unwrap_or_else(|| ulid::Ulid::new().to_string()),
+        },
+        to_write,
+    );
```
gastonfournier, line 404: Also, same consideration as above, the code is the same as for /proxy/client/register, but perhaps it's fine

gastonfournier (approved):
I think this looks good, just some code duplication but maybe it's fine as is

server/src/frontend_api.rs
chriswk: Yeah, this is the third time we duplicate it, so might be worth extracting to a common method. Thanks for giving me the push.

sighphyre (approved):
LGTM

Issue #148, opened 2023-04-11
Support filtering /api/client/features by namePrefix query parameter


chriswk:
### Describe the feature request

Unleash's /api/client/features endpoint support filtering by tags, projects and namePrefix. Edge should support the same to make it opaque for our SDKs whether they connect to an Edge instance or an Unleash instance

### Background

Tag filter:
- Edge can't support filtering by tags, since that is not data we get from Unleash (it's not part of the client features contract), and Edge does not have access to the database.

Project filter:
- This was a stopgap while we didn't have project access/filtering in the token, so we could add this, but it feels like you should instead use a token with correct access to the projects you want to filter by

Name prefix filter:
- This should be easily doable

### Solution suggestions

Keep current logic in edge, but add another filter after having fetched the features the token has access to filtering by featurename starting with prefix.

PR #149, opened 2023-04-11
feat: add namePrefix filtering support

chriswk:
fixes: #148


sighphyre (approved):
Nice looks good to me. We'll probably get a bit of push back on tags/project filter but we can just... push back harder

PR #151, opened 2023-04-12
feat: Added single feature evaluation endpoint

chriswk:
For now, we resolve all toggles, waiting for feature improvement in Yggdrasil that allows us to query single ResolvedToggle by name


sighphyre (approved):
LGTM

PR #153, opened 2023-04-12
Feat/custom client headers

### LINKED ISSUE #152: Add possibility to include custom client headers

We've had multiple customer request the possibility to add custom client headers to requests going to upstream client that does the feature fetching.

### Background

You could have an API gateway in front of the upstream server that requires its own client key. This feature would allow us to configure the http client used to fetch features upstream with extra headers, thus successfully authenticating with both the API gateway and the unleash server.

### Solution suggestions

_No response_

chriswk:
As stated in #152 we've seen the need to add support for adding more than the Authorization header to the client used for upstream communication.

This PR adds a `--custom-client-headers` argument which accepts a comma separated list of <headername>:<headervalue> pairs or passed in with multiple invocations of `--custom-client-headers` or a comma separated list in the environment variable `CUSTOM_CLIENT_HEADERS`. It also uses the recognizable shorthand from curl for headers `-H` :)

fixes #152


server/Cargo.toml
```diff
@@ -29,6 +29,7 @@ dashmap = "5.4.0"
 dotenv = {version = "0.15.0", features = ["clap"]}
 futures = "0.3.28"
 futures-core = "0.3.28"
+iter_tools = "0.1.4"
```
sighphyre, line 32: Itertools is great, we probably should have added this one ages ago!

server/src/cli.rs
```diff
@@ -39,6 +41,28 @@ pub struct EdgeArgs {
...
+
+    /// Expects curl header format (-H <HEADERNAME>: <HEADERVALUE>)
+    /// for instance `-H X-Api-Key: mysecretapikey`
+    #[clap(short = 'H', long, env, value_delimiter = ',', value_parser = string_to_header_tuple)]
+    pub custom_client_headers: Vec<(String, String)>,
+}
+
+pub fn string_to_header_tuple(s: &str) -> Result<(String, String), String> {
+    let format_message =
+        "Please pass headers in the format <headername>: <headervalue>".to_string();
```
sighphyre, line 53: If I'm reading the code correctly, that whitespace is optional there
```suggestion
        "Please pass headers in the format <headername>:<headervalue>".to_string();
```

server/src/http/unleash_client.rs
```diff
@@ -97,30 +100,49 @@ impl UnleashClient {
...
         } else {
             client_req
         }
     }
 
-    #[instrument]
+    fn header_map(&self, api_key: Option<String>) -> HeaderMap {
+        let mut header_map = HeaderMap::new();
+        if let Some(key) = api_key {
+            header_map.insert(header::AUTHORIZATION, key.parse().unwrap());
```
sighphyre, line 122: Minor detail but these unwraps can fail if someone passes goop into the headers. Agree that a panic is a good choice here, but an `expect` over an `unwrap`, might be nice for understanding what the user did wrong here. WDYT?

sighphyre (approved):
Nice one! LGTM, left you a few small thoughts. Feel free to ignore them if you disagree

server/src/http/unleash_client.rs
chriswk: Agreed. will fix

PR #156, opened 2023-04-14
fix: Fixes incorrect parsing of extra arguments.

chriswk:
Currently we do not parse extra parameters into the properties holder for the context. This PR updates edge to make sure that overflows (properties we haven't explicitly defined) get parsed into the properties map


sighphyre (approved):
LGTM!

PR #157, opened 2023-04-17
feat: allow cli option to disable ssl verification

sighphyre:
This adds an option to the CLI arguments to turn off SSL verification when talking to the upstream Unleash server. This defaults to false and is only intended to ease of development - as per feature request from end user


server/src/http/unleash_client.rs
```diff
@@ -478,4 +523,34 @@ mod tests {
...
+        let srv = test_features_server_with_untrusted_ssl().await;
+        let client = UnleashClient::new_insecure(srv.surl("/").as_str()).unwrap();
+
+        let validate_result = client
+            .validate_tokens(ValidateTokensRequest {
+                tokens: vec![TEST_TOKEN.to_string()],
+            })
+            .await;
+
+        validate_result.unwrap();
```
chriswk, line 539:
```suggestion
        assert!(validate_result.is_ok());
```

chriswk (approved):
LGTM

Issue #158, opened 2023-04-18
Edge does not handle `properties[key]=value` as part of context


chriswk:
### Describe the bug

After v1.3.1 we managed to use serde's flatten to get unknown query variables inserted into the properties field on the Context struct. However, we do not handle the proxy-client's `properties[key]=value` syntax

### Steps to reproduce the bug

1. [Define a custom context field](https://docs.getunleash.io/how-to/how-to-define-custom-context-fields)
   1. Call it email 
2. Setup a constraint on the custom context field "IN test@test.com"
3. curl -H"Authorization: <yourFrontendToken>" http://<EDGEURL>/api/proxy?properties%5Bemail%5D=test@test.com
4. Observe that no toggles comes back as enabled, even though our context should have included email

### Expected behavior

Context should include key with value when passed in using `properties[key]=value` in the GET url.

### Subscription type

None

### Hosting type

None

### SDK information (language and version)

unleash-edge 1.3.1 and unleash-edge 1.4.0

PR #160, opened 2023-04-19
fix: Building context parameters using map syntax

chriswk:
Currently Edge only supports the explicit context variables; properties[key]=value does not work. Since the proxy clients use the `properties[key]` syntax for setting custom context, this PR adds serde-qs for parsing query strings using the `properties[key]=value` syntax.

It also adds a quick proxy-client-js app for verifying that we can set custom context.
 
fixes: #158


sighphyre (approved):
LGTM

PR #161, opened 2023-04-19
task: added ecr-public login and push target

chriswk:
We've had requests for unleash-proxy to be pushed to ECR public gallery. So in order to get ahead of possible requests, this pushes to ECR public gallery as well as dockerhub and Github Package Registry.

Already added a separate ecr-public role and added the needed secrets.


gastonfournier (approved):
LG

PR #164, opened 2023-04-20
fix: Use split_once.

chriswk:
We had forgotten that `:` is a valid part of a header value, so splitting on `:` ended up splitting into too many parts and thus failing the parser. This PR changes to use split_once, which splits on first occurrence. Since Header Names are not allowed to contain `:` this will be fine


sighphyre (approved):
LGTM

PR #170, opened 2023-04-26
docs: update note about modes and docker

chriswk:
## About the changes
Update the docs to point out that even when running in docker you have to choose which mode you're running in.


sighphyre (approved):
LGTM!

PR #174, opened 2023-05-02
fix: Handle both upper and lowercase of apitokentype.

chriswk:
Unleash has suddenly started returning token type with uppercase. This PR makes us handle both UPPER and lower case for token type.

The real fix would be for Unleash to obey its own contract with lowercase for token types, but this fix makes us more tolerant to mistakes in Unleash code.


sighphyre (approved):
Well that's silly.

Nice catch! LGTM

PR #175, opened 2023-05-02
feat: switch aws auth to oidc

gardleopard:
This is to be more secure and rely on what repo it is rather than a secret to push to ECR


chriswk (approved):
LGTM

PR #182, opened 2023-05-11
feat: Add more visible info and warn logging for http errors

chriswk:
### What
Make sure that upstream certificate and client identity do not depend on each other.
In addition adds more warn and info logging for all upstream communication attempts. Most are added at info!, but the features_fetch call is important enough that we're adding a warn message.


sighphyre (approved):
LGTM

PR #187, opened 2023-05-17
chore: bump yggdrasil to pull through fix for rollout not working with random

sighphyre:
This bumps Yggdrasil to 0.5.6, which fixes an issue where gradual rollout with a random stickiness would always resolve to false


gastonfournier (approved):
G!

PR #188, opened 2023-05-18
fix: allow multiple client tokens at startup

sighphyre:
Fixes an issue where providing multiple tokens on startup with identical environments and projects would result in no tokens being registered.

This was caused by overeager subsumption of the tokens - identical tokens would subsume each other, and therefore the result of two identical tokens would be no tokens. I've patched this by creating a new list of only tokens that do not have a match for identical project/environment before we do the subsumption.


gastonfournier (approved):
I think this looks fine, especially the test case gives me confidence that this is the right thing to do :+1:

sighphyre:
> I think this looks fine, especially the test case gives me confidence that this is the right thing to do +1

Me too - that's why we write tests haha

PR #197, opened 2023-06-01
feat: makes a post to api/client/features possible by setting a cli arg

sighphyre:
## What

This exposes a boolean CLI arg called `enable_post_features` that will allow POST requests to `api/client/features` 

## Why

This is to support a very specific but ephemeral need for a user. This doesn't do anything useful outside of this need and it's absolutely not useful for our SDKs. To that end, I've placed the CLI arg in a new `flatten`ed CliArg called `ExperimentalArgs` and kept the code here isolated so that it's easy to tear out at a later stage

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -135,6 +135,13 @@ pub struct TlsOptions {
 }
 
 #[derive(Args, Debug, Clone)]
+pub struct ExperimentalArgs {
+    /// Exposes the api/client/features endpoint for POST requests. This may be removed in a future release
+    #[clap(env, long, default_value_t = false)]
+    pub enable_post_features: bool,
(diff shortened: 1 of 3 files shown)


ivarconr (approved):

chriswk:
Was going to suggest making that argument global, so you don't have to remember whether to add it before or after the sub command

sighphyre:
> Was going to suggest making that argument global, so you don't have to remember whether to add it before or after the sub command

That would have been a good idea, I should have thought of that. We can move it in the next major version if we want but (and this is a terrible thing to say), I don't care much about this feature. When relevant user is done I'd either like to remove it or put it behind a compiler flag and leave that off by default

Issue #200, opened 2023-06-02
I'd like to be able to configure redis with multiple env variables


dramamask:
### Describe the feature request

I'd like to be able to configure redis with multiple env variables, instead of just the `REDIS_URL` variable. For example:

```
REDIS_SCHEME
REDIS_HOST
REDIS_PASSWORD
REDIS_PORT
```

Unleash-edge could use those to assemble the correct redis URL. I'd prefer that since I now have to write my own logic to assemble the redis URL, taking things like password and scheme into account. Other services and frameworks use these types of env variables (e.g. Laravel). 

Especially `REDIS_SCHEME` would be very helpful. You would configure either `tcp` or `tls`. Logic inside unleash-edge would then automatically use either `redis://` or `rediss://` as the prefix for the Redis URL.

The `REDIS_PASSWORD` variable would be very helpful to because the password needs to be included as part of the `REDIS_URL` right now. (See https://github.com/Unleash/unleash-edge/issues/199 for info on how to use a redis password)

### Background

Right now I have to write my own logic to assemble the proper redis URL. We use different environments that require different REDIS_URLs (with or without password, some are tcp, some are tls).

Issue #199, opened 2023-06-02
Document how to use a redis instance that needs a password


dramamask:
### Describe the feature request

II'd love to see the following info added to the documentation.

To use a redis instance that needs a password, use the following redis url:

`redis://default:your_password@redis:6379`

For redis v5, you have to use `default` as the username. Redis v6 introduced usernames, so you just use whatever username you  have set.

pransh15:
Hey @dramamask 👋 

Thanks for submitting this feature request. Someone from the team will be on it soon! 😄

PR #202, opened 2023-06-05
feat: add multiple env variables for configuring Redis

chriswk:
In #199 and #200 we discussed that there are instances of Redis where configuring it with a single REDIS_URL variable doesn't make sense. This PR keeps compatibility with previous REDIS_URL configuration, but allows you to use REDIS_HOST, REDIS_USER, REDIS_PASSWORD, REDIS_PORT, REDIS_SCHEME and REDIS_SECURE (overrides scheme and sets it to rediss) to configure Redis as well.

Combinations like REDIS_URL and REDIS_USER and REDIS_PASSWORD is supported. If connecting to REDIS v5 with password you *must* set the REDIS_USER variable to `default`

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -1,6 +1,7 @@
+use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 
-use clap::{ArgGroup, Args, Parser, Subcommand};
+use clap::{ArgGroup, Args, Parser, Subcommand, ValueEnum};
 
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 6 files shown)


server/src/cli.rs
```diff
@@ -11,6 +12,73 @@ pub enum EdgeMode {
...
+                }
+                if self.redis_username.is_some() {
+                    base_url.set_username(&self.redis_username.clone().unwrap()).expect("Failed to set username");
+                }
+                base_url.set_port(self.redis_port).expect("Failed to set port");
+                base_url
+            }).map(|almost_finished_url| {
+                let mut base_url = almost_finished_url;
+                if self.redis_secure {
+                    println!("Yves should be secure");
```
sighphyre, line 74: We should probably remove this one!

sighphyre, line 74: Probably don't want this one

```suggestion
```

PR #204, opened 2023-06-06
Task/healthcheck subcommand

chriswk:
We've had requests from users/customers that use Edge with docker-compose. Since our docker image does not contain any other binary besides the edge binary itself, they were not able to use the `healthcheck` directive for their compose service. This PR adds a `health` subcommand to the edge binary which will perform the needed health check to be used in a docker-compose service declaration.


sighphyre (approved):
LGTM

PR #206, opened 2023-06-07
feat: Use service account to create client tokens

chriswk:
### What
This pull request configures Edge to accept a service-account token at startup. If there is a Service Account token present. Edge will use this to request a client-token upstream for frontend tokens which Edge does not already have coverage for with the known client tokens.

This should make it easier to start using Edge in production, as long as you're comfortable with allowing Edge access to creating tokens.

### Discussion points
Currently, SA tokens can only be scoped as Admin | Editor | Viewer, and it's only Admin that is allowed to create api tokens, so this would still require granting Edge full admin access and risking an Admin token leaking.

@nunogois is working on an improvement to Service Accounts | Root roles which should allow users to configure a root role with only access to creating API tokens of a specific type. This should increase the security and decrease the risk of leaking the SA token used in starting up Edge.


server/src/frontend_api.rs
```diff
@@ -768,10 +785,12 @@ mod tests {
                 "*:development.03fa5f506428fe80ed5640c351c7232e38940814d2923b08f5c05fa7",
             ))
             .to_request();
+        let r = test::call_service(&app, req).await;
+        debug!("{}", r.status());
+        /*        let result: FrontendResult = test::call_and_read_body_json(&app, req).await;
```
sighphyre, line 790: If we don't need it anymore, can we yeet it? Or is this something that's helpful to uncomment for testing the tests?

server/src/http/unleash_client.rs
```diff
@@ -171,6 +172,27 @@ impl UnleashClient {
...
+    ) -> Self {
+        Self {
+            urls: UnleashUrls::from_base_url(server_url),
+            backing_client: new_reqwest_client(
+                "unleash_edge".into(),
+                skip_ssl_verification,
+                client_identity,
+                upstream_certificate_file,
+            )
+            .unwrap(),
```
sighphyre, line 193: Just a double check - we expect this to be a safe unwrap? Or can we swap it for an expect?

chriswk: Yes, it should be, and if it crashes we do want the panic to happen.

sighphyre (approved):
LGTM! Very nice feature to have!

PR #209, opened 2023-06-09
Task/trust proxy

chriswk:
Allow you to set `trust-proxy` flag in order to use connection_info(), otherwise we will do peer_ip() to populate remote Address in our context. This has revealed an error in Yggdrasil (we believe).


sighphyre (approved):
LGTM

sighphyre:
> Allow you to set `trust-proxy` flag in order to use connection_info(), otherwise we will do peer_ip() to populate remote Address in our context. This has revealed an error in Yggdrasil (we believe).

Turned out not to be a bug for anyone who comes looking at a later point, just struggles with test setups

PR #210, opened 2023-06-12
chore: remove unneeded import

sighphyre:
Does what it says on the tin. Not actually sure how this one dodged rustc and Clippy


chriswk (approved):
LGTM

PR #215, opened 2023-06-23
chore: allows resolving a single toggle to do that directly instead of iterating the whole map

sighphyre:
Allows Edge to tap into `resolve` in Yggdrasil rather than `resolve_all` when checking a single toggle. This means Edge doesn't have to walk the entire hashmap of state to resolve a single toggle. This provides a modest performance improvement when working with thousands of toggles


gastonfournier (approved):
LGTM!

chriswk (approved):
Yes!

PR #220, opened 2023-06-28
perf: improve memory usage during request lifecycle

sighphyre:
## What

This inverts the way we do filtering of the feature toggles when resolving data out of our cache. Previously we'd materialise the entire response, cloning each. Now we build the filtering at a higher layer and pass the filter in to the method that resolves the toggles. This means that the filter is applied **before** the cloning.

## Why

We're a little inefficient with the number of clones we apply when resolving a toggle. This means we stack allocate a **lot** of unnecessary memory. This reduces this to one clone (two on one cold path of retrieving toggles when they're not present in the cache). 

All in all, this doubles the response speed when working with a large number of toggles and halves the memory usage to achieve that.

### Baseline benchmark

Summary:
  Total:	14.3805 secs
  Slowest:	10.0794 secs
  Fastest:	0.9730 secs
  Average:	5.5344 secs
  Requests/sec:	7.8579
  

Response time histogram:
  0.973 [1]	|■
  1.884 [4]	|■■■■
  2.794 [7]	|■■■■■■■
  3.705 [7]	|■■■■■■■
  4.616 [9]	|■■■■■■■■■
  5.526 [18]	|■■■■■■■■■■■■■■■■■■
  6.437 [39]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  7.347 [10]	|■■■■■■■■■■
  8.258 [13]	|■■■■■■■■■■■■■
  9.169 [1]	|■
  10.079 [4]	|■■■■

Latency distribution:
  10% in 2.8589 secs
  25% in 4.7205 secs
  50% in 5.8351 secs
  75% in 6.4597 secs
  90% in 7.4248 secs
  95% in 8.9751 secs
  99% in 10.0794 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0015 secs, 0.9730 secs, 10.0794 secs
  DNS-lookup:	0.0004 secs, 0.0000 secs, 0.0016 secs
  req write:	0.0001 secs, 0.0000 secs, 0.0003 secs
  resp wait:	3.2266 secs, 0.0585 secs, 7.3207 secs
  resp read:	1.8054 secs, 0.0001 secs, 6.3262 secs

Status code distribution:
  [200]	107 responses
  [408]	6 responses

### Baseline memory usage:

After 1 prewarm up strike and benchmarks run: 4.0 GB

### Benchmark after patch
Summary:
  Total:	12.3002 secs
  Slowest:	6.2325 secs
  Fastest:	1.0501 secs
  Average:	2.9045 secs
  Requests/sec:	15.6909
  

Response time histogram:
  1.050 [1]	|■
  1.568 [1]	|■
  2.087 [5]	|■■■
  2.605 [65]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  3.123 [73]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  3.641 [30]	|■■■■■■■■■■■■■■■■
  4.160 [7]	|■■■■
  4.678 [5]	|■■■
  5.196 [3]	|■■
  5.714 [2]	|■
  6.232 [1]	|■

Latency distribution:
  10% in 2.3857 secs
  25% in 2.4603 secs
  50% in 2.8073 secs
  75% in 3.1863 secs
  90% in 3.5532 secs
  95% in 4.3718 secs
  99% in 6.2325 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0003 secs, 1.0501 secs, 6.2325 secs
  DNS-lookup:	0.0001 secs, 0.0000 secs, 0.0009 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0004 secs
  resp wait:	1.7042 secs, 0.3664 secs, 3.4124 secs
  resp read:	1.1997 secs, 0.1047 secs, 3.1137 secs

Status code distribution:
  [200]	193 responses

### Memory usage after patch

After 1 prewarm up strike and benchmarks run: 2.2GB

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -1,11 +1,11 @@
 use crate::error::{EdgeError, FeatureError};
 use crate::filters::{
-    filter_client_features, name_prefix_filter, project_filter, FeatureFilterSet,
+    filter_client_features, name_match_filter, name_prefix_filter, project_filter, FeatureFilterSet,
 };
 use crate::http::feature_refresher::FeatureRefresher;
 use crate::metrics::client_metrics::MetricsCache;
 use crate::tokens::cache_key;
-use crate::types::{EdgeJsonResult, EdgeResult, EdgeToken, FeatureFilters, ProjectFilter};
+use crate::types::{EdgeJsonResult, EdgeResult, EdgeToken, FeatureFilters};
 use actix_web::web::{self, Data, Json, Query};
 use actix_web::{get, post, HttpRequest, HttpResponse};
 use dashmap::DashMap;
@@ -127,25 +127,24 @@ pub async fn get_feature(
         .get(&edge_token.token)
         .map(|e| e.value().clone())
         .ok_or(EdgeError::AuthorizationDenied)?;
-    let client_features = match req.app_data::<Data<FeatureRefresher>>() {
-        Some(refresher) => refresher.features_for_token(validated_token).await,
+
+    let filter_set = FeatureFilterSet::from(Box::new(name_match_filter(feature_name.clone())))
+        .with_filter(project_filter(&edge_token));
+
+    match req.app_data::<Data<FeatureRefresher>>() {
+        Some(refresher) => {
+            refresher
+                .features_for_filter(validated_token.clone(), filter_set)
+                .await
+        }
         None => features_cache
             .get(&cache_key(&edge_token))
-            .map(|features| features.value().clone())
-            .map(|client_features| ClientFeatures {
-                features: client_features
-                    .features
-                    .filter_by_projects(&validated_token),
-                ..client_features
-            })
+            .map(|client_features| filter_client_features(&client_features, filter_set))
             .ok_or(EdgeError::ClientFeaturesFetchError(FeatureError::Retriable)),
-    }?;
-    client_features
-        .features
-        .into_iter()
-        .find(|feature| feature.name == feature_name.clone())
-        .map(Json)
-        .ok_or(EdgeError::FeatureNotFound(feature_name.into_inner()))
+    }
+    .map(|client_features| client_features.features.into_iter().next())?
+    .ok_or(EdgeError::FeatureNotFound(feature_name.into_inner()))
+    .map(Json)
 }
 
 #[utoipa::path(
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -1,5 +1,7 @@
 use crate::error::{EdgeError, FeatureError};
-use crate::filters::{name_prefix_filter, project_filter, FeatureFilterSet};
+use crate::filters::{
+    filter_client_features, name_prefix_filter, project_filter, FeatureFilterSet,
+};
 use crate::http::feature_refresher::FeatureRefresher;
 use crate::metrics::client_metrics::MetricsCache;
 use crate::tokens::cache_key;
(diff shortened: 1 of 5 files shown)


chriswk (approved):
After the pair review, I really like this idea. And the numbers look good.

server/src/client_api.rs
```diff
@@ -65,44 +68,39 @@ async fn resolve_features(
...
+    } else {
+        FeatureFilterSet::default()
+    }
+    .with_filter(project_filter(&edge_token));
+
+    let client_features = match req.app_data::<Data<FeatureRefresher>>() {
+        Some(refresher) => {
+            refresher
+                .features_for_filter(validated_token.clone(), filter_set)
+                .await
```
gastonfournier, line 92: I feel that this call to the refresher should be equal to calling `features_cache.get(&cache_key(&edge_token))` (because it would populate the cache on the specific same key). Then mapping and filtering should happen independently of the path we took to get the client_features.

Something like:
```
if (I have to refresh) 
   client_features = get from the refresher
else 
   client_features = get from cache
client_features.map(|client_features| filter_client_features(&client_features, filter_set))
            .ok_or(EdgeError::ClientFeaturesFetchError(FeatureError::Retriable))
```

So I'm a bit confused because the refresher also accesses a features_cache (maybe a different one?)

server/src/http/feature_refresher.rs
```diff
@@ -172,14 +177,14 @@ impl FeatureRefresher {
...
+        filters: FeatureFilterSet,
+    ) -> Option<ClientFeatures> {
         self.features_cache
             .get(&cache_key(token))
-            .map(|e| e.value().clone())
-            .map(|features_response| ClientFeatures {
-                features: features_response.features.filter_by_projects(token),
-                ..features_response
-            })
+            .map(|client_features| filter_client_features(&client_features, filters))
```
gastonfournier, line 187: Ok, I think the answer to my comment above lies in this method. We're cloning the list inside. I'm just wondering if we could return a reference to the features instead of a clone, and then let the client create the clone if they need to.

The current way is a bit safer, but if we're going to go down the optimization path might be something to consider (although not for this PR)

gastonfournier (approved):
Looks good, I think I've answered my own comment and that could potentially be something for future improvement. This looks good!

server/src/client_api.rs
sighphyre: Yeah the code is a little confusing here, this needs a little love in the near future perhaps. What's happening here is that the `feature_refresher` holds its own internal cache in `edge` mode and if you're in `offline` mode you use the other cache.

The filtering on demand is a lot harder to do with our architecture, unfortunately, this is us not being kind to Rust's memory model, but ultimately, it's hard to return references to the `client_features` without massively violating the lifetime rules. I would like to get to your suggestion but without a large overhaul it's tough.

Needs some refactor in a future body of work I think

server/src/http/feature_refresher.rs
sighphyre: Probably not sadly, this would require us to marry the lifetime of the DashMap data to the lifetime of the features we want to return. Because our dashmap is currently a self reference that lifetime is already has some hefty constraints on it

I don't think this matters that much, though, the data in dashmap is borrowed by nature and the data we require to send over HTTP needs to be an owned reference so at least one clone is absolutely necessary

PR #221, opened 2023-06-28
chore: reduce unnecessary public api surface

sighphyre:
This narrows the access modifiers for a number of functions and structs that make no sense to be public. This is a nuisance because internal changes become major/minor version changes for no good reason

### DIFF
--- a/server/src/http/unleash_client.rs
+++ b/server/src/http/unleash_client.rs
@@ -476,6 +476,15 @@ mod tests {
 
     use super::{EdgeTokens, UnleashClient};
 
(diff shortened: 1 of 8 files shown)


server/src/http/unleash_client.rs
```diff
@@ -476,6 +476,15 @@ mod tests {
 
     use super::{EdgeTokens, UnleashClient};
 
+    impl ClientFeaturesRequest {
```
sighphyre, line 479: Test only code. So it got moved into the relevant test module

PR #226, opened 2023-07-04
chore: switch docker publish role

gardleopard:
Switching to new ecr publish role that is managed by code.


chriswk (approved):
Nice. Thank you

PR #228, opened 2023-07-12
feat: added timing for client feature fetch operations

chriswk:
Realised we were only logging fetch failures, not fetches that succeeded in any way. This PR adds a HistogramVec for timing all client feature fetch requests per http status.


sighphyre (approved):
LGTM!

PR #241, opened 2023-09-07
fix: make fe tokens respect token cache

sighphyre:
Makes the check for incoming front end tokens correctly compare against present tokens so that multiple API calls are not made

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -22,6 +22,17 @@ use crate::{
     types::{ClientFeaturesRequest, ClientFeaturesResponse, EdgeToken, TokenRefresh},
 };
 
+fn frontend_token_is_covered_by_tokens(
+    frontend_token: &EdgeToken,
+    tokens_to_refresh: Arc<DashMap<String, TokenRefresh>>,
+) -> bool {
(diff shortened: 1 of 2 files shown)


server/src/tokens.rs
```diff
@@ -311,4 +311,80 @@ mod tests {
 
         assert_eq!(actual, expected);
     }
+
```
sighphyre, line 314: These aren't all that relevant but they were missing tests and nice to add while we're in this space

chriswk (approved):
LGTM

server/src/http/feature_refresher.rs
```diff
@@ -22,6 +22,17 @@ use crate::{
...
 };
 
+fn frontend_token_is_covered_by_tokens(
+    frontend_token: &EdgeToken,
+    tokens_to_refresh: Arc<DashMap<String, TokenRefresh>>,
+) -> bool {
+    tokens_to_refresh.iter().any(|client_token| {
+        client_token
+            .token
+            .same_environment_and_broader_or_equal_project_access(frontend_token)
```
sighphyre, line 32: This is the magic, this flips fe token and present tokens - we previously did the check the wrong way around

PR #242, opened 2023-09-07
feat: add hot reloading and an optional, simpler file format to offline mode

sighphyre:
This adds some nice tooling for local development in offline mode. That's two things, firstly the user can now specify a reload interval, which will cause Edge to reload the offline file every X seconds. The second is a simpler format to specify the offline bootstrap in. That's a very, very simplified format, which Edge maps into a full ClientFeatures response matching the expected set properties. Together these allow the user to fiddle with toggles when in offline mode easily so that they can force their SDKs into the states that they want

Almost all of this code does not touch `edge` mode, which is primarily what production environments use. It's also all opt in. So I think this is relatively safe to merge


chriswk (approved):
Love the added simple format here.

PR #243, opened 2023-09-08
docs: escapes the token format in the docs

sighphyre:
Escapes the token format in the docs. This seems to break our documentation generation on Unleash when a fresh build is done


nunogois (approved):

chriswk (approved):
LGTM

PR #250, opened 2023-09-11
feat: added token info endpoint

chriswk:
In order to facilitate debug session, this adds an /internal-backstage/tokens endpoint, which exposes all tokens and the current state that Edge knows about as well as which tokens Edge is refreshing data for.


sighphyre (approved):
LGTM

PR #252, opened 2023-09-12
feat: add /internal-backstage/ready endpoint

chriswk:


gardleopard:
I'm thinking about lazy loading here. If using ready endpoint, lazy loading won't work. Maybe we should communicate that somehow?

server/src/internal_backstage.rs
```diff
@@ -6,8 +6,10 @@ use actix_web::{
     get,
     web::{self, Json},
 };
-use serde::Serialize;
-#[derive(Debug, Serialize)]
+use dashmap::DashMap;
+use serde::{Deserialize, Serialize};
+use unleash_types::client_features::ClientFeatures;
+#[derive(Debug, Serialize, Deserialize)]
 pub struct EdgeStatus {
```
sighphyre, line 13: Might be time for the inner type to be an enum here. Should be okay to do something like 

enum Status {
  Ok,
  Ready,
  NotReady
}

Small suggestion, I'm okay for this to go without it

sighphyre (approved):
LGTM

PR #253, opened 2023-09-12
feat: more information in error logs

chriswk:
This adds more output to our error messages when we fail to validate tokens as well as when we receive 401 or 404 from client features endpoints.

### DIFF
--- a/server/src/error.rs
+++ b/server/src/error.rs
@@ -91,7 +91,7 @@ pub enum EdgeError {
     ClientBuildError(String),
     ClientCertificateError(CertificateError),
     ClientFeaturesFetchError(FeatureError),
-    ClientFeaturesParseError,
+    ClientFeaturesParseError(String),
     ClientRegisterError,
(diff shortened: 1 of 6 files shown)


server/src/error.rs
```diff
@@ -132,12 +134,16 @@ impl Display for EdgeError {
                     f,
                     "Could not fetch client features because api key was not allowed"
                 ),
+                FeatureError::NotFound => write!(
+                    f,
+                    "Could not fetch features because upstream url was not found"
```
nunogois, line 139: Could this one be a bit misleading between "No upstream URL was set" vs "We got a 404 back from the the upstream URL that you set"?

nunogois (approved):

server/src/error.rs
chriswk: No upstream URL is not possible. Clap handles the fact that you have to set an upstream URL if you're running in Edge mode

nunogois: Yep, I suspected that would be the case 👍

PR #261, opened 2023-09-13
feat: docker debug container

gardleopard:
This adds a docker image that has some tools in it, the intent of this is to make it easier to debug edge.


chriswk (approved):
Nice. Let's try this

PR #264, opened 2023-09-13
fix: allow startup tokens to continue to validate against upstream

sighphyre:
Forces Edge to continue to validate startup tokens until they succeed. This fixes an issue where if Edge couldn't validate the startup tokens against Unleash, it would discard them and never try again. This does make the assumption that Edge should assume the startup tokens are valid and continue to check until it gets a response from Unleash. While the code looks a little icky, the loop becomes a no-op on a background thread once the tokens have been validated and no further calls will occur. If the tokens provided for startup are not valid, Edge will log that in the loop. I think that's correct, since the tokens _should_ be correct, or the user should get a strong warning that the config is broken. If the tokens can be validated during Edge's startup sequence, then the loop will never make HTTP calls

### DIFF
--- a/server/src/auth/token_validator.rs
+++ b/server/src/auth/token_validator.rs
@@ -1,3 +1,4 @@
+use crate::http::feature_refresher::FeatureRefresher;
 use crate::http::unleash_client::UnleashClient;
 use crate::persistence::EdgePersistence;
 use crate::types::{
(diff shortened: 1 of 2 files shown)


gardleopard (approved):
Nice !

chriswk (commented):
What's the motivation for this one? If a token can't be validated on the first try, why would it succeed the second?

I'd be more inclined to accept this if we actually considered the reason for why the token failed to validate first. Anything in the 40x range won't change even if we try again. If there's an actual client error I seem to remember we already retry. Let's have a chat.

sighphyre:
>  If there's an actual client error I seem to remember we already retry. 
Turns out we don't. Not for startup tokens at least. They get discarded

For network errors, anything in the 500 range is valid to retry quietly until it succeeds. For anything else, logging angrily is okay but not great.

PR #267, opened 2023-09-15
feat: Added ready subcommand to cli

chriswk:
So, it's nice to have the /internal-backstage/ready endpoint, but since we don't include curl in the docker image, it makes it hard to check ready status when running docker; since both ready and not ready returns 200, we need to parse the body of the response. This PR adds a ready subcommand which does a http request to the /internal-backstage/ready endpoint of a running edge instance.


sighphyre (approved):
The logic looks fine to me but the unreachable branches are getting a bit noisy at this point. There's a bit too much for a code suggestion here so I did this: https://github.com/Unleash/unleash-edge/pull/269/files I think we should consider merging that into this first

Can live without it though

PR #269, opened 2023-09-15
small clean ups of unreachable branches

sighphyre:
Our unreachable code is getting a bit noisy. If we trust rustc and the tests, we should be able to do this without trouble


chriswk (approved):
Sure

Issue #268, opened 2023-09-15
Improve documentation for the edge offline mode


DAGRSAG:
### Describe the feature request

It would be very nice, if you can extend your documentation for the edge offline mode. With the available documentation it was not so easy to get it running. My two main issues are:

## Reload config from json

> To make local development easier, you can specify a reload interval in seconds; this will cause Edge to reload the features file from disk every X seconds. This can be useful for local development.

Not sure if I have done an error on that (see also #265) but I was not able to get it running, that the flags are reloaded without any restart

## Environment/Project handling

> When using offline mode you must specify one or more tokens at startup. These tokens will let your SDKs access Edge. Tokens following the Unleash API format [project]:[environment].secret allow Edge to recognize the project and environment specified in the token, returning only the relevant features to the calling SDK

The question for me is, is it possible to have more than one project / environment for project in this mode? If yes, how to configure those in the json file

And furthermore I tried to use `default.development.my-token` but get a 403 error. If I only use the `my-token`, without any further information, the authentication against edge worked well. (see also https://github.com/Unleash/unleash-edge/issues/266#issuecomment-1721297590)

## Value for `toggle.instance-id`

From my experience, it doesn't matter what or if anything is specified here. However, I searched at the beginning to know if and if so what, I must specify here. I could have saved that if I had known it didn't matter 😄 

With `instance-id` I mean the setting `instanceId` for the sdk.

chriswk:
Hi @DAGRSAG - thanks for your effort in making this issue. Updates to docs coming this week. :+1:

Issue #266, opened 2023-09-15
Metrics request leads to an issue


DAGRSAG:
### Describe the bug

Hi,

I have started the unleash-edge in offline mode with the following command:

`docker run -v LOCAL_PATH:/edge/data -p 3063:3063 -e BOOTSTRAP_FILE=/edge/data/development.json -e RELOAD_INTERVAL=5 -e TOKENS=my-secret-123,other-secret-456 unleashorg/unleash-edge:v8.0.1 offline`

Toggles get loaded from json file. But at any point in time, the client (java  sdk) sends a POST request to edge to update metrics. The offline mode do not have metrics at all, which is fine.

But in my Spring Boot application as well in the log of the unleash-edge is an error visible:

docker log from unleash-edge:
![image](https://github.com/Unleash/unleash-edge/assets/74786365/96ae9d1f-bad2-46d6-82a9-76af98a78f99)

Log from Spring boot app:
![image](https://github.com/Unleash/unleash-edge/assets/74786365/674ddaf7-ebce-437e-a6f4-6f469cdde476)

The client doesn't know that the call is send against edge in offline but, but why the edge throws this error? Wouldn't it be possible to just ignore metrics call from sdk?

### Expected behavior

No exception for metrics call from SDK

### Unleash version

5.1.8

### Subscription type

Open source

### Hosting type

Self-hosted

### SDK information (language and version)

unleash-client-java:7.1.0

chriswk:
Hi @DAGRSAG - Thanks for the report. Particularly the stack trace from Edge. It caused me to realise what we've done here. All tokens that can be used in Edge mode has an environment, however, in offline mode this is not a guarantee; so I'll update the code to not just assume the environment is set. I'll make a patch on Monday. I'm about to head home for the weekend.

DAGRSAG:
Hi @chriswk. Wow thanks for the fast answers and also for taking care.

Please allow me a question.
In the documentation, I have seen the following: 

> When using offline mode you must specify one or more tokens at startup. These tokens will let your SDKs access Edge. Tokens following the Unleash API format [project]:[environment].<somesecret> allow Edge to recognize the project and environment specified in the token, returning only the relevant features to the calling SDK

But to be honest, I was not successful to use one of the tokens with project and environment. That is the reason, why I just used the token without any further information.

How can I use the token with project (I assume the value from the field within the json file) and the environment? Here is an example from my current json file

```json
{
    "strategies": [],
    "impressionData": false,
[...]

Issue #162, opened 2023-04-19
TLS documentation


sebastian-bury:
### Describe the feature request

Add details on the docs for: 

- --tls-server-key needing to be a file, not text
- supported file formats

PR #270, opened 2023-09-18
fix: Update docs about environment in offline mode

chriswk:
fix: Fall back to default 'development' environment value if token does not contain an environment

fixes: #266
fixes: #268
fixes: #162


sighphyre (approved):
G

PR #271, opened 2023-09-19
Quote angle brackets; fix readme

thomasheartman:
This surrounds angle bracketed content in backticks, so that markdown renders it as expected instead of parsing it as html tags. The wrong parsing also causes our doc build to fail.

It currently renders as nothing on github too:
image


sighphyre (approved):
LGTM!

PR #272, opened 2023-09-20
DO NOT MERGE: Testing if organising features in ENV -> PROJECT -> FEATURES is easier

gardleopard:
Testing if organising features in ENV -> PROJECT -> FEATURES is easier to work with, compared to ENV -> FEATURES.

I think it is, but it needs more work to be efficient. I think the old way of storing should be removed in favour of this way. The way of returning toggles should be changed to select by project and environment instead of selecting by environment and then filtering by projects.

This solves the problem of archived toggles not dissapearing from the features endpoint.

This is just an example to look at, nothing to merge here


sighphyre:
We've gone with a different approach. At least for now

Open to a conversation around moving the data representation but there's more to consider than just this fix

PR #275, opened 2023-09-26
fix(#274): use the validated token to calculate flags to return

### LINKED ISSUE #274: Client features filtering doesn't work correctly when tokens have multiple projects
### Describe the bug

Given two tokens A and B, where at least A is a multi-project token and where B has access to at least one project that A doesn't:

If you make a request with token A, then with token B, and then with token A again, the response for the last request will also contain flags that should only be visible to token B.

Similarly, if you first make a request with token B and then one with token A, everything appears to be in order. But if you make another request with token A, the response to that second request will also include the flags that should only be delivered to B.

Setup

1. In Unleash, create 3 different projects. Add a flag in each project.
2. Generate two different client API keys, one with access to projects 1 and 2, and one with access to project 3.

Via edge: make requests to the `api/client/features` endpoint
1. Using token A. Observe that you only get flags from projects 1 and 2
2. Using token B: Observe that you only get flags from project 3
3. Using token A: Observer that you get flags from all three projects.

Similarly, for case 2, it works like this:
1. Using token B: Observe that you only get flags from project 3
2. Using token A. Observe that you only get flags from projects 1 and 2
3. Using token A again: Observe that you get flags from all three projects.

### Expected behavior

When using a client token that only has access to a specified list of projects, I'd expect to only ever receive flags related to those projects.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

It seems likely that the issue is related to the [project filter code](https://github.com/Unleash/unleash-edge/blob/main/server/src/client_api.rs#L86).

### Unleash version

_No response_

### Subscription type

None

### Hosting type

None

_No response_

thomasheartman:
This change fixes a bug where the client API would return all flags
that existed in the cache, even if the api token did not have access
to those flags. Crucially, the API token had to have access to
multiple (but not all) projects for this to happen.

The root cause is that we used the incoming edge token to check which
flags to return. Before it gets validated, its `projects` property is
just an empty list. In the filtering, this causes edge to return all
available. Features.

The solution was to instead use the validated edge token that we
create further up.

There is also a test that confirms this behavior is what we expect.

## Discussion point

Could we make it so that the `with_filter` function can only take
validated edge tokens or would that break something else? Might be a
good way to future proof it.

---

Closes #274

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -635,7 +635,6 @@ mod tests {
         token_a.projects = vec!["dx".into(), "eg".into()];
         token_a.status = TokenValidationStatus::Validated;
         token_a.token_type = Some(TokenType::Client);
-        println!("Created token A: {:?}", token_a);
         token_cache.insert(token_a.token.clone(), token_a.clone());
 
         let mut token_b =
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -129,7 +129,7 @@ pub async fn get_feature(
         .ok_or(EdgeError::AuthorizationDenied)?;
 
     let filter_set = FeatureFilterSet::from(Box::new(name_match_filter(feature_name.clone())))
-        .with_filter(project_filter(&edge_token));
+        .with_filter(project_filter(&validated_token));
 
     match req.app_data::<Data<FeatureRefresher>>() {
         Some(refresher) => {
@@ -138,7 +138,7 @@ pub async fn get_feature(
                 .await
         }
         None => features_cache
-            .get(&cache_key(&edge_token))
+            .get(&cache_key(&validated_token))
             .map(|client_features| filter_client_features(&client_features, filter_set))
             .ok_or(EdgeError::ClientFeaturesFetchError(FeatureError::Retriable)),
     }
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -83,7 +83,7 @@ async fn resolve_features(
     } else {
         FeatureFilterSet::default()
     }
-    .with_filter(project_filter(&edge_token));
+    .with_filter(project_filter(&validated_token));
 
     let client_features = match req.app_data::<Data<FeatureRefresher>>() {
         Some(refresher) => {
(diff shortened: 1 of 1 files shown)


server/src/client_api.rs
```diff
@@ -92,7 +92,7 @@ async fn resolve_features(
                 .await
         }
         None => features_cache
-            .get(&cache_key(&edge_token))
+            .get(&cache_key(&validated_token))
```
thomasheartman, line 95: I don't think this line actually matters, as I'm guessing that only the actual token value gets used. So we could remove this if you want to. I just changed it to make it clearer that we're done with the actual edge token at this point.

server/src/client_api.rs
```diff
@@ -129,7 +129,7 @@ pub async fn get_feature(
         .ok_or(EdgeError::AuthorizationDenied)?;
 
     let filter_set = FeatureFilterSet::from(Box::new(name_match_filter(feature_name.clone())))
-        .with_filter(project_filter(&edge_token));
+        .with_filter(project_filter(&validated_token));
```
thomasheartman, line 132: I don't know, but I suspect this line would suffer from the same problem?

thomasheartman: Checking it out at the moment

thomasheartman: Can confirm that this does fix another bug, but there's more issues with it, so I want to add a test in a new PR.

chriswk: I think you're correct, this has been a lesson in misplaced assumptions. Thank you for fixing it @thomasheartman :clap:

chriswk (approved):
Nice. I now want to meet "biff the magic flagon". :)

server/src/client_api.rs
thomasheartman: The other issue floating around is here: https://github.com/Unleash/unleash-edge/issues/276

server/src/client_api.rs
```diff
@@ -617,6 +617,59 @@ mod tests {
...
+                .service(web::scope("/api/client").service(get_features)),
+        )
+        .await;
+
+        let mut token_a =
+            EdgeToken::try_from("[]:production.puff_the_magic_dragon".to_string()).unwrap();
+        token_a.projects = vec!["dx".into(), "eg".into()];
+        token_a.status = TokenValidationStatus::Validated;
+        token_a.token_type = Some(TokenType::Client);
+        println!("Created token A: {:?}", token_a);
```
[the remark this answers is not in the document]
thomasheartman: Yeah, I missed this, sorry! Good catch 😄

Issue #273, opened 2023-09-21
Archived features remains in edge


gardleopard:
### Describe the bug

When a feature toggle is archived it does not get removed from edge.

### Steps to reproduce the bug

1. connect edge to an unleash instance
2. fetch feature toggles from edge
3. archive a feature toggle in unleash
4. fetch feature toggles from edge, observe that the feature is still present

### Expected behavior

The archived feature is removed from edge. 

### Additional context

This draft pr has a mitigation of the issue: https://github.com/Unleash/unleash-edge/pull/272

### Subscription type

None

### Hosting type

None

PR #277, opened 2023-09-26
initial draft of refreshing projects

chriswk:
Previously, we made a best effort to keep all known data when an update
came in. Realising that the Unleash API currently does not allow for
partial project tokens. Any refreshed data for a single project, can
drop the existing project and replace it with the incoming

fixes: #273


sighphyre (approved):
LGTM

Issue #276, opened 2023-09-26
Single-flag requests sometimes give wrong results


thomasheartman:
### Describe the bug

Given api tokens A and B where A has access to project 1 and B has access to project 2. There is a feature called `edge-flag-1` in project 1 and one called `edge-flag-3` in project 2.

If I try to get /api/client/features/edge-flag-3 with token B, I get `edge-flag-3` as expected.

If I try the same URL with token A, I instead get `edge-flag-1`.

If I try the same URL with token A **again,** I get an error, saying it failed to find the flag.

### Steps to reproduce the bug

1. In Unleash, create 2 different projects. Add a flag in each project.
2. Generate two different client API keys, one with access to projects 1, and one with access to project 2.
3. Create a feature in each of the projects (in my instance, their names are `edge-flag-1` and `edge-flag-3` respectively).

Via edge: make requests to the `api/client/features/edge-flag-3` endpoint
1. Using token B. You should get the right flag back.
2. Using token A. Observe that you get `edge-flag-1` back, even if the URL says `edge-flag-3`.
3. Using token A again: You get an error. (as expected)

### Expected behavior

You should never get data for a different flag than the one in the URL.

### Additional context

Discovered when working on #274.

### Subscription type

None

### Hosting type

None

PR #283, opened 2023-09-27
fix: No longer return wrong feature toggle

chriswk:
Previously, upon first access with a new token which Edge hadn't seen before our rehydration code only filtered by project of the token, so once it had the data it would stupidly return the token that was available for the new token (in the wrong project).

After this refactor/fix we now apply all filters to all queries for feature toggles from the cache, which should reduce the combinations of possible filter mistakes.

fixes: #276

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -16,7 +16,7 @@ use unleash_yggdrasil::EngineState;
 
 use super::unleash_client::UnleashClient;
 use crate::error::{EdgeError, FeatureError};
-use crate::filters::{filter_client_features, project_filter, FeatureFilterSet};
+use crate::filters::{filter_client_features, FeatureFilterSet};
 use crate::types::{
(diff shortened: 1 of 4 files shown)


server/src/http/feature_refresher.rs
```diff
@@ -190,15 +190,11 @@ impl FeatureRefresher {
         frontend_token_is_covered_by_tokens(frontend_token, self.tokens_to_refresh.clone())
     }
 
-    pub(crate) async fn register_and_hydrate_token(
-        &self,
-        token: &EdgeToken,
-    ) -> EdgeResult<ClientFeatures> {
-        let filter = FeatureFilterSet::from(project_filter(token));
+    /// This method
```
sighphyre, line 193: Yes?

sighphyre: Yeeted xD

sighphyre (approved):
LGTM

PR #286, opened 2023-09-30
chore: update DATABASE_URL to use the database created via POSTGRES_D…

snosratiershad:
…B in compose

## About the changes
Same as https://github.com/Unleash/unleash/pull/4836

### DIFF
--- a/examples/docker-compose.yml
+++ b/examples/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       retries: 5
       start_period: 10s
   edge:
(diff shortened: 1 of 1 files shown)


snosratiershad:
I've also noticed it uses version 6.0.0 of unleash-edge, which needs to be updated. Just updated it to 13.0.0

daveleek (approved):
Thank you for your contribution, this looks good to me!

chriswk (approved):
Thank you for this.

PR #292, opened 2023-10-03
chore(deps): Start work upgrading to opentelemetry 0.20

chriswk:
Seems to be working the same way the previous one did. We should consider setting up a telemetry feature here.


sighphyre (approved):
Looks like plumbing and the tests are green so LGTM

PR #300, opened 2023-10-10
fix: update Yggdrasil

chriswk:
Previously, yggdrasil returned a disabled variant if the strategy variants representation came back as an empty list instead of null. With Yggdrasil 0.6 this is now fixed.


gastonfournier (approved):
Looks good!

PR #302, opened 2023-10-12
chore: bump utoipa and unleash types

sighphyre:
Does what it says on the tin, this doesn't alter any features but it does mean client SDKs that support dependent features work correctly against Edge


chriswk (approved):
LGTM

PR #304, opened 2023-10-13
fix: move etag middleware to last in chain to ensure it gets added in…

daveleek:
## About the changes

fix: move etag middleware to last in chain to ensure it gets added In gziped responses 

Closes #
https://github.com/Unleash/unleash-client-dotnet/issues/165


sighphyre (approved):
Nice catch!

PR #313, opened 2023-10-24
feat: make edge log bad request information from upstream

chriswk:
We have one customer that is reporting that their Edge installation is getting 400 from upstream when reporting batch metrics. This PR extends the data we keep when Edge receives a 400 from upstream batch endpoint and tries to parse the message that comes back from ajv.

It should make it easier to debug why Edge is unable to format metrics correctly.


sighphyre (approved):
LGTM

PR #322, opened 2023-11-01
feat: Add support for setting log format at startup

chriswk:
We've had a customer request to be able to change log formatting. This PR adds support for plain (what we currently have), pretty (a bit friendlier format) and json 
```json
{"timestamp":"2023-11-01T09:00:16.007926Z","level":"DEBUG","fields":{"message":"No update needed. Will update last check time with \"537b2ba0:127766\""},"target":"unleash_edge::http::feature_refresher"}
```


sighphyre (approved):
LGTM

Issue #323, opened 2023-11-01
Cache refresh does not respect deleted/archived projects


chriswk:
### Describe the bug

It seems that the cache refresher is not able to remove features belonging to a deleted/archived project.

### Steps to reproduce the bug

* Create a new project
* Add a new toggle
* Setup any strategy
* Archive the toggle
* Delete the project

### Expected behavior

I was expecting that since the project is deleted, it would remove the toggle name, strats etc from edge but that is not the case here.

### Unleash version

Edge 16.0.0

### Subscription type

None

### Hosting type

None

PR #324, opened 2023-11-01
fix: Handle archived/deleted projects

chriswk:
Previously, our cache refresh algorithm assumed that the response from upstream contained all projects we wanted to do updates to. Wayfair correctly reported this breaking their opportunity to archive/delete projects, since the cache would still contain deleted projects.

This patch updates Edge to use the projects the token has access to decide whether or not to keep the elements in cache.

New flow:
1. Fetch projects from token
2. Filter out all features belonging to these projects
3. Extend remaining list with update from response
4. Return extended list.

fixes: #323

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -37,8 +37,13 @@ fn frontend_token_is_covered_by_tokens(
     })
 }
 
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
LGTM!

PR #330, opened 2023-11-08
fix: Wildcard token refreshing.

chriswk:
Previously our refreshing algorithm assumed that we always had at least one project with an explicit name for deciding what to keep. When in reality we could have a wildcard token (*) for updating. This means that we should just return the update as our new set of data for this token.

Added another test with a wildcard token to verify that we do indeed only keep the update.

In addition, added a filter for unique feature names to what we're keeping to avoid mixing multiple features of the same name.

fixes: #331

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -4,8 +4,6 @@ use std::{sync::Arc, time::Duration};
 use actix_web::http::header::EntityTag;
 use chrono::Utc;
 use dashmap::DashMap;
-use itertools::Itertools;
-use tracing::log::trace;
 use tracing::{debug, warn};
 use unleash_types::client_features::Segment;
 use unleash_types::client_metrics::ClientApplication;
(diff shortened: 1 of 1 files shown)


server/src/http/feature_refresher.rs
```diff
@@ -80,16 +81,20 @@ fn update_projects_from_feature_update(
...
+            let mut to_keep: Vec<ClientFeature> = original
+                .iter()
+                .filter(|toggle| {
+                    let p = toggle.project.clone().unwrap_or_else(|| "default".into());
+                    !projects_to_update.contains(&p)
+                })
+                .cloned()
+                .collect();
+            to_keep.extend(updated.iter().cloned());
+            to_keep.iter().unique_by(|cf| &cf.name).cloned().collect()
```
chriswk, line 96: Don't think this is needed, we should now be deleting every project the incoming token has access to, which means extending with the update should be just fine without eliminating duplicates.

chriswk: I'll remove it again.

sighphyre (approved):
LGTM

PR #337, opened 2023-11-13
fix: upgrades yggdrasil to 0.8.0, this fixes an issue with quotation marks in constraints

sighphyre:


server/src/frontend_api.rs
```diff
@@ -168,7 +168,7 @@ fn post_all_features(
     let engine = engine_cache.get(&key).ok_or_else(|| {
         EdgeError::FrontendNotYetHydrated(FrontendHydrationMissing::from(&edge_token))
     })?;
-    let feature_results = engine.resolve_all(&context_with_ip).unwrap();
+    let feature_results = engine.resolve_all(&context_with_ip, &None).unwrap();
```
sighphyre, line 171: All of these changes are just to deal with the API change in Yggdrasil. This is the hook for custom strategies, that we're explicitly ignoring

chriswk (approved):
LGTM

PR #339, opened 2023-11-16
feat: obey http status responses. backoff when 429 or 50x

chriswk:
Same thing here as in node and Java. In order to respect our upstream servers trying to tell us to chill out, this PR adds handling for 429, 500, 502, 503, 504. Edge already handled 401,403 and 404, so didn't need to change anything for that.


sighphyre (approved):
Yeah, this looks good to me

PR #351, opened 2023-11-27
fix: readd http metrics

chriswk:
After upgrading opentelemetry we lost our http metrics on our prometheus metrics dump due to a different call setup for getting it to work. This PR again sets up our configured meter provider to be used for all prometheus metrics. Which allows us to have our custom middleware work again.


gardleopard (approved):
LGTM

Issue #353, opened 2023-11-29
unleash-edge returns `null` for strategy variants, causing issues for Ruby SDK


acr92:
### Describe the bug

Unleash-edge modifies the response from Unleash, and sets the `variants` field in `strategies` to null if it's not specified. The response from Unleash omits the field if it's not set.

This causes an issue with the Ruby SDK, where `resolve_variants` only handles the case where the field is omitted, but not when the field is null ( https://github.com/Unleash/unleash-client-ruby/blob/b513382a176805962988a2ca9b0cd4692a08a4b2/lib/unleash/feature_toggle.rb#L205C7-L205C37 )

```
irb(main):003:0> {"variants" => nil}.fetch("variants")
=> nil
irb(main):004:0> {"variants" => nil}.fetch("variants", [])
=> nil
irb(main):005:0> {"variants" => nil}.fetch("cake", [])
=> []
```

### Expected behavior

I expect the response from Unleash to be the same, or the Unleash SDKs to handle the situations where unleash-edge does something different than unleash.

### Logs, error output, etc.

```bash
Response from Unleash:


{
  "version": 2,
  "features": [
    {
      "name": "<redacted>",
      "type": "release",
      "enabled": false,
      "project": "default",
      "stale": false,
      "strategies": [
        {
          "name": "flexibleRollout",
          "constraints": [],
          "parameters": {
            "groupId": "<redacted>",
            "rollout": "100",
            "stickiness": "default"
          }
        }
      ],
      "variants": [],
      "description": "<redacted>",
      "impressionData": false
    },
[...]


Response from Unleash-edge:
```json{
  "version": 2,
  "features": [
    {
      "name": "<redacted>",
      "type": "release",
      "description": "<redacted>",
      "enabled": false,
      "stale": false,
      "impressionData": false,
      "project": "default",
      "strategies": [
        {
          "name": "flexibleRollout",
          "constraints": [],
          "parameters": {
            "groupId": "<redacted>",
            "rollout": "100",
            "stickiness": "default"
          },
          "variants": null
        }
      ],
      "variants": []
    },
```
```


### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

Unleash 5.4.1, Unleash-edge 16.0.4

### Subscription type

Open source

### Hosting type

Self-hosted

### SDK information (language and version)

Ruby

gastonfournier:
[...]

PR #354, opened 2023-11-30
fix: bump deps to fix issue serializing

gastonfournier:
## About the changes
This bump unleash-types-rs to the version that fixes https://github.com/Unleash/unleash-edge/issues/353

Fixes: #353


server/src/client_api.rs
```diff
@@ -996,4 +996,72 @@ mod tests {
         let res = test::call_service(&local_app, request).await;
         assert_eq!(res.status(), StatusCode::NOT_FOUND);
     }
+
+    #[tokio::test]
+    pub async fn returns_emtpy_strategy_variants_when_upstream_strategy_variants_is_none() {
```
gastonfournier, line 1001: Not the most succinct test I've wrote, but gets the job done

PR #361, opened 2023-12-18
fix: updated release workflow to new recommended by cargo-dist

chriswk:
Cargo dist was updated, so this PR regenerates our release workflow.


gardleopard (approved):
LGTM

PR #362, opened 2023-12-19
fix: Do not remove features cache if upstream goes away

chriswk:
If upstream goes away, edge should remain durable and keep the features it has gotten for the tokens. Since we already keep the authentication status for the tokens we've seen this fix allows you to start edge, manage to talk to upstream once, then sever connection to upstream, but keep serving feature toggles for known tokens.

There's some superfluous version updates here, just to not have to merge 6 different PRs in addition to this one.

### DIFF
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -11,7 +11,7 @@ use actix_web::{
 };
 use dashmap::DashMap;
 use serde_qs::actix::QsQuery;
-use tracing::{debug, instrument, warn};
+use tracing::{debug, instrument};
 use unleash_types::client_features::Context;
 use unleash_types::client_metrics::{ClientApplication, ConnectVia};
 use unleash_types::{
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -11,7 +11,7 @@ use actix_web::{
 };
 use dashmap::DashMap;
 use serde_qs::actix::QsQuery;
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, warn};
 use unleash_types::client_features::Context;
 use unleash_types::client_metrics::{ClientApplication, ConnectVia};
 use unleash_types::{
@@ -214,7 +214,7 @@ security(
 )
 )]
 #[get("")]
-#[instrument(skip(engine_cache, token_cache))]
+#[instrument(skip(edge_token, req, engine_cache, token_cache))]
 async fn get_enabled_frontend(
     edge_token: EdgeToken,
     engine_cache: Data<DashMap<String, EngineState>>,
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -5,7 +5,7 @@ use actix_web::http::header::EntityTag;
 use chrono::Utc;
 use dashmap::DashMap;
 use reqwest::StatusCode;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 use unleash_types::client_features::Segment;
 use unleash_types::client_metrics::ClientApplication;
 use unleash_types::{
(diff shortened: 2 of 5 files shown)


server/src/http/feature_refresher.rs
```diff
@@ -898,6 +891,56 @@ mod tests {
...
+        feature_refresher.refresh_features().await;
+        assert!(!feature_refresher.tokens_to_refresh.is_empty());
+        assert!(!feature_refresher.features_cache.is_empty());
+        assert!(!feature_refresher.engine_cache.is_empty());
+        server.stop().await;
+        tokio::time::sleep(std::time::Duration::from_millis(5)).await; // To ensure our refresh is due
+        feature_refresher.refresh_features().await;
+        assert!(feature_refresher.tokens_to_refresh.is_empty());
+        assert!(!feature_refresher.features_cache.is_empty());
+        assert!(!feature_refresher.engine_cache.is_empty());
```
chriswk, line 941: The crux is here. Previously, when we receive 404 or connection failed we'd remove the features_cache and the engine_cache as well, thus removing all the durability built into Edge for when upstream goes away.

server/src/http/feature_refresher.rs
```diff
@@ -392,23 +392,15 @@ impl FeatureRefresher {
                                 }
                             }
                             FeatureError::NotFound => {
-                                warn!("Had a bad URL when trying to fetch features. Removing ourselves");
+                                info!("Had a bad URL when trying to fetch features. Removing ourselves from the list of refresh tasks");
```
gastonfournier, line 395: What happens when upstream is back online? I think 404 response code depends on each server/LB implementation, but maybe one reason for 404 is no pods available in a k8s cluster. 

If we consider the backoff implementation we did in the SDKs I think this is correct as it's aligned with that. I'm just wondering if this behavior changes when edge is still not ready (i.e. before receiving at least 1 successful response from upstream)

thomasheartman, line 395: What is "ourselves" pointing to here? It's a little unclear to me. Judging from the next line, it's the token? Maybe it's clear to the users, but could it be something like 

> Removing the refresh task for this token

or

> Removing this token from the list of refresh tasks

?

server/src/frontend_api.rs
```diff
@@ -214,7 +214,7 @@ security(
 )
 )]
 #[get("")]
-#[instrument(skip(engine_cache, token_cache))]
+#[instrument(skip(edge_token, req, engine_cache, token_cache))]
```
thomasheartman, line 217: Just so I'm clear: does this man we're not logging out the token and the request anymore? Any reason we changed it? (I'm sure it's fine, just curious)

server/src/http/feature_refresher.rs
thomasheartman, line 1: What's the reason for downgrading the log level from warn to info here?

server/src/http/feature_refresher.rs
thomasheartman: This sounds relevant, but it's a little early for me to understand exactly what you're saying 🥱 Would you be willing to explain this comment a little more for me?

thomasheartman (approved):
This looks good to me, but I'd listen to Gaston's comment here, it sounds relevant.

server/src/http/feature_refresher.rs
chriswk: So, every time the same key gets used we do the evaluation in the following order
1. Is key even allowed to access - this is stored in the TokenCache
2a. We have equal or wider access to the token's environment already, if so, we should be able to get features from the feature cache
2b We do not have equal or wider access, add the token to the list of tokens our featurerefresher cares about
3. Once we have features, give them to Yggdrasil to build an EngineState
4. Evaluate toggles if /api/frontend, return toggles if /api/client/features
---
What Gaston is worrying about is whether or not we repeat step 2, if upstream comes back online. From what I gather, we'll attempt to register it for refresh once every hour for normal tokens. I'm starting to think we should change the default here to 5 minutes or so.

For startup tokens, we try once every second until they achieve TokenStatus::Validated, then they too get moved into the refresh once per hour.

chriswk: yeah, agreed.

server/src/frontend_api.rs
chriswk: We don't want to log the token. And the request when looking at it didn't give us enough information.

thomasheartman: Got it; thanks!

server/src/http/feature_refresher.rs
chriswk: It's not really a warn, you can't really do anything about it.

server/src/http/feature_refresher.rs
thomasheartman: Right, and by step 2, you're referring to 

> 2. Once we have features, give them to Yggdrasil to build an EngineState

Right? Or 2a/2b? Anyway, it sounds like you've got it covered?

PR #372, opened 2024-01-05
task: Added version header to client register response

chriswk:
Related to https://github.com/Unleash/unleash/pull/5774 - This adds an X-Edge-Version header to Edge's responses to `/api/client/register`. 
This will allow Edge to compare the version header with the known version where we added `/api/client/metrics/bulk` to decide where to post metrics.


gastonfournier (approved):
LGTM!

PR #373, opened 2024-01-08
feat: added client authenticated bulk metrics

chriswk:
So, reflecting what we're doing here: https://github.com/Unleash/unleash/pull/5779 this PR adds a client token authenticated endpoint for posting bulk metrics. We also filter the metrics by the environment the token used to POST metrics has access to.

### DIFF
--- a/server/src/types.rs
+++ b/server/src/types.rs
@@ -1,6 +1,6 @@
 use std::cmp::min;
 use std::fmt;
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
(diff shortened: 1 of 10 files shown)


server/src/types.rs
```diff
@@ -82,6 +82,28 @@ pub struct EdgeToken {
     pub status: TokenValidationStatus,
 }
 
+impl Debug for EdgeToken {
```
sighphyre, line 85: Ayyyy!

sighphyre (approved):
Yep, LGTM

PR #390, opened 2024-01-15
Feat/use etag from upstream

chriswk:
This changes our client features endpoint to rely on etag we get from upstream, avoiding doing the service call and just returning 304 if we match.

### DIFF
--- a/server/src/builder.rs
+++ b/server/src/builder.rs
@@ -15,6 +15,7 @@ use crate::{
     auth::token_validator::TokenValidator,
     cli::{CliArgs, EdgeArgs, EdgeMode, OfflineArgs},
     error::EdgeError,
+    filters, hashing,
     http::{feature_refresher::FeatureRefresher, unleash_client::UnleashClient},
     types::{EdgeResult, EdgeToken, TokenType},
 };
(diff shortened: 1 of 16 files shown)


server/src/builder.rs
```diff
@@ -205,3 +233,101 @@ pub async fn build_caches_and_refreshers(args: CliArgs) -> EdgeResult<EdgeInfo>
...
 }
+
+#[cfg(test)]
+mod tests {
+    use unleash_types::client_features::{ClientFeature, Strategy};
+
+    use super::*;
+
+    #[test]
+    fn building_an_initial_state_from_backup_also_adds_etags_for_features() {
```
sighphyre, line 244: Nice test case!

server/src/middleware/etag.rs
```diff
@@ -0,0 +1,154 @@
...
+        .map(|h| h.to_str().unwrap())
+        .and_then(|header_str| EdgeToken::from_str(header_str).ok())
+}
+
+fn we_know_this_etag_from_upstream(
+    etag_cache: Arc<DashMap<EdgeToken, EntityTag>>,
+    client_token: &Option<EdgeToken>,
+    if_none_match: &Option<IfNoneMatch>,
+) -> bool {
+    match (if_none_match, client_token) {
```
sighphyre, line 148: The match one looks odd to me here. You can do a tuple destructuring `if let Some` here:

fn we_know_this_etag_from_upstream(
    etag_cache: Arc<DashMap<EdgeToken, EntityTag>>,
    client_token: &Option<EdgeToken>,
    if_none_match: &Option<IfNoneMatch>,
) -> bool {
    if let (Some(if_none), Some(token)) = (if_none_match, client_token) {
        etag_cache.get(token).map_or(false, |etag| {
            if_none == &IfNoneMatch::Any || if_none.to_string() == etag.to_string()
        })
    } else {
        false
    }
}

Entirely optional

sighphyre (approved):
Nice! Small comment, up to you if you want to do anything with it

PR #399, opened 2024-01-19
feat: add endpoints for metrics that support using the proxy/all

sighphyre:
Adds two new endpoints to the metrics handling code to alias the metrics endpoints to metrics/all. This allows the existing proxy SDKs to use the `features/all` endpoints and still get metrics without patches or rerouting.


chriswk (approved):
You might want to add the endpoints to the openapi spec as well, but the code looks sane.

sighphyre:
> You might want to add the endpoints to the openapi spec as well, but the code looks sane.

Oh I probably do, I see I missed one, thank you!

PR #403, opened 2024-01-23
fix: include strategy variant stickiness

sighphyre:
Changes the responses in Edge to include the stickiness property in Strategy Variants.

This technically shouldn't be necessary but this allows us to support versions on our Node SDK that incorrectly use this property for stickiness calculations. 

Note that this PR includes necessary patches to Yggdrasil to handle cases where updating features could fail and previously resulted in crashes in Edge. It's probably useful to review the commit that makes the changes for this patch in isolation: https://github.com/Unleash/unleash-edge/pull/403/commits/ecd3070262b1160a363389686dfce27f62122cf7

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -280,7 +280,9 @@ mod tests {
     use maplit::hashmap;
     use reqwest::StatusCode;
     use ulid::Ulid;
-    use unleash_types::client_features::{ClientFeature, Constraint, Operator, Strategy};
+    use unleash_types::client_features::{
+        ClientFeature, Constraint, Operator, Strategy, StrategyVariant,
+    };
     use unleash_types::client_metrics::{
         ClientMetricsEnv, ConnectViaBuilder, MetricBucket, ToggleStats,
     };
@@ -431,7 +433,12 @@ mod tests {
                     project: Some("default".into()),
                     strategies: Some(vec![
                         Strategy {
-                            variants: None,
+                            variants: Some(vec![StrategyVariant {
+                                name: "test".into(),
+                                payload: None,
+                                weight: 7,
+                                stickiness: Some("sticky-on-something".into()),
+                            }]),
                             name: "standard".into(),
                             sort_order: Some(500),
                             segments: None,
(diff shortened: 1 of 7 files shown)


server/src/client_api.rs
```diff
@@ -507,6 +514,52 @@ mod tests {
...
+        .unwrap();
+        production_token.token_type = Some(TokenType::Client);
+        production_token.status = TokenValidationStatus::Validated;
+        token_cache.insert(production_token.token.clone(), production_token.clone());
+        let req = make_features_request_with_token(production_token.clone()).await;
+        let res: ClientFeatures = test::call_and_read_body_json(&app, req).await;
+
+        assert_eq!(res.features.len(), cached_client_features().features.len());
+        let strategy_variant_stickiness = res
+            .features
```
sighphyre, line 542: @chriswk I hate this chain so much but I can't spot an obvious way accurately assert the stickiness is Some without rooting around in the data structure. I'm very open to suggestions!

server/src/middleware/client_token_from_frontend_token.rs
```diff
@@ -195,7 +195,7 @@ mod tests {
         upstream_features_cache.insert(client_token.environment.clone().unwrap(), features.clone());
         let upstream_engine_cache: Arc<DashMap<String, EngineState>> = Arc::new(DashMap::default());
         let mut engine = EngineState::default();
-        engine.take_state(features.clone());
+        engine.take_state(features.clone()).unwrap();
```
sighphyre, line 198: Unwrap is probably okay, this is a test

server/src/client_api.rs
chriswk: I'd say bring some lens library. ;) in all seriousness though, this is terrible, but consequence of having a deep data structure I guess.

server/src/client_api.rs
```diff
@@ -507,6 +514,52 @@ mod tests {
...
+            App::new()
+                .app_data(Data::from(features_cache.clone()))
+                .app_data(Data::from(token_cache.clone()))
+                .service(web::scope("/api/client").service(get_features)),
+        )
+        .await;
+
+        features_cache.insert("production".into(), cached_client_features());
+        let mut production_token = EdgeToken::try_from(
+            "*:production.03fa5f506428fe80ed5640c351c7232e38940814d2923b08f5c05fa7".to_string(),
```
chriswk, line 531: I think I added some test helper methods here for creating client tokens already validated, also, you don't need to use an actual secret. Anything behind the . could probably just be 'somesecret'.

chriswk (approved):
Some suggestions left to simplify (you don't need to do anything about them) but this LGTM

server/src/client_api.rs
sighphyre: Oof I really don't want to take on a new dev dependency for a single test when I don't see us getting much out of it for other things. I can live with this (with a small amount of crying maybe)

PR #406, opened 2024-01-25
chore: make docker example use up to date edge and unleash versions

sighphyre:
A few small clean ups on our docs an examples:

- Removes all the constant versions in our markdown and examples and replaced with version templates or `latest`
- Makes the minimum version of Unleash required clear
- Marks Unleash as `pull_always` in the example compose


chriswk (approved):
LGTM

PR #414, opened 2024-02-01
feat: add metrics for failed requests

### LINKED ISSUE #412: Add metrics for failed requests

I want to know what error status codes (4xx|5xx) were returned by edge, so that I can fine tune monitoring and runbooks.
Currently those are explicitly ommited (https://github.com/Unleash/unleash-edge/blob/d93c28d1a431c24a640ed7013da5fb44d841592f/server/src/metrics/actix_web_metrics.rs#L370), but I believe would still provide much value in speeding up debugging process.

### Background

_No response_

### Solution suggestions

Move metrics update before `if let Ok(res) = res {}` block

nunogois:
Adds metrics for failed requests.

Closes #412

### DIFF
--- a/server/src/metrics/actix_web_metrics.rs
+++ b/server/src/metrics/actix_web_metrics.rs
@@ -363,32 +363,31 @@ where
 
         let request_metrics = self.metrics.clone();
         Box::pin(self.service.call(req).map(move |res| {
-            request_metrics
-                .http_server_active_requests
-                .add(-1, &attributes);
-
-            // Ignore actix errors for metrics
-            if let Ok(res) = res {
-                attributes.push(HTTP_RESPONSE_STATUS_CODE.i64(res.status().as_u16() as i64));
-                let response_size = res
-                    .response()
-                    .headers()
-                    .get(CONTENT_LENGTH)
-                    .and_then(|len| len.to_str().ok().and_then(|s| s.parse().ok()))
-                    .unwrap_or(0);
-                request_metrics
-                    .http_server_response_size
-                    .record(response_size, &attributes);
-
-                request_metrics.http_server_duration.record(
-                    timer.elapsed().map(|t| t.as_secs_f64()).unwrap_or_default(),
-                    &attributes,
-                );
-                Ok(res)
-            } else {
-                res
-            }
-        }))
+          let status_code = match &res {
+              Ok(res) => res.status().as_u16() as i64,
+              Err(e) => {
+                e.as_response_error().status_code().as_u16() as i64
+              },
+          };
+      
+          attributes.push(HTTP_RESPONSE_STATUS_CODE.i64(status_code));
+          request_metrics.http_server_active_requests.add(-1, &attributes);
+      
+          let response_size = match &res {
+              Ok(res) => res.response().headers().get(CONTENT_LENGTH)
+                  .and_then(|len| len.to_str().ok().and_then(|s| s.parse().ok()))
+                  .unwrap_or(0),
+              Err(_) => 0,
+          };
+          request_metrics.http_server_response_size.record(response_size, &attributes);
+      
+          request_metrics.http_server_duration.record(
+              timer.elapsed().map(|t| t.as_secs_f64()).unwrap_or_default(),
+              &attributes,
+          );
+      
+          res
+      }))
     }
 }
 
--- a/server/src/metrics/actix_web_metrics.rs
+++ b/server/src/metrics/actix_web_metrics.rs
@@ -364,11 +364,10 @@ where
         let request_metrics = self.metrics.clone();
         Box::pin(self.service.call(req).map(move |res| {
           let status_code = match &res {
-              Ok(res) => res.status().as_u16() as i64,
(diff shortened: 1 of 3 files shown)


server/src/metrics/actix_web_metrics.rs
```diff
@@ -363,32 +363,31 @@ where
...
+              Ok(res) => res.status().as_u16() as i64,
+              Err(e) => {
+                e.as_response_error().status_code().as_u16() as i64
+              },
+          };
+      
+          attributes.push(HTTP_RESPONSE_STATUS_CODE.i64(status_code));
+          request_metrics.http_server_active_requests.add(-1, &attributes);
+      
+          let response_size = match &res {
```
sighphyre, line 376: Personal taste but I think match statements where the error handling is just return a Copy type are better expressed as map/unwrap_or.

You can do this:


            let response_size = res
                .as_ref()
                .map(|res| {
                    res.response()
                        .headers()
                        .get(CONTENT_LENGTH)
                        .and_then(|len| len.to_str().ok().and_then(|s| s.parse().ok()))
                        .unwrap_or(0u64)
                })
                .unwrap_or(0);

sighphyre (approved):
LGTM, left some suggestions. The I think you should consider doing the match/map chain but the logic looks good to me and the tests seem to do what they need!

server/src/metrics/actix_web_metrics.rs
```diff
@@ -428,3 +427,72 @@ impl dev::Handler<actix_web::HttpRequest> for PrometheusMetricsHandler {
...
+  }
+
+    #[tokio::test]
+    async fn test_middleware_response_metrics() {
+      let registry = Registry::new();
+      let (_, request_metrics) = prom_metrics::test_instantiate_without_tracing_and_logging(Some(registry.clone()));
+
+      let mut app = test::init_service(
+        App::new()
+          .wrap(request_metrics.clone())  // Initialize your middleware
```
sighphyre, line 465:
```suggestion
          .wrap(request_metrics.clone())
```

server/src/metrics/actix_web_metrics.rs
```diff
@@ -428,3 +427,72 @@ impl dev::Handler<actix_web::HttpRequest> for PrometheusMetricsHandler {
...
+      assert_eq!(resp_server_error.status(), StatusCode::INTERNAL_SERVER_ERROR);
+
+      let mut buffer = Vec::new();
+      let encoder = TextEncoder::new();
+      let metric_families = registry.gather();
+      encoder.encode(&metric_families, &mut buffer).unwrap();
+      let metrics_output = String::from_utf8(buffer).unwrap();
+      
+      let value_ok = parse_metrics_for_status_code(&metrics_output, 200).expect("Metric with status code 200 not found");
+      assert_eq!(value_ok, -1.0, "Metric value for status code 200 did not match expected");
```
chriswk, line 490: why -1.0 and not 1?

server/src/prom_metrics.rs
```diff
@@ -134,3 +134,10 @@ fn register_custom_metrics(registry: &prometheus::Registry) {
         ))
         .unwrap();
 }
+
+#[cfg(test)]
+pub fn test_instantiate_without_tracing_and_logging(registry: Option<prometheus::Registry>) -> (PrometheusMetricsHandler, RequestMetrics) {
+  let registry = registry.unwrap_or_else(instantiate_registry);
+  register_custom_metrics(&registry);
+  instantiate_prometheus_metrics_handler(registry)
+}
```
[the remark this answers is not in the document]
chriswk: :+1: - great catch.

server/src/metrics/actix_web_metrics.rs
chriswk: and same thing for the other values. I see we reduce by -1.0 once it's done, I'm expecting the active_requests to fluctuate between 0 and however many simultaneous connections we can handle. so it being `-1.0` would mean it only went through the `it's done` branch instead of both the `started` and `it's done` branch.

server/src/metrics/actix_web_metrics.rs
nunogois: Addressed in https://github.com/Unleash/unleash-edge/pull/414/commits/ff0a6184b1dee4dd46be7eb577a66833025c433c

server/src/metrics/actix_web_metrics.rs
```diff
@@ -363,32 +363,31 @@ where
...
-                request_metrics.http_server_duration.record(
-                    timer.elapsed().map(|t| t.as_secs_f64()).unwrap_or_default(),
-                    &attributes,
-                );
-                Ok(res)
-            } else {
-                res
-            }
-        }))
+          let status_code = match &res {
```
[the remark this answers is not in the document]
nunogois: Nice catch! Addressed in https://github.com/Unleash/unleash-edge/pull/414/commits/175e782f2bb45fb3c9d64e21da38ce0d3c7ad724

chriswk (approved):
LGTM

PR #419, opened 2024-02-09
experiment: implementation of Edge running in AWS Lambda

sighphyre:
DO NOT MERGE THIS. THIS WILL BREAK EDGE IN ITS CURRENT FORM AND IS A WORK IN PROGRESS

# What

This is a spike to get Edge running in AWS lambda

# Mucking about locally

Should be able to do

`cargo lambda watch` to run the lambda locally (you'll need https://www.cargo-lambda.info/guide/getting-started.html)

An HTTP hit to localhost:9000, will trigger the lambda and start the Edge server

# Warnings

- This duplicates Edge's main method with some fiddling - this needs a common abstraction
- This currently removes shadow-rs, which doesn't want to compile on the AWS lambda runtime. I'm not all that surprised that doesn't compile since compilation occurs during execution
- This doesn't make a proper bridge between the lambda HTTP infrastructure and Edge's HTTP setup. This will spawn a lambda that will respond to HTTP calls but do nothing, the Edge server is mounted on a separate thread and is spawned once during execution until the process is destroyed 
- All the setup is hardcoded so I don't have to deal with Clap args, this needs to be delegated to the lambda setup
- This introduces a new binary, which makes `cargo run` ambiguous (very solvable)


sighphyre:
Fun but dead experiment, closing this, I don't think we're pursuing this in this way anymore

PR #420, opened 2024-02-14
feat: Add /internal-backstage/metricsbatch

chriswk:
This allows server admins to debug what is currently in our cached metrics ready for the next batch. Behind /internal-backstage to signify that you shouldn't be exposing this part.


sighphyre (approved):
lg

PR #422, opened 2024-02-22
feat: internal backstage features endpoint

chriswk:
Adds an endpoint /internal-backstage/features displaying which features we store per cache key


sighphyre (approved):
:heart: 

Thank you!

PR #425, opened 2024-02-26
docs: remove service account token documentation

gardleopard:
Removing this documentation because we want to phase out the usage of this functionality as fast as possible, there will be PRs addressing this code, but this can be merged before that is ready to make it harder to discover functionality we do not recommend to use.


chriswk (commented):
Looks like you might be removing a bit too much here. Shouldn't it just be removing the --service-account flag? Not the entire output for the help command?

gardleopard:
> Looks like you might be removing a bit too much here. Shouldn't it just be removing the --service-account flag? Not the entire output for the help command?

I removed all that on purpose, I did not see why the help command output was there. But I now understand why it was there. Ill get it back.

chriswk (approved):
Nice. LGTM

PR #426, opened 2024-02-27
feat: remove service account support

gardleopard:
This will remove the service account token functionality. This functionality makes it easy to use frontend-tokens with unleash-edge.

To keep frontend-tokens working in Unleash Edge, replace service account tokens with client tokens when configuring Unleash Edge. Use wildcard client tokens for each environment you want to serve with Edge and your frontend tokens should still work.

example:
```
./unleash-edge edge -u https://your.unleash.instance/url -t "*:development.7bde4d213a946b20af49960d1d457aa437a592d6728250c1a48c5ff4" -t "*:production.08256d30ee6fc1a16677ab812c1c72f4bf5b8f07cefd870dec705479"
```

The rationale for removing this functionality is that it is quite complex logic to maintain and using environment specific wildcard client tokens gives similar functionality and is more explicit on what environments that can be used.


ivarconr:
We should at least have an error log if the user tries to send in a service token.

gardleopard:
> We should at least have an error log if the user tries to send in a service token.

works after i removed the option as well:
```
error: unexpected argument '--service-account-token' found

  tip: a similar argument exists: '--metrics-interval-seconds'

Usage: unleash-edge edge --upstream-url <UPSTREAM_URL> --tokens <TOKENS> --metrics-interval-seconds <METRICS_INTERVAL_SECONDS>

For more information, try '--help'.
```

chriswk (approved):
Nice. LGTM :+1:

PR #435, opened 2024-02-29
fix: degrade broken toggles gracefully in frontend api

sighphyre:
This upgrades Yggdrasil to a version that's tolerant to all kinds of toggle compile failures. It also makes the frontend/proxy apis return errors rather than panic when resolving toggles that should exist but don't

### DIFF
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -314,7 +314,7 @@ fn get_enabled_features(
     })?;
     let feature_results = engine.resolve_all(&context_with_ip, &None).ok_or_else(|| {
         EdgeError::FrontendExpectedToBeHydrated(
-            "Feature cache has not been hydrated yet, but it was expected to be".into(),
+            "Feature cache has not been hydrated yet, but it was expected to be. This can be due to a race condition from calling edge before it's ready. This error might auto resolve as soon as edge is able to fetch from upstream".into(),
         )
     })?;
     Ok(Json(frontend_from_yggdrasil(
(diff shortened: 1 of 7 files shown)


server/src/frontend_api.rs
```diff
@@ -220,7 +220,11 @@ fn post_all_features(
     let engine = engine_cache.get(&key).ok_or_else(|| {
         EdgeError::FrontendNotYetHydrated(FrontendHydrationMissing::from(&edge_token))
     })?;
-    let feature_results = engine.resolve_all(&context_with_ip, &None).unwrap();
+    let feature_results = engine.resolve_all(&context_with_ip, &None).ok_or_else(|| {
+        EdgeError::FrontendExpectedToBeHydrated(
+            "Feature cache has not been hydrated yet, but it was expected to be".into(),
```
gastonfournier, line 225: Ok, maybe too verbose, but as a user this error would tell me nothing
```suggestion
            "Feature cache has not been hydrated yet, but it was expected to be. This can be due to a race condition from calling edge before it's ready. This error might auto resolve as soon as edge is able to fetch from upstream".into(),
```

gastonfournier (approved):
Looks fine, I'd just try to have an error message that'd help users understand why this error is happening

server/src/frontend_api.rs
sighphyre: This isn't that error though, that's handled by Edge hydration in the line above. This is deep into "things that should never happen". The only path to this error happening is if Yggdrasil has been inserted into the compiled cache without having been compiled and that should now not happen.

The error log that points out the toggles are broken is logged in the state hydration handling code

But this looks silly. Lemme patch Yggdrasil so this special case goes away

sighphyre: Nice, I took your suggestion. It makes sense now. The only way this can happen is if Ygg has not been hydrated now

chriswk (approved):
LGTM

Issue #436, opened 2024-02-29
Edge metrics POST interval grows unexpectedly. 


chriswk:
### Describe the bug

After running edge for a while I spotted that it was suddenly sleeping for 62 seconds, and then 64 seconds. Double checking I see that we add jitter to previous interval instead of resetting back to initial configured duration before adding jitter.

### Steps to reproduce the bug

1. Start edge with `RUST_LOG=warn,unleash_edge=trace`
2. Observe 
```
2024-02-29T12:09:33.846888Z TRACE unleash_edge::http::background_send_metrics: Done posting traces. Sleeping for 62 seconds and then going again
2024-02-29T12:09:33.846888Z TRACE unleash_edge::http::background_send_metrics: Done posting traces. Sleeping for 70 seconds and then going again
```
3. ?
4. Profit? 

### Expected behavior

Interval should never exceed initial configured interval + max_jitter unless we're erroring. Since all posts succeeded, we probably shouldn't add jitter at all. the normal interval due to random startup times should be enough

### Logs, error output, etc.

```bash
2024-02-29T12:09:33.846888Z TRACE unleash_edge::http::background_send_metrics: Done posting traces. Sleeping for 70 seconds and then going again
```

### Unleash version

Edge 17.1.0

### Subscription type

None

### Hosting type

None

PR #437, opened 2024-02-29
fix: make metrics interval not climb forever

chriswk:
Previously we used the previous iteration of the loop's interval to decide next interval. This PR changes to always use the initial send_interval as what to base the interval calculation of

fixes: #436


gardleopard (approved):
LGTM

sighphyre (approved):
Yes!

PR #446, opened 2024-03-26
fix: Removed /edge/metrics and logic for posting to /edge/metrics

chriswk:
We no longer want /edge/metrics to be available, since we have released 17.0.0, 17.1.0, 18.0.0 and 18.0.1 with compatibility for both `/edge/metrics` and `/api/client/metrics/bulk` this PR now strips the old endpoint in preparation for releasing 19.0.0. After this is released, you will need at least Unleash 5.9.0 to get metrics from Edge.


sighphyre (approved):
Fantastic!

PR #449, opened 2024-04-25
fix: bump unleash types so that impression data is correctly camelcased

sighphyre:
This fixes an issue where the `impression_data` field would not be correctly camelCased on front-end responses

Done via a bump in unleash types. Yggdrasil gets a bump too since that has a dependency on types and needs this to compile correctly


chriswk (approved):
:hammer: :timer_clock:

PR #453, opened 2024-05-07
Feature/add redis cluster support

chriswk:
Adding a `--redis-mode single | cluster` CLI option. To be used if your redis is configured as a cluster, to ensure our client listens to REDIRECT to slot commands when doing writes.


sighphyre (approved):
LGTM!

PR #457, opened 2024-05-07
fix/remove feature refresher

sighphyre:
Removes the feature refresher flow from persistence. Having this persisted means that we can get into inconsistent states with the e tag. This will now force Edge to always check upstream for an update on startup.


chriswk (approved):
Nice. and feels good

Issue #463, opened 2024-05-15
UNLEASH EDGE DAISY CHAIN ISSUE


pi194046:
### Describe the bug

i created a unleash edge running in offline mode at port 3064
I created a unleash edge running in edge mode at port 3063 pointing to uleash edge offline instance port 3064

when I testing using the token and curl request , I am able to successfully make the API call on offline instance at port 3064 but using the same token on port 3063 is giving 403 Forbidden ie token is invalid

If I use the same approach and have unleash server running at port 4242, and I create a unleash edge running in edge mode at pointing to unleash server then the curl request is successful at both 4242 and port 3063

I had used the same token created on unleash server while configuring the unleash offline instance 

### Steps to reproduce the bug

Run the below command in one terminal
docker run --rm -e RUST_LOG="trace,unleash_edge=trace" -v `pwd`/config:/tmp/config -p 3064:3063 unleashorg/unleash-edge offline -b /tmp/config/a.file -r 2 -t 'default:development.40df9a2423b500ad80b53f3f97153af8f6768df483e09aa712345678'

please the a.file in the location in the config folder in the current directory

Run the below command in another terminal
docker run -p 3063:3063 -e RUST_LOG="trace,unleash_edge=trace" --rm unleashorg/unleash-edge edge -u http://your-domain.com:3064 -t 'default:development.40df9a2423b500ad80b53f3f97153af8f6768df483e09aa712345678' -f 10 --token-revalidation-interval-seconds 10

the below call is successful
curl -X GET http://your-domain.com:3064/api/client/features -H 'Accept: application/json' -H 'Authorization: default:development.40df9a2423b500ad80b53f3f97153af8f6768df483e09aa712345678' -v

the below call is a failure
curl -X GET http://your-domain.com:3063/api/client/features -H 'Accept: application/json' -H 'Authorization: default:development.40df9a2423b500ad80b53f3f97153af8f6768df483e09aa712345678' -v

### Expected behavior

daily chaining unleash edge instances with edge instances pointing to an offline instance should work properly

### Subscription type

None

### Hosting type

Self-hosted

chriswk:
I moved this issue to Unleash-edge rather than the Unleash repo.

[...]

PR #465, opened 2024-05-15
feat: added support for client and frontend tokens to offline mode.

chriswk:
Also assumes tokens passed in `--tokens` are client tokens, so if instantiated in offline mode with the current way of `--tokens <token1>`, another node connected to it, will get a validation result saying this token is a client token. If one wants to add frontend tokens, use `--frontend-tokens`.

### Discussion points
Not a 100% sure we really want to add this, but it feels like a better solution than what we currently have where offline nodes just refuses any edge clients. (It still works with an offline node from a normal SDK, but Edge is not a normal SDK)

fixes: #463


server/src/cli.rs
```diff
@@ -221,6 +221,12 @@ pub struct OfflineArgs {
     /// Tokens that should be allowed to connect to Edge. Supports a comma separated list or multiple instances of the `--tokens` argument
     #[clap(short, long, env, value_delimiter = ',')]
     pub tokens: Vec<String>,
```
sighphyre, line 223: More a question than a suggestion but can we mark this as deprecated now?

sighphyre (approved):
LGTM

server/src/cli.rs
chriswk: Sure

PR #466, opened 2024-05-22
fix: Only replace projects covered by token

chriswk:
Previously, if the update was empty, we assumed we needed to replace all stored features. This is not the case. We only need to remove the projects that are connected to the token we're using for the merge.

In addition, debug output on `/internal-backstage/tokens` now includes how many features was received in the update

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -1,27 +1,27 @@
-use std::{sync::Arc, time::Duration};
 use std::collections::HashSet;
+use std::{sync::Arc, time::Duration};
 
 use actix_web::http::header::EntityTag;
 use chrono::Utc;
 use dashmap::DashMap;
 use reqwest::StatusCode;
 use tracing::{debug, info, warn};
+use unleash_types::client_features::Segment;
+use unleash_types::client_metrics::ClientApplication;
 use unleash_types::{
     client_features::{ClientFeature, ClientFeatures},
     Deduplicate,
 };
-use unleash_types::client_features::Segment;
-use unleash_types::client_metrics::ClientApplication;
 use unleash_yggdrasil::EngineState;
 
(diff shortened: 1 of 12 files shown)


server/src/http/feature_refresher.rs
```diff
@@ -70,29 +70,25 @@ fn merge_segments_update(
...
-        }
+        let mut to_keep: Vec<ClientFeature> = original
+            .iter()
+            .filter(|toggle| {
+                let p = toggle.project.clone().unwrap_or_else(|| "default".into());
+                !projects_to_update.contains(&p)
+            })
+            .cloned()
+            .collect();
+        to_keep.extend(updated.iter().cloned());
```
nunogois, line 90: I'm assuming that, in case `updated` is empty, this will act as:
`to_keep.extend([])` effectively returning the existing `to_keep` without any changes.

nunogois (approved):
LGTM

server/src/http/feature_refresher.rs
chriswk: yup, to_keep.extend will keep what's in to_keep, but add from updated, and since we've already removed all toggles from to_keep that matches one of the projects this token has access to, we won't have to worry about duplicates.

server/src/http/feature_refresher.rs
```diff
@@ -1199,6 +1202,7 @@ mod tests {
             last_refreshed: None,
             last_check: None,
             failure_count: 0,
+            last_feature_count: 0,
```
sighphyre, line 1205: An Option surely? (Yes I'm calling you Sirely).

I'd rather know if there's no update but I think the adjacent critical fix is more important so I won't fuss

chriswk: Agreed, fixing.

PR #467, opened 2024-05-29
chore: added dependency review workflow

chriswk:
Same as https://github.com/Unleash/unleash/pull/7206. Due to new OSS requirements from customers, we need to make sure we're not adding extra work for distributing by using licenses requiring redistributions of source code.


gastonfournier (approved):
LG

PR #474, opened 2024-07-03
feat: strict behavior

nunogois:
https://linear.app/unleash/issue/2-2370/add-closed-mode

Adds the concept of "strict" and "dynamic" behaviors when running in Edge mode.

 - **Strict**: The new recommended behavior. New tokens are not accepted unless they are subsumed by registered tokens.
 - **Dynamic**: The default behavior. New tokens are automatically registered and features will be fetched for them.

This means no change in behavior for existing users. However we recommend edge mode is started with `--strict` from now on. The plan is to drop this sometime in the future and instead always default to strict mode.

Some tests assumed the "dynamic" behavior by default, so they were adapted accordingly.


server/src/http/feature_refresher.rs
```diff
@@ -111,6 +112,7 @@ impl Default for FeatureRefresher {
             features_cache: Default::default(),
             engine_cache: Default::default(),
             persistence: None,
+            open: Default::default(),
```
sighphyre, line 115: Personally a huge fan of explicit stuff, especially for primitives

```suggestion
            open: false,
```

server/src/http/feature_refresher.rs
```diff
@@ -154,6 +158,7 @@ impl FeatureRefresher {
             engine_cache: Arc::new(Default::default()),
             refresh_interval: chrono::Duration::seconds(10),
             persistence: None,
+            open: Default::default(),
```
sighphyre, line 161: Me again with my love of explicit stuff

```suggestion
            open: false,
```

server/src/http/feature_refresher.rs
```diff
@@ -515,6 +526,7 @@ mod tests {
             engines_cache,
             duration,
             None,
+            false,
```
sighphyre, line 529: Won't complain if you don't feel like doing this but this struct has a `Default`, this big chunk of noise where you have to update this in tests a lot of times wouldn't happened if these tests used default + destructuring. If you want to be a good citizen that would prevent this problem coming back (but probably not make this PR any simpler)

server/src/error.rs
```diff
@@ -86,7 +86,7 @@ impl Display for CertificateError {
 
 #[derive(Debug)]
 pub enum EdgeError {
-    AuthorizationDenied,
+    AuthorizationDenied(String),
```
sighphyre, line 89: I'd recommend making a new error case for this rather than modifying an existing one. You should be able to get exactly the same behavior with fewer changes and it'll have the nice side effect of being more explicit

server/src/http/feature_refresher.rs
```diff
@@ -1245,6 +1267,7 @@ mod tests {
             engine_cache: Arc::new(Default::default()),
             refresh_interval: chrono::Duration::seconds(0),
             persistence: None,
+            open: Default::default(),
```
sighphyre, line 1270: For tests I'm an even bigger fan of explicit

```suggestion
            open: false,
```

sighphyre (commented):
Ye, it's sexy. Nothing here that I feel the need to block this for but I do feel strongly about adding the new error type over modifying the existing one. It's a personal feeling so I won't cry if no one agrees with me

sighphyre (approved):
Meant to approve 

:banana:

chriswk:
We should make sure that we fail hard and exit if you start edge in closed mode with no tokens.

server/src/cli.rs
```diff
@@ -174,6 +174,10 @@ pub struct EdgeArgs {
     /// Token header to use for both edge authorization and communication with the upstream server.
     #[clap(long, env, global = true, default_value = "Authorization")]
     pub token_header: TokenHeader,
+
+    /// If set to true, Edge starts in open mode. Open mode means that Edge will accept tokens outside of the scope of the startup tokens
+    #[clap(short, long, env, default_value_t = true)]
```
chriswk, line 179: Shouldn't we default to false here?

nunogois: Good point, addressed in https://github.com/Unleash/unleash-edge/pull/474/commits/e31c1ce63377e5ca945c63f88b70dbcdf156ddde

nunogois:
> We should make sure that we fail hard and exit if you start edge in closed mode with no tokens.

Agreed, that's something that came to mind after our session. What do you think of this approach? https://github.com/Unleash/unleash-edge/pull/474/commits/8459d1ab9133ce86e32e46f4f5fd005ea44c44ca

Also touches offline mode for consistency (and scouting, seems like offline mode was happy starting with no tokens?), but I'm happy to leave that out of this PR if you prefer.

gardleopard (changes requested):
Closed vs Open mode. I think this is the major thing in this project for UX. I think that we should be careful on what we name the modes. Its not clear to me what closed and open means here. I think consistent mode vs eventual cosistent or deterministic vs non deterministic might be some options to consider for this. Id like us to involve @FredrikOseberg before selecting the name of this mode.

server/src/error.rs
nunogois: You mean something like this? https://github.com/Unleash/unleash-edge/pull/474/commits/e39da3970e92cb67d8350fae6a9f8eb30b9550f6

FredrikOseberg:
Looks like I'm a bit late to the party and without that much context. But it _looks_ to me like we already have Edge mode and Offline mode as top level modes. These modes seem to be changing the default behavior of Edge mode if I'm reading this right? So we have two sub-modes of Edge mode? Is that the correct interpretation here?

server/src/http/feature_refresher.rs
nunogois: Hmm do you mean this? https://github.com/Unleash/unleash-edge/pull/474/commits/3407d2d444d685c04fcd66eacce610e1e3a750da

nunogois:
> Looks like I'm a bit late to the party and without that much context. But it _looks_ to me like we already have Edge mode and Offline mode as top level modes. These modes seem to be changing the default behavior of Edge mode if I'm reading this right? So we have two sub-modes of Edge mode? Is that the correct interpretation here?

I think that's spot on. 

Edge and Offline modes as top level modes, just like before.

Edge behaved like the new "open" mode. We now want it to behave like the new "closed" mode by default, but leave "open" mode as an option.

The difference between the modes is explained in the PR description. Essentially "open" mode allows you to hit Edge with any token. Once validated, that token is added to the list of tokens to refresh (fetch features for). "Closed" mode will return 403 for tokens that, once validated, are not subsumed by the currently known tokens.

The idea is that e.g. you start Edge with a wildcard token for an environment, and any token that matches that environment should be OK, even if it's narrower in scope. However tokens for other environments are denied.

FredrikOseberg:
> > Looks like I'm a bit late to the party and without that much context. But it _looks_ to me like we already have Edge mode and Offline mode as top level modes. These modes seem to be changing the default behavior of Edge mode if I'm reading this right? So we have two sub-modes of Edge mode? Is that the correct interpretation here?
> 
> I think that's spot on.
> 
> Edge and Offline modes as top level modes, just like before.
> 
> Edge behaved like the new "open" mode. We now want it to behave like the new "closed" mode by default, but leave "open" mode as an option.
> 
> The difference between the modes is explained in the PR description. Essentially "open" mode allows you to hit Edge with any token. Once validated, that token is added to the list of tokens to refresh (fetch features for). "Closed" mode will return 403 for tokens that, once validated, are not subsumed by the currently known tokens.
> 
> The idea is that e.g. you start Edge with a wildcard token for an environment, and any token that matches that environment should be OK, even if it's narrower in scope (e.g. same environment, but specific project). However tokens for other environments are denied, instead of dynamically added to the list.

Is it confusing to refer to it as a mode then? I would expect mode to be a state that I can start Edge in based on the current implementation of mode. For example: 

`Edge mode | Offline mode | Open mode | Closed mode`

While in reality it seems like these are behaviors of edge mode. Would it make more sense to refer to it as Edge mote behavior: `strict` or `dynamic` behavior? I feel like we might be mixing concepts we already have here.

server/src/error.rs
sighphyre: Ye! That's nice!

server/src/http/feature_refresher.rs
sighphyre: Yaaaas

sighphyre:
Okay I thought about this a lot more and I spoke to @gardleopard. What's our plan for `open/closed` (or whatever we want to call it). I'd really like to make the open mode go away medium term. I think it's a trap and my personal feelings around the flag here is to give people a window to migrate away from open mode

That aside, I think @FredrikOseberg's suggestion of `dynamic/strict` is a goodie.

sighphyre:
> > Edge mode | Offline mode | Open mode | Closed mode
> 
> While in reality it seems like these are behaviors of edge mode. Would it make more sense to refer to it as Edge mote behavior: strict or dynamic behavior? I feel like we might be mixing concepts we already have here.

Okay. Also this. What makes sense to me is `Edge` | `Offline` | `Closed` (the names need some love maybe). But really Open/Closed can't affect offline mode

nunogois:
Following the recent comments and discussions I made some [changes](https://github.com/Unleash/unleash-edge/pull/474/commits/de8eb868adff7b977901eb87de8419b09122fcf0). 
I recommend reading the updated PR description: https://github.com/Unleash/unleash-edge/pull/474#issue-2388193722

I think dynamic/strict behaviors are nice terms for this. Definitely better than open/closed modes.

I agree that we should introduce this gracefully. The default is now the "dynamic" behavior, so this means no breaking changes to existing users. However we're introducing the "strict" behavior, which is our new recommended way of starting Edge. We can have our quickstart command include `--strict` when we tackle the docs in a separate PR.

The idea is that sometime later we drop this in favor of always defaulting to the "strict" behavior instead.

WDYT?

server/src/error.rs
```diff
@@ -204,6 +207,7 @@ impl Display for EdgeError {
             EdgeError::NotReady => {
                 write!(f, "Edge is not ready to serve requests")
             }
+            EdgeError::InvalidTokenOnClosedMode => write!(f, "Edge is running in closed mode and the token is not subsumed by any registered tokens"),
```
FredrikOseberg, line 210: This needs to be updated to strict behavior I assume?

FredrikOseberg (commented):
This looks good to me. I found a small remnant of the old naming.

gardleopard:
Are we not gonna put strict as the default behaviour and deprecate but support the dynamic behaviour?

gardleopard (approved):

server/src/error.rs
nunogois: Yes, nice catch! https://github.com/Unleash/unleash-edge/pull/474/commits/136373471997b062052369cabb8de68eb879b324

PR #476, opened 2024-07-04
chore: add deprecation logs for dynamic behavior

nunogois:
https://linear.app/unleash/issue/2-2385/add-deprecation-logs-to-dynamic-behavior

Adds deprecation logs for dynamic behavior.

This should cover phase 1 of our [migration plan](https://linear.app/unleash/document/migration-plan-b3d7f35abc4e)

No explicit behavior:
image

Explicitly dynamic:
image

Explicitly strict:
image

PR #477, opened 2024-07-05
docs: strict and dynamic mode

chriswk:


CLI.md
```diff
@@ -48,7 +48,7 @@ This document contains the help content for the `unleash-edge` command-line prog
   Default value: `3043`
 * `--instance-id <INSTANCE_ID>` — Instance id. Used for metrics reporting
 
-  Default value: `01HX9QXY3MNKKAMP7XRS6RG2Q4`
+  Default value: `01J20TW2327N1JHE5XYHBE55KM`
```
nunogois, line 51: Hmm won't this be a new ulid every time?
```suggestion
  Default value: Automatically generated ulid.
```

I guess the reason for this, and especially this change in the value, is that we just copy paste the CLI output 🙂

chriswk: Yup, good point. I'll update the default.

chriswk: Hmm, no I won't since this is fully auto generated. For now we'll just accept that this will update each time we generate CLI.md

chriswk: Hmm, I changed it to <GENERATED_ULID> now. We should see if we can have our generator replace that value automatically

README.md
```diff
@@ -43,6 +43,41 @@ Running Edge in Docker with our recommended setup:
 docker run -it -p 3063:3063 -e STRICT=true -e UPSTREAM_URL=<yourunleashinstance> unleashorg/unleash-edge:<mostrecentversion> edge
 ```
 
+## Strict behavior (Since 19.2.0)
```
nunogois, line 46:
```suggestion
## Edge behaviors

As of version 19.2.0, Unleash Edge now supports two behaviors when running in edge mode: **strict** and **dynamic**. We recommend adopting the new **strict** behavior, while **dynamic** remains as a legacy option that will be deprecated and removed in a future release.

For legacy reasons, **dynamic** behavior is still the default. However, a warning will be logged at startup to indicate its deprecation.

Please note that these behaviors are mutually exclusive.

### Strict behavior
```

CLI.md
chriswk: For now we'll have to remember to replace the value with the placeholder again if we regenerate CLI.md

README.md
```diff
@@ -43,6 +43,41 @@ Running Edge in Docker with our recommended setup:
...
+because that would change the scope of access.
+
+Strict behavior still allows using tokens (both frontend and client) with narrower access than what the tokens it was
+started with has.
+
+E.g. If you start with one wildcard token with access to the development
+environment `*:development.<somelongrandomstring>` and your clients use various tokens with access to specific projects
+in the development environment, Edge will filter features to only grant access to the narrower scope.
+
+## Dynamic behavior
```
nunogois, line 75: In case we adopt https://github.com/Unleash/unleash-edge/pull/477/files#r1666502705

```suggestion
### Dynamic behavior
```

README.md
```diff
@@ -43,6 +43,41 @@ Running Edge in Docker with our recommended setup:
...
+
+With dynamic behavior, Edge behaves as it has since v1.0.0, any new client tokens are validated against upstream and if
```
nunogois, line 77:
```suggestion
With dynamic behavior, Edge behaves as it has since v1.0.0. Any new client tokens are validated against upstream and if
```

README.md
```diff
@@ -43,6 +43,49 @@ Running Edge in Docker with our recommended setup:
...
+
+Please note that these behaviors are mutually exclusive.
+
+### Strict behavior
+
+If started with the `--strict` flag or the `STRICT` environment variable, Edge now starts with strict behavior and must
+be
+given
+tokens at startup.
+Edge will lock down access for new tokens that have a wider or different access scope than the tokens it was started
```
chriswk, line 60:
```suggestion
Edge will refuse access for new tokens that have a wider or different access scope than the tokens it was started
```

README.md
```diff
@@ -43,6 +43,41 @@ Running Edge in Docker with our recommended setup:
...
+found
+to be valid, a refresh job will be configured with the minimum set of tokens that are able to fetch all projects and
+environments Edge has seen. Set with `--dynamic` or the `DYNAMIC` environment variable. (19.2.0 / July 4th 2024): We're
+looking to deprecate this
+behavior. If your company needs this behavior, reach out to us on Slack, the final decision has not been made yet. In
+19.2.0
+this behavior is still the default, but Edge will log that you should make a choice between dynamic or strict behavior.
+If
+you
+explicitly choose dynamic behavior, Edge will warn about the planned deprecation.
```
nunogois, line 87: Even if you don't explicitly choose it and we default to it, we still show the warning. It behaves as shown in the PR description of https://github.com/Unleash/unleash-edge/pull/476

nunogois (approved):
Maybe some things in the behavior descriptions could be rephrased for clarity, but I think this is good for now 👍

CLI.md
nunogois: Yep, makes sense 👍 Either way just something I noticed that is out of scope for the PR, so no worries.

README.md
chriswk: That's true.

chriswk: Agreed. Changed text to `When dynamic behavior is selected (by default or by choice), Edge will print a warning about the planned deprecation`

CLI.md
```diff
@@ -32,7 +32,8 @@ This document contains the help content for the `unleash-edge` command-line prog
 * `-b`, `--base-path <BASE_PATH>` — Which base path should this server listen for HTTP traffic on
 
   Default value: ``
-* `-w`, `--workers <WORKERS>` — How many workers should be started to handle requests. Defaults to number of physical cpus
+* `-w`, `--workers <WORKERS>` — How many workers should be started to handle requests. Defaults to number of physical
+  cpus
```
sighphyre, line 36: Doesn't really affect the markdown but what's going on here? IntelliJ enforcing a 120 character line?

README.md
```diff
@@ -43,6 +43,47 @@ Running Edge in Docker with our recommended setup:
...
+
+Please note that these behaviors are mutually exclusive.
+
+### Strict behavior
+
+If started with the `--strict` flag or the `STRICT` environment variable, Edge now starts with strict behavior and must
+be
+given
+tokens at startup.
+Edge will lock down access for new tokens that have a wider or different access scope than the tokens it was started
```
sighphyre, line 60: I think this is cool but I'd like to get a bit more specific in terms of what access means here. We do a nice job providing examples but I think we can do even better just by explaining what this is.

Let's lift this:

> Strict behavior still allows using tokens (both frontend and client) with narrower access than what the tokens it was started with has.

And combine it with this section and then build on that. Something like...?

Edge will refuses requests from SDKs that have a wider or different access scope than the initial tokens. Specifically, this means incoming requests must have a token that exactly matches the environment and is bound to a project or projects specified in that token.

CLI.md
chriswk: Looks like it

README.md
sighphyre: I think we can also tell a story on why this is valuable. Struggling to not go "full marketer" here but something like "this gives you more predictable control over which SDKs have access to Edge and prevents you from expanding the scope at runtime accidentally"

README.md
```diff
@@ -43,6 +43,47 @@ Running Edge in Docker with our recommended setup:
...
+with.
+
+E.g. if you start with one wildcard token with access to development `*:development.<somelongrandomstring` and later a
+client comes with a
+token accessing a different environment, say production `*:production.<somedifferentrandomstring>` Edge will return 403,
+because that would change the scope of access.
+
+Strict behavior still allows using tokens (both frontend and client) with narrower access than what the tokens it was
```
sighphyre, line 68: I think we could take this and the paragraph below out

README.md
```diff
@@ -43,6 +43,47 @@ Running Edge in Docker with our recommended setup:
...
+in the development environment, Edge will filter features to only grant access to the narrower scope.
+
+### Dynamic behavior
+
+With dynamic behavior, Edge behaves as it has since v1.0.0. Any new client tokens are validated against upstream and if
+found
+to be valid, a refresh job will be configured with the minimum set of tokens that are able to fetch all projects and
+environments Edge has seen. Set with `--dynamic` or the `DYNAMIC` environment variable. (19.2.0 / July 4th 2024): We're
+looking to deprecate this
+behavior. If your company needs this behavior, reach out to us on Slack, the final decision has not been made yet. In
```
sighphyre, line 82: Just feels a bit more friendly

```suggestion
behavior. If you need this behavior, reach out to us on Slack, the final decision has not been made yet. In
```

sighphyre (approved):
I like it! Left some thoughts but I think this tells the story we need it to tell

PR #478, opened 2024-07-11
fix: add timeout for redis read and write ops.

chriswk:
Previously, if you gave us a redis url that was valid, but used the wrong protocol (so rediss when connecting to insecure, or redis when connecting to secure), operations against Redis got stuck. This PR adds a timeout setting (setting it to default of 2000 ms, but configurable).

This does add time when url is completely invalid (i.e. there's nothing listening), but it prevents the client from getting stuck indefinitely if something answers, but redis can't get an answer.


gastonfournier (approved):
LG!

PR #479, opened 2024-07-11
task: Dependency maintenance

chriswk:
So, this bumps us to most recent version of actix, tokio, opentelemetry and rustls to keep on top of maintenance.

It changed a couple of the metric names for opentelemetry as well as change how rustls was init'ed so updated the TLS code to make sure we stay on ring, since I couldn't get the new aws-lc-rs crypto provider to work without installing more buildtools. 

Maybe by 0.25 of rustls this will be smoother.


server/src/client_api.rs
```diff
@@ -642,7 +641,10 @@ mod tests {
             .send()
             .await;
         assert!(status.is_ok());
-        assert_eq!(status.unwrap().status(), StatusCode::FORBIDDEN);
+        assert_eq!(
+            status.unwrap().status().as_u16(),
+            StatusCode::FORBIDDEN.as_u16()
```
gastonfournier, line 646: I see that in line 580 StatusCode is prefixed with actix_http but not on this line, is that intended? 
Is it because this StatusCode comes from `reqwest::StatusCode`?

server/src/metrics/actix_web_metrics.rs
```diff
@@ -158,28 +144,28 @@ pub(super) fn metrics_attributes_from_request(
...
-    attributes.push(SERVER_SOCKET_ADDRESS.string(conn_info.host().to_string()));
-    attributes.push(URL_PATH.string(http_target.to_owned()));
+    attributes.push(KeyValue::new(SERVER_ADDRESS, conn_info.host().to_string()));
+    attributes.push(KeyValue::new(URL_PATH, http_target.to_owned()));
     attributes.push(KeyValue::new(URL_SCHEME, http_scheme(conn_info.scheme())));
 
     let server_name = req.app_config().host();
     if server_name != conn_info.host() {
-        attributes.push(SERVER_ADDRESS.string(server_name.to_string()));
+        attributes.push(KeyValue::new(SERVER_ADDRESS, server_name.to_string()));
```
gastonfournier, line 153: Are we intentionally overriding SERVER_ADDRESS from line 147? I think we are using the server_name and only if not present we use conn_info.host(). If they're equal we could just use server_name and remove line 147... I think we can remove line 147 and remove the if in line 152

gastonfournier (approved):
LG! Just some minor details but no blocker

server/src/client_api.rs
chriswk: Yeah, bumping to actix_http 4.8 seems to make those two separate, I've tried to stick to reqwest status codes for the client and actix for the server (since actix wants to return its own status codes, not reqwests). They should be interchangeable, the http status code spec is complete.

server/src/metrics/actix_web_metrics.rs
chriswk: Good point, this is copied from an example, so I guess I haven't really thought this through. Something to improve on when needed

PR #489, opened 2024-07-19
feat: upgrade types library to add new metrics data and backwards compat for impression_data

sighphyre:


chriswk (approved):
Maybe we should've implemented default for MetricsMetadata?

Not a blocker. Here's a :+1:

sighphyre:
> Maybe we should've implemented default for MetricsMetadata?
> 
> Not a blocker. Here's a 👍

I did think about it, I kinda want this to be a conscious choice. If it turns out its a pain we add it later

PR #492, opened 2024-07-19
fix: trim tokens when parsing

chriswk:


sighphyre (approved):
Yep, just a `trim` in places! LGTM

PR #508, opened 2024-08-14
fix(redis): handle uninitialized/nil redis key for unleash-tokens

Defman:
## About the changes
This change handles the case where the redis backend have not yet been initialized.

Closes #

### DIFF
--- a/server/src/persistence/redis.rs
+++ b/server/src/persistence/redis.rs
@@ -87,8 +87,9 @@ impl EdgePersistence for RedisPersister {
                 conn.get(TOKENS_KEY)?
             }
         };
-        serde_json::from_str::<Vec<EdgeToken>>(&raw_tokens)
+        serde_json::from_str::<Option<Vec<EdgeToken>>>(&raw_tokens)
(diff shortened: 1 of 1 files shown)


sighphyre:
Hey @Defman, I'm not 100% sure what this solves. If I'm reading the code right, this block [here](https://github.com/Unleash/unleash-edge/blob/933659b0461b8a53cb89caf0984963918fe82b04/server/src/builder.rs#L52), should already be handling this case.

Would you be able to add a test case to show what this solves?

Defman:
You are right, the error that I encounted just happend after the above warning was thrown. You are correct in that the edge currently handles this case. However it does not differentiate between deserialization errors caused by nil or other more severe deserialization errors.

sighphyre:
> However it does not differentiate between deserialization errors caused by nil or other more severe deserialization errors.
Yep, in practice there isn't a lot that Edge can do to recover from not being able to load tokens from Redis, outside of defaulting to an empty list. Completely open to some more fine grained error messaging if you'd like to open a PR though

PR #514, opened 2024-09-04
test: modify to validate edge accepts non strings

gastonfournier:
Many SDKs can send arbitrary data inside context.properties and edge should not fail. Ignoring that data seems the simplest thing to do, but maybe we should do something else (e.g. the number could be converted to string).


nunogois (approved):
LGTM - Maybe we should consider converting them to strings?
Possibly relevant: https://github.com/Unleash/unleash-client-node/pull/628

chriswk:
Superseded by David's fix in 19.9.1. Closing this issue

Issue #509, opened 2024-08-30
libssl.so.1.1: cannot open shared object file: No such file or directory


mraymond:
### Describe the bug

I'm attempting to set up a unleash-edge within a docker image built on a newer OS(Debian 12.6), and when starting the container, it returns the error
`/unleash-edge: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

Example Dockerfile, with latest binary(19.3.0) in same directory:
```
FROM debian:12.6

COPY ./ /

ENTRYPOINT ["/unleash-edge"]
```

### Steps to reproduce the bug

1. Setup Dockerfile using later versions of OSes(Debian 12) and unleash-edge binary in same folder
2. Run Dockerfile

### Expected behavior

Should start up unleash-edge without errors

### Logs, error output, etc.

```bash
/unleash-edge: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

### Unleash version

19.3.0

### Subscription type

Open source

### Hosting type

Self-hosted

chriswk:
Hi @mraymond - With newer debians, debian uses openssl 3.0, I think we're still built against libssl (openssl 1.1). I'll see if I can't get a build out with openssl 3.0. Thanks for reporting this.

chriswk:
I'm assuming you downloaded the binary from the repository here, that's still built with openssl 1.1.

I'd probably make a builder using debian 12 as well. I'll make a PR with a new Dockerfile with Openssl 3.0

mraymond:
Yep, confirming the binaries I was trying were from the releases on Github. Thanks for the future update.

PR #516, opened 2024-09-05
feat: add debian 12 build file

chriswk:
Also made the dockerignore file just ignore the target folder except for the binaries. Which will allow us to have a docker image that does the building.

The debian12 recipe uses cargo chef to have better layer caching, as well as debian:bookworm-slim as the base image for runtime, which is slightly bigger than distroless.

However, it does not build for both amd64 and aarch yet

Should help for #509

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
-*
+target
 !target/aarch64-unknown-linux-gnu/release/unleash-edge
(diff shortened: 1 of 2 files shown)


chriswk:
Bah, this is just the first step, because we need the binary  attached to our releases as well. I'll extend this PR with the work needed to have an openssl 3.0 compatible binary.

PR #531, opened 2024-09-17
fix: Update to explicitly type RedisError.

chriswk:
Work around for Rust 2024 implementing Never type (!)

https://github.com/rust-lang/rust/issues/123748


nunogois (approved):
LGTM

PR #548, opened 2024-10-01
task: remove request tracing

chriswk:
Originally we wanted to support OpenTelemetry for tracing, but we no longer support setting up an OTEL_COLLECTOR, so removing the request tracer seems sensible.

If we ever get around to adding an opentelemetry feature, this can be added back in.


sighphyre (approved):
Nice

PR #557, opened 2024-10-09
fix: remove bucket creation in s3 persister

sighphyre:
Having the creation like this breaks buckets in regions other than us-east-1


gardleopard (approved):
LGTM

PR #558, opened 2024-10-09
fix: frontend all endpoints now accept context correctly

### LINKED ISSUE #555: Context properties are not evaluated when using /api/proxy/all
### Describe the bug

Contextual properties should be evaluated in the same way regardless of the routing /api/proxy/all, /api/proxy or /api/frontend

1. Create a strategy that uses context properties such as the ones defined in the test file `with_custom_constraint.json`

2. Call the API using `/api/proxy/all?properties[companyId]=bricks`

3. The strategy will not match the output of `/api/frontend?properties[companyId]=bricks` or `/api/proxy?properties[companyId]=bricks`

### Expected behavior

Contextual properties from /api/proxy/all should be evaluated in the same way than /api/frontend and /api/proxy.
In that case the flag should be `enabled` when `companyId=bricks`

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

# Client impact
/api/proxy/all is mandatory for Dart SDK users to be able to rely on edge to fetch unleash flags and ensure metrics propagation 

# How to reproduce the issue usings tests
The test `can_handle_custom_context_fields` in `frontend_api.rs` does not check the flag `enabled` property.
This code allows to see the difference between the different endpoints.

```
        let req = test::TestRequest::get()
            .uri("/api/frontend?properties[companyId]=bricks")
            .insert_header(ContentType::json())
            .insert_header(("Authorization", auth_key.clone()))
            .to_request();
        let no_escape: FrontendResult = test::call_and_read_body_json(&app, req).await;
        assert!(no_escape.toggles.iter().all(|toggle| toggle.enabled));

        let req = test::TestRequest::get()
            .uri("/api/proxy?properties[companyId]=bricks")
            .insert_header(ContentType::json())
            .insert_header(("Authorization", auth_key.clone()))
            .to_request();
        let no_escape: FrontendResult = test::call_and_read_body_json(&app, req).await;
        assert!(no_escape.toggles.iter().all(|toggle| toggle.enabled));

        let req = test::TestRequest::get()
        .uri("/api/proxy/all?properties[companyId]=bricks")
        .insert_header(ContentType::json())
        .insert_header(("Authorization", auth_key.clone()))
        .to_request();
        let no_escape: FrontendResult = test::call_and_read_body_json(&app, req).await;
        assert!(no_escape.toggles.iter().all(|toggle| toggle.enabled));
```
Note: The flags are evaluated correctly with the custom properties using POST calls but not on the ones that rely on GET such as the ones from the Flutter / Dart SDK.

### Unleash version

v19.5.0

### Subscription type

Enterprise

### Hosting type

Self-hosted

curl and Dart SDK

sighphyre:
Moves the all endpoints for the frontend API to use the same context parsing path as the non all endpoints. Fixes an issue where context properties would not be correctly considered in the feature evaluation

Closes #555

### DIFF
--- a/server/src/frontend_api.rs
+++ b/server/src/frontend_api.rs
@@ -20,7 +20,6 @@ use unleash_types::{
 };
 use unleash_yggdrasil::{EngineState, ResolvedToggle};
 
-use crate::error::EdgeError::ContextParseError;
 use crate::types::{ClientIp, IncomingContext, PostContext};
 use crate::{
     error::{EdgeError, FrontendHydrationMissing},
@@ -48,13 +47,14 @@ pub async fn get_proxy_all_features(
     edge_token: EdgeToken,
     engine_cache: Data<DashMap<String, EngineState>>,
     token_cache: Data<DashMap<String, EdgeToken>>,
+    context: QsQuery<IncomingContext>,
     req: HttpRequest,
 ) -> EdgeJsonResult<FrontendResult> {
     get_all_features(
         edge_token,
         engine_cache,
         token_cache,
-        req.query_string(),
+        &context.into_inner().into(),
         req.extensions().get::<ClientIp>(),
     )
 }
@@ -75,13 +75,14 @@ pub async fn get_frontend_all_features(
     edge_token: EdgeToken,
     engine_cache: Data<DashMap<String, EngineState>>,
     token_cache: Data<DashMap<String, EdgeToken>>,
+    context: QsQuery<IncomingContext>,
     req: HttpRequest,
 ) -> EdgeJsonResult<FrontendResult> {
     get_all_features(
         edge_token,
         engine_cache,
         token_cache,
-        req.query_string(),
+        &context.into_inner().into(),
         req.extensions().get::<ClientIp>(),
     )
 }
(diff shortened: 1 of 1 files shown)


server/src/frontend_api.rs
```diff
@@ -746,20 +747,16 @@ pub fn get_all_features(
...
-    let context: Context = raw_context.into();
     let context_with_ip = if context.remote_address.is_none() {
         Context {
             remote_address: client_ip.map(|ip| ip.to_string()),
-            ..context
+            ..context.clone()
         }
     } else {
-        context
+        context.clone()
```
gastonfournier, line 759: I wonder if we do have to clone here, there's no risk other than a bit more of memory allocation

gastonfournier (approved):
LG!

server/src/frontend_api.rs
sighphyre: The clone here is pretty light, in this context it's probably over optimizing to fix given the amount of memory that's consumed in processing an HTTP request

...but it looks dumb and it's easy to fix so yeah, I took it out! Good eyes!

PR #563, opened 2024-10-16
feat: Support prometheus remote write protocol for metrics

chriswk:


sighphyre (approved):
Noice! We can probably remove the "push" feature for prometheus tho

PR #564, opened 2024-10-16
feat: flush metrics on clean shutdown

sighphyre:
Flushes metrics to upstream on shutdown. Needs a bit of love and/or discussion before merge:

1) Be cool if this was tested by the same things that test the standard metrics send but I don't see those tests (currently tested by humans)
2) Is 5 parameters to a function too much? It feels like too much


chriswk (approved):
LGTM

PR #590, opened 2024-11-25
Add example docker command

melindafekete:
Readme: Add an example for a docker command to illustrate what an upstream URL looks like


sighphyre (approved):
Yes! LGTM!

PR #592, opened 2024-12-02
experimental: sse client support

thomasheartman:
this is experimental and a work in progress.


server/src/client_api.rs
```diff
@@ -36,6 +36,31 @@ pub async fn get_features(
...
+    let features = resolve_features_2(
+        query,
+        validated_token.clone(),
+        filter_set,
+        features_cache,
+        req.clone(),
+    )
+    .await;
+    match (req.app_data::<Data<FeatureRefresher>>(), features) {
+        (Some(refresher), Ok(features)) => Ok(refresher.broadcaster.new_client(features).await),
```
thomasheartman, line 59:
```suggestion
         // we should pass in query, filter_set, and validated_token to store in the broadcaster, I think
        (Some(refresher), Ok(features)) => Ok(refresher.broadcaster.new_client(features).await),
```

server/Cargo.toml
```diff
@@ -29,6 +29,7 @@ actix-http = "3.9.0"
 actix-middleware-etag = "0.4.2"
 actix-service = "2.0.2"
 actix-web = { version = "4.9.0", features = ["rustls-0_23", "compress-zstd"] }
+actix-web-lab = "0.23.0"
```
kwasniew, line 32: what do we use it for?

server/src/http/feature_refresher.rs
```diff
@@ -299,6 +306,105 @@ impl FeatureRefresher {
...
     }
 
+    /// This is where we set up a listener per token.
+    pub async fn start_streaming_features_background_task(&self) {
+        let refreshes = self.get_tokens_due_for_refresh();
+        for refresh in refreshes {
+            let token = refresh.token.clone();
+            let streaming_url = format!("{}/client/streaming", self.unleash_client.urls.api_url);
+
+            let es_client = match eventsource_client::ClientBuilder::for_url(&streaming_url) {
```
kwasniew, line 316: can we simplify the builder pattern here?
```rs
let es_client = es::ClientBuilder::for_url(url)?
        .header("Authorization", &token.token)?
        .reconnect(
            es::ReconnectOptions::reconnect(true)
                .retry_initial(true)
                .delay(Duration::from_secs(5))
                .backoff_factor(2)
                .delay_max(Duration::from_secs(30))
                .build(),
        )
        .build();
```

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,190 @@
...
+            loop {
+                interval.tick().await;
+                this.remove_stale_clients().await;
+            }
+        });
+    }
+
+    /// Removes all non-responsive clients from broadcast list.
+    async fn remove_stale_clients(&self) {
+        let active_connections = self.inner.lock().active_connections.clone();
```
kwasniew, line 87: we obtain the lock only on this line. By the time we get to line 102 someone in a race condition may change the active_connections. Maybe the lock should be in a variable scoped to the whole method?

thomasheartman: quite possibly! there's also the possibility that there's a much better data structure to use for this 💁🏼

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,190 @@
...
+                group.clients.push(tx.clone());
+            })
+            .or_insert(ClientGroup {
+                clients: vec![tx.clone()],
+                filter_set,
+                token,
+            });
+
+        tx.send(
+            sse::Data::new_json(features)
```
kwasniew, line 137: I checked and `data.into()` can be also expressed as more explicit and verbose `sse::Event::Data(sse::Data::new("unleash-connected"))`  but probably it is less idiomatic in Rust

server/src/http/feature_refresher.rs
```diff
@@ -299,6 +311,97 @@ impl FeatureRefresher {
         }
     }
 
+    /// This is where we set up a listener per token.
+    #[cfg(feature = "streaming")]
+    pub async fn start_streaming_features_background_task(&self) -> anyhow::Result<()> {
```
thomasheartman, line 316: Because this came up as a question during code walkthrough in the spike: I'm using anyhow here to allow us to easily return errors to use the `?` operator. We should probably make it an EdgeResult or something instead, but hey, it's a spike.

server/src/http/broadcaster.rs
chriswk: When it comes to data_structure. I believe the answer might be in the data structure used in the rest of Edge. The DashMap. It's meant for concurrent work.

thomasheartman: Yeah, yeah; that's what I was thinking of. But it wasn't used in the example and I didn't wanna start mucking about with it under the time pressure we were looking at last week. I'll see if I can't find some time to try it out before the review sesh tomorrow.

server/src/http/feature_refresher.rs
```diff
@@ -299,6 +311,97 @@ impl FeatureRefresher {
...
+                                    // send the data to the broadcaster. This should probably just send the new
+                                    // feature set OR even just a "filter flags"
+                                    // function. The broadcaster will take care
+                                    // of filtering the flags per listener.
+                                    // let data = Data::new(event.data).event("unleash-updated");
+                                    // broadcaster.rebroadcast(actix_web_lab::sse::Event::Data(data)).await;
+                                    refresher.broadcaster.broadcast().await;
+                                }
+                                eventsource_client::SSE::Event(event) => {
+                                    debug!(
```
thomasheartman, line 381: let's bump this to info! and set rust log for Edge to info. (for monitoring in prod)

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use crate::{
+    filters::{filter_client_features, name_prefix_filter, project_filter, FeatureFilterSet},
+    tokens::cache_key,
+    types::{EdgeToken, FeatureFilters},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+struct QueryWrapper {
+    query: FlagQuery,
```
thomasheartman, line 33: check if the hashes feature is used for unleash-types. It should be hashable.

(update unleash types to impl hash)

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,190 @@
...
+            });
+
+        tx.send(
+            sse::Data::new_json(features)
+                .unwrap()
+                .event("unleash-connected")
+                .into(),
+        )
+        .await
+        .unwrap();
```
thomasheartman, line 143: return edge error

server/src/http/feature_refresher.rs
```diff
@@ -299,6 +311,97 @@ impl FeatureRefresher {
...
+                                eventsource_client::SSE::Connected(_) => {
+                                    debug!("SSE Connection established");
+                                }
+                                eventsource_client::SSE::Comment(_) => {
+                                    // purposefully left blank.
+                                },
+                            }
+                        }
+                    })
+                    .map_err(|e| warn!("Error in SSE stream: {:?}", e));
```
thomasheartman, line 395: handle 404s etc, maybe launch polling if can't connect

chriswk (approved):
After the Google meet review, I'm comfortable with giving you a :+1: on this one. Merge when you feel you've made enough of the nitpick changes.

I'm ready for a pairing session on getting some tests in place later this week or next.

server/src/http/broadcaster.rs
thomasheartman: yea, "into" is pretty idiomatic Rust code, so I'm gonna leave this as is.

server/src/http/feature_refresher.rs
thomasheartman: This is a more complex topic. I suggest we move handling this one to a little later.

server/src/http/broadcaster.rs
nunogois: I'm guessing we decided to leave this one for later?

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,207 @@
...
+        });
+
+        Broadcaster::spawn_ping(Arc::clone(&this));
+
+        this
+    }
+
+    /// Pings clients every 30 seconds to see if they are alive and remove them from the broadcast
+    /// list if not.
+    fn spawn_ping(this: Arc<Self>) {
```
nunogois, line 67:
```suggestion
    fn spawn_heartbeat(this: Arc<Self>) {
```

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,207 @@
...
+
+            loop {
+                interval.tick().await;
+                this.remove_stale_clients().await;
+            }
+        });
+    }
+
+    /// Removes all non-responsive clients from broadcast list.
+    async fn remove_stale_clients(&self) {
```
nunogois, line 79: I see this method's main responsibility as the "heartbeat". The fact that it removes non-sendable clients from the list feels like a detail. Like in Unleash's implementation.

```suggestion
    async fn heartbeat(&self) {
```

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,207 @@
...
+            // ok_clients.iter().filter(|client| client.token_is_valid())
+
+            group.clients = ok_clients;
+        }
+    }
+
+    /// Registers client with broadcaster, returning an SSE response body.
+    /// The current impl takes the feature set as input and sends it to the client as a connected event.
+    ///
+    /// The commented-out arguments are what we'll need to store per client so
```
nunogois, line 103: I see no commented-out arguments 😄

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,207 @@
...
+        match features {
+            Some(features) => Ok(Json(ClientFeatures {
+                query: Some(query),
+                ..features
+            })),
+            // Note: this is a simplification for now, using the following assumptions:
+            // 1. We'll only allow streaming in strict mode
+            // 2. We'll check whether the token is subsumed *before* trying to add it to the broadcaster
+            // If both of these are true, then we should never hit this case (if Thomas's understanding is correct).
+            None => todo!(),
```
nunogois, line 176: I agree with the assumptions, but should we return an error instead if this is ever reached?

server/src/http/broadcaster.rs
```diff
@@ -0,0 +1,207 @@
...
+
+            for client in &group.clients {
+                client_events.push((client.clone(), event.clone()));
+            }
+        }
+        // try to send to all clients, ignoring failures
+        // disconnected clients will get swept up by `remove_stale_clients`
+        let send_events = client_events
+            .iter()
+            .map(|(client, event)| client.send(event.clone()));
```
nunogois, line 203: Do we need this clone?

server/src/main.rs
```diff
@@ -149,31 +150,67 @@ async fn main() -> Result<(), anyhow::Error> {
...
+                    }
+                    _ = validator.schedule_validation_of_known_tokens(edge.token_revalidation_interval_seconds) => {
+                        tracing::info!("Token validator validation of known tokens was unexpectedly shut down");
+                    }
+                    _ = validator.schedule_revalidation_of_startup_tokens(edge.tokens, lazy_feature_refresher) => {
+                        tracing::info!("Token validator validation of startup tokens was unexpectedly shut down");
+                    }
+                    _ = metrics_pusher::prometheus_remote_write(prom_registry_for_write, edge.prometheus_remote_write_url, edge.prometheus_push_interval, edge.prometheus_username, edge.prometheus_password, app_name) => {
+                        tracing::info!("Prometheus push unexpectedly shut down");
+                    }
```
nunogois, line 213: Wish we could simplify this and reduce the duplicated code but I don't have any better ideas atm.

nunogois (approved):
I think you did an amazing job with this! 🥇

server/src/client_api.rs
```diff
@@ -36,6 +38,28 @@ pub async fn get_features(
 ) -> EdgeJsonResult<ClientFeatures> {
     resolve_features(edge_token, features_cache, token_cache, filter_query, req).await
 }
+
+#[cfg(feature = "streaming")]
+#[get("/streaming")]
+pub async fn stream_features(
```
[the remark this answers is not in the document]
thomasheartman: Sure. But as the API endpoint, maybe it should be more explicit? Feels like `stream_features` is clearer to me. But then again, it's a separate controller in Unleash. I don't care much either way, but this feels more explicit in this context.

server/src/error.rs
```diff
@@ -291,5 +295,13 @@ impl From<serde_json::Error> for EdgeError {
     }
 }
 
+#[cfg(feature = "streaming")]
+impl From<SendError<Event>> for EdgeError {
+    // todo: create better enum representation. use this is placeholder
+    fn from(_value: SendError<Event>) -> Self {
+        EdgeError::TlsError
```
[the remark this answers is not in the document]
thomasheartman: Yes, I agree. Hence the comment. I just didn't want to start implementing a new error enum member when cleaning this up. We should handle this as part of testing, because I don't think it is the right errer.

server/src/http/feature_refresher.rs
```diff
@@ -154,12 +164,14 @@ impl FeatureRefresher {
...
             tokens_to_refresh: Arc::new(DashMap::default()),
-            features_cache: features,
+            features_cache: features.clone(),
             engine_cache: engines,
             refresh_interval: features_refresh_interval,
             persistence,
             strict,
             app_name: app_name.into(),
+            #[cfg(feature = "streaming")]
+            broadcaster: Broadcaster::new(features.clone()),
```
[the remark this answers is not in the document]
thomasheartman: Smart 🧠

server/src/main.rs
thomasheartman: Yeah, I know. I hate this. But we can refactor it later. I don't think this'll be the final version (I hope not!).

server/src/http/broadcaster.rs
thomasheartman: Sure thing!

server/src/http/broadcaster.rs
thomasheartman: Yeah, I can agree with that. Renamed for now.

server/src/http/broadcaster.rs
thomasheartman: Good point! Removed.

server/src/http/broadcaster.rs
thomasheartman: yeah, let's do it. better to be on the safe side 🤷🏼

server/src/http/broadcaster.rs
thomasheartman: At the moment, yes we do. Because `iter` returns a pair of references we only get an `&Event`, and the send function needs to own the event. So I don't think it's possible to send these events without cloning the way we do it now.

As a potential refactor (which would probably be beneficial), we could:

- Loop through the groups and collect the right features for each.
- Loop through the clients in each group and generate the event for each client (because you can generate the event from a reference to the features).

There might be smarter ways too, but let's look at that as its own task 😄

server/src/http/broadcaster.rs
thomasheartman: Handling in https://github.com/Unleash/unleash-types-rs/pull/43

PR #595, opened 2024-12-06
fix: bootstrap reload in offline mode.

chriswk:
After adding support for client and frontend tokens, we did not extend the reloader to check client and frontend token Vecs, this PR extends tokens with FE and Client tokens, to ensure that we refresh the data for all our tokens.

In addition we make /internal-backstage/tokens useful for offline mode as well, to at least be able to see which tokens you added to Edge.

In addition, since offline mode did not set up a token validator, I extended the auth middleware to check if our offline token has a token type set and if they do, limit their access to what they were intended for (client for /api/client, frontend for /api/frontend | /api/proxy)

Fixes: #594

### DIFF
--- a/server/src/middleware/validate_token.rs
+++ b/server/src/middleware/validate_token.rs
@@ -14,7 +14,6 @@ pub async fn validate_token(
     req: ServiceRequest,
     srv: crate::middleware::as_async_middleware::Next<impl MessageBody + 'static>,
 ) -> Result<ServiceResponse<impl MessageBody>, actix_web::Error> {
-    trace!("Validating req: {}", req.path());
     let maybe_validator = req.app_data::<Data<TokenValidator>>();
     let token_cache = req
         .app_data::<Data<DashMap<String, EdgeToken>>>()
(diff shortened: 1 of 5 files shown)


server/src/middleware/validate_token.rs
```diff
@@ -27,18 +25,15 @@ pub async fn validate_token(
...
                         {
-                            trace!("Was allowed to access");
                             srv.call(req).await?.map_into_left_body()
                         } else {
                             req.into_response(HttpResponse::Forbidden().finish())
                                 .map_into_right_body()
                         }
                     }
                     Some(TokenType::Client) => {
-                        trace!("Got Client token validated {:?}", known_token);
```
nunogois, line 41: Any reason why we're getting rid of these now?

nunogois (approved):
LGTM

server/src/middleware/validate_token.rs
chriswk: We've verified that it triggers, and we have the /internal-backstage/tokens endpoint to see which tokens we know about, so cleaned it up.

DAGRSAG:
@chriswk Thanks for fixing both things. The question is, when is a new release with the fix available? Do you have an information for me?

PR #596, opened 2024-12-10
fix: bump opentelemetry dependencies to 0.27

chriswk:
So. Finally opentelemetry prometheus updated to 0.27, so we can move on from 0.24.

This PR updates all opentelemetry dependencies to 0.27.

### DIFF
--- a/server/src/metrics/actix_web_metrics.rs
+++ b/server/src/metrics/actix_web_metrics.rs
@@ -5,7 +5,7 @@ use actix_web::http::{Method, StatusCode, Version};
 use futures::{future, FutureExt};
 use futures_core::future::LocalBoxFuture;
 use opentelemetry::metrics::{Histogram, Meter, MeterProvider, UpDownCounter};
-use opentelemetry::{KeyValue, Value};
+use opentelemetry::{InstrumentationScope, KeyValue, Value};
 use opentelemetry_semantic_conventions::trace::{
     CLIENT_ADDRESS, HTTP_REQUEST_METHOD, HTTP_RESPONSE_STATUS_CODE, NETWORK_PROTOCOL_NAME,
     NETWORK_PROTOCOL_VERSION, SERVER_ADDRESS, SERVER_PORT, URL_PATH, URL_SCHEME,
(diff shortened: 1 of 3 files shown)


server/src/metrics/actix_web_metrics.rs
```diff
@@ -187,12 +185,7 @@ impl RequestMetricsBuilder {
 
 /// construct meters for this crate
 fn get_versioned_meter(meter_provider: impl MeterProvider) -> Meter {
-    meter_provider.versioned_meter(
-        "unleash_edge",
-        Some(env!("CARGO_PKG_VERSION")),
-        Some(opentelemetry_semantic_conventions::SCHEMA_URL),
-        None,
-    )
+    meter_provider.meter("unleash_edge")
```
gastonfournier, line 188: Shouldn't we have open telemetry schema url and cargo pkg version here?

I found some docs indicating we can do:
```rust
    meter_provider.meter_with_scope(InstrumentationScope::builder("unleash_edge")
        .with_version(env!("CARGO_PKG_VERSION"))
        .with_schema_url(opentelemetry_semantic_conventions::SCHEMA_URL)
        .build())
```
but maybe this change is intentional

server/src/metrics/actix_web_metrics.rs
```diff
@@ -331,10 +324,7 @@ impl PrometheusMetricsHandler {
         let encoder = TextEncoder::new();
         let metric_families = self.registry.gather();
         let mut buf = Vec::new();
-        if let Err(err) = encoder.encode(&metric_families[..], &mut buf) {
-            global::handle_error(MetricsError::Other(err.to_string()));
-        }
-
+        let _ = encoder.encode(&metric_families[..], &mut buf);
```
gastonfournier, line 327: Are we missing here the global handle error call?

chriswk: I removed it because it wasn't being exposed anymore

server/src/metrics/actix_web_metrics.rs
chriswk: No, you found what I was looking for. Updated to use your suggestion. Thanks :)

gastonfournier (approved):
Cool, LGTM!

PR #600, opened 2024-12-12
chore:add basic streaming tests

thomasheartman:
Adds a happy-path e2e test for streaming from an upstream server via edge to a client. 

Or at least, it does it conceptually. We've changed how the broadcaster is set up, so it actually breaks the functionality for now.

We should set it up so that we test actual Edge and not a mock impl, I think.


chriswk (approved):
This is LGTM :)

PR #605, opened 2024-12-17
feat: streaming now enabled/disabled with runtime flag

chriswk:


thomasheartman (approved):

server/src/http/feature_refresher.rs
```diff
@@ -86,25 +86,47 @@ fn client_application_from_token_and_name(
...
             features_cache,
             engine_cache: engines,
-            refresh_interval: features_refresh_interval,
+            refresh_interval: config.features_refresh_interval,
             persistence,
-            strict,
-            app_name: app_name.into(),
+            strict: config.strict,
+            streaming: config.streaming,
+            app_name: config.app_name.into(),
```
thomasheartman, line 129: Do we need the into here? We're not using config anymore, so we should be able to just take it, yeah? I might be wrong, of course.

```suggestion
            app_name: config.app_name,
```

thomasheartman (approved):

server/src/builder.rs
```diff
@@ -258,15 +259,18 @@ async fn build_edge(args: &EdgeArgs, app_name: &str) -> EdgeResult<EdgeInfo> {
         unleash_client: unleash_client.clone(),
         persistence: persistence.clone(),
     });
-
+    let feature_config = FeatureRefreshConfig::new(
+        Duration::seconds(args.features_refresh_interval_seconds as i64),
+        args.streaming,
+        args.strict,
```
thomasheartman, line 265: I think this is wrong. The argument order is strict first, then streaming.

So yeah, can we model this differently? E.g. an enum that's `dynamic  | strict | streaming`? (although they're not all edge modes, they cover all the cases here. It also prevents the "strict = false, streaming = true" issue, which should be invalid, right?

thomasheartman (approved):

server/src/builder.rs
```diff
@@ -259,10 +259,14 @@ async fn build_edge(args: &EdgeArgs, app_name: &str) -> EdgeResult<EdgeInfo> {
         unleash_client: unleash_client.clone(),
         persistence: persistence.clone(),
     });
+    let refresher_mode = match (args.strict, args.streaming) {
+        (_, true) => FeatureRefresherMode::Streaming,
+        (true, _) => FeatureRefresherMode::Strict,
+        _ => FeatureRefresherMode::Dynamic,
+    };
```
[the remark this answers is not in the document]
chriswk: Yeah, LGTM

PR #608, opened 2024-12-18
docs: notice about feature parity with unleash-server

### LINKED ISSUE #601: Filter /api/client/features by tag query parameter

I would like to request support for the `tag` query parameter in the `/api/client/features` endpoint for Unleash Edge

### Background

Currently, the `tag` parameter is supported by Unleash but this functionality is not yet available in Unleash Edge

I saw a previous issue with fix for `namePrefix ` #149 and it was mentioned:
>Edge can't support filtering by tags, since that is not data we get from Unleash (it's not part of the client features contract), and Edge does not have access to the database

Despite mentioned issue adding support for the tag parameter would help bring consistency between the Unleash API and Edge API providing a seamless experience for users

Additionally i can't find mentions in the docs saying that the Edge API could work differently from the Unleash API. If there are other differences it would be really helpful to include these in the docs and create a list of unsupported features to assist users in understanding the limitations of the Edge

### Solution suggestions

Possible solutions could be extending current API or create different endpoint that includes tags information

If these options couldn't be implemented in nearest future it would be great to note this details in docs

gastonfournier:
## About the changes
Unleash edge does not keep feature parity with unleash-server. Despite most of the features are currently supported, some are not and we don't have plans to support them.

Closes #601

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@
 
 Unleash Edge is a fast and lightweight proxy layer between your Unleash API and SDKs. It acts as a read replica of your Unleash instance and is designed to help you scale Unleash. It allows you to support thousands of connected SDKs without increasing the number of requests you make to your Unleash instance.
 
+**Important notice:** Unleash edge does not keep 100% feature parity with Unleash server. In particular filtering by tag or feature prefix is not supported.
+
 Edge supports both client-side and server-side SDKs and has multi-environment and project awareness. You can daisy-chain Edge instances to support more complex setups, such as multi-cloud deployments.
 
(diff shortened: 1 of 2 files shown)


melindafekete:
@gastonfournier can I take over this PR? 🙏 I'd rather introduce a new section called **Versioning and availability**, and go into a bit more detail

gastonfournier:
> @gastonfournier can I take over this PR? 🙏 I'd rather introduce a new section called **Versioning and availability**, and go into a bit more detail

Sure!! I appreciate it :)

PR #609, opened 2024-12-18
task: added gauge for showing number of streaming clients connected

chriswk:
Added a single label here (environment).

Do we need anything else?


chriswk:
Label won't work, we can have multiple tokens fetching for an environment. Will move back to simply count

thomasheartman (approved):

PR #613, opened 2024-12-20
chore(1-3216): send the same headers in streaming as for polling

thomasheartman:
This pr adds the same headers to the streaming clients as we use for polling clients.


chriswk (approved):
Pair pgorammed

PR #615, opened 2024-12-20
chore(1-3217): run SSE listener in infinite loop

thomasheartman:
Instead of using `while let Ok(...)` To run the listening loop, we'll just use a plain old loop. This is an experiment to see if this'll make it always reconnect succesfully.


chriswk (approved):
LGTM

PR #616, opened 2024-12-20
feat: start using instance_id argument for instance_id header

### LINKED ISSUE #614: Edge's instance IDs should be unique (and configurable?)

As of today, we use the constant `unleash_edge` as the instanced id (which is also not the same as `unleash-edge`, which is the app name). We should make it so that each instance has a different id be default.

### Background

_No response_

### Solution suggestions

_No response_

chriswk:
Closes #614

### DIFF
--- a/server/src/http/feature_refresher.rs
+++ b/server/src/http/feature_refresher.rs
@@ -276,8 +276,7 @@ impl FeatureRefresher {
     /// This is where we set up a listener per token.
     pub async fn start_streaming_features_background_task(
         &self,
-        app_name: String,
-        instance_id: String,
+        client_meta_information: ClientMetaInformation,
         custom_headers: Vec<(String, String)>,
     ) -> anyhow::Result<()> {
         use anyhow::Context;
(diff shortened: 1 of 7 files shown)


server/src/http/feature_refresher.rs
```diff
@@ -277,6 +277,7 @@ impl FeatureRefresher {
     pub async fn start_streaming_features_background_task(
         &self,
         app_name: String,
+        instance_id: String,
         custom_headers: Vec<(String, String)>,
```
thomasheartman, line 281: Ooh, this is a little sticky again. Having two very similar arguments with very similar values of the same type next to each other is a little dodgy. It'll be easy to flip them around. Can we chuck 'em in a config struct or use the type system to differentiate them?

thomasheartman (approved):

server/src/http/feature_refresher.rs
chriswk: Changes this to also take the meta_information. Don't know why I hadn't already done that.

PR #618, opened 2025-01-02
refactor: store less information per streaming client

thomasheartman:
This PR refactors the broadcaster and what it stores per client:
- I've created a new StreamingQuery struct that has only the data we need,
- I've moved the token storing from client group level to individual client level (and created a ClientData struct)

I realized when setting up the tests for this that the Query, EdgeToken, and FilterQuery all contain more or less the same bits of data. But all we really need is:
1. The actual token string (so that we can boot clients if the token expires)
2. Name prefix, projects, and env.

In the Unleash Types `Query` type, projects and env are optional, but we need them to be present to perform the calculation.

So I created a `StreamingQuery` struct, which consolidates the data we need from the Query and EdgeToken. I also copied in the methods we use for this elsewhere in Unleash and slightly adapted them. I've added notes inline.

### DIFF
--- a/server/src/http/broadcaster.rs
+++ b/server/src/http/broadcaster.rs
@@ -61,7 +61,6 @@ struct ClientData {
 #[derive(Clone, Debug)]
 struct ClientGroup {
     clients: Vec<ClientData>,
-    // last_hash: u64
 }
 
 pub struct Broadcaster {
@@ -221,11 +220,7 @@ impl Broadcaster {
     }
 
     /// Broadcast new features to all clients.
-    pub async fn broadcast(
-        &self,
-        environment: Option<String>,
-        // connections_to_update: &DashMap<QueryWrapper, ClientGroup>
-    ) {
+    pub async fn broadcast(&self, environment: Option<String>) {
         let mut client_events = Vec::new();
 
         for entry in self.active_connections.iter().filter(|entry| {
@@ -376,7 +371,7 @@ mod test {
             loop {
                 if let Some(event) = rx.recv().await {
                     match event {
-                        Event::Data(data) => {
+                        Event::Data(_) => {
                             panic!("Received an update for an env I'm not subscribed to!");
                         }
                         _ => {
--- a/server/src/http/broadcaster.rs
+++ b/server/src/http/broadcaster.rs
@@ -15,7 +15,7 @@ use unleash_types::client_features::{ClientFeatures, Query};
 
 use crate::{
     error::EdgeError,
-    feature_cache::{FeatureCache, UpdateType},
+    feature_cache::FeatureCache,
     filters::{filter_client_features, name_prefix_filter, FeatureFilter, FeatureFilterSet},
     types::{EdgeJsonResult, EdgeResult, EdgeToken},
 };
@@ -108,14 +108,7 @@ impl Broadcaster {
         tokio::spawn(async move {
             while let Ok(key) = rx.recv().await {
                 debug!("Received update for key: {:?}", key);
-                match key {
-                    UpdateType::Full(env) | UpdateType::Update(env) => {
-                        this.broadcast(Some(env)).await;
-                    }
-                    UpdateType::Deletion => {
-                        this.broadcast(None).await;
-                    }
-                }
+                this.broadcast().await;
             }
         });
     }
@@ -217,16 +210,10 @@ impl Broadcaster {
     }
 
     /// Broadcast new features to all clients.
-    pub async fn broadcast(&self, environment: Option<String>) {
+    pub async fn broadcast(&self) {
         let mut client_events = Vec::new();
 
-        for entry in self.active_connections.iter().filter(|entry| {
-            if let Some(env) = &environment {
-                entry.key().environment == *env
-            } else {
-                true
-            }
-        }) {
+        for entry in self.active_connections.iter() {
             let (query, group) = entry.pair();
 
             let event_data = self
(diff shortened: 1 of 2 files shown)


server/src/http/broadcaster.rs
```diff
@@ -1,54 +1,70 @@
...
+    fn from(value: StreamingQuery) -> Self {
+        Self {
+            tags: None,
+            name_prefix: value.name_prefix,
+            environment: Some(value.environment),
+            inline_segment_constraints: Some(false),
+            projects: Some(value.projects),
+        }
+    }
+}
```
thomasheartman, line 40: We don't use `inline_segment_constraints` and `tags` in the actual feature resolution. 

I've set `inline_segment_constraints` to `false` because that indicates that the SDK supports resolution and if the SDK supports streaming, I think we can assume that it supports segment resolution.

As for tags: Edge doesn't support tag queries.

server/src/http/broadcaster.rs
```diff
@@ -101,88 +124,88 @@ impl Broadcaster {
...
-            for client in &group.clients {
-                if client
+            for ClientData { token, sender } in clients {
+                if sender
                     .send(sse::Event::Comment("keep-alive".into()))
                     .await
                     .is_ok()
                 {
-                    ok_clients.push(client.clone());
+                    ok_clients.push(ClientData { token, sender });
```
thomasheartman, line 136: These are memory shenanigans that we probably don't need, but might be nice?

I initially wanted to just avoid cloning each client (especially now that we also have a string in there). To do that, we'd have to take ownership, which [`std::mem::take`](https://doc.rust-lang.org/std/mem/fn.take.html) lets you do. 

This way, we take ownership of the clients and swap it in with an empty vec. We then mutably borrow that new vec and insert only the ok clients. 

Love to have a discussion about this, but it appears to be working.

server/src/http/broadcaster.rs
```diff
@@ -101,88 +124,88 @@ impl Broadcaster {
...
-        let (tx, rx) = mpsc::channel(10);
+        self.create_connection(StreamingQuery::from((&query, &token)), &token.token)
+            .await
+            .map(Sse::from_infallible_receiver)
+    }
 
-        let features = &self
-            .resolve_features(&token, filter_set.clone(), query.clone())
-            .await?;
+    async fn create_connection(
```
thomasheartman, line 155: I split `connect` into `connect` and `create_connection` to facilitate easier testing. Instead of passing in an EdgeToken and a Query (which are almost the same thing), I wanted to just pass the bare minimum that we needed.

server/src/http/broadcaster.rs
```diff
@@ -101,88 +124,88 @@ impl Broadcaster {
...
-        let filter_set = if let Some(name_prefix) = query_filters.name_prefix {
-            FeatureFilterSet::from(Box::new(name_prefix_filter(name_prefix)))
+    fn get_query_filters(query: &StreamingQuery) -> FeatureFilterSet {
+        let filter_set = if let Some(name_prefix) = &query.name_prefix {
+            FeatureFilterSet::from(Box::new(name_prefix_filter(name_prefix.clone())))
         } else {
             FeatureFilterSet::default()
         }
-        .with_filter(project_filter(token));
+        .with_filter(project_filter(query.projects.clone()));
```
thomasheartman, line 194: This looks like a new clone, but it's actually just replacing an old clone. The `project_filter` function in `server/src/filters.rs` looks like this:

```
pub(crate) fn project_filter(token: &EdgeToken) -> FeatureFilter {
    let token = token.clone();
    Box::new(move |feature| {
        if let Some(feature_project) = &feature.project {
            token.projects.is_empty()
                || token.projects.contains(&"*".to_string())
                || token.projects.contains(feature_project)
        } else {
            false
        }
    })
}
```

and clones the whole token. Instead, we'll just clone the projects list here. 

I would like to suggest moving the `project_filter` function in this file back to `filters`, renaming it, and having the original `project_filter` function call this new one instead.

server/src/http/broadcaster.rs
```diff
@@ -194,13 +217,20 @@ impl Broadcaster {
...
         let mut client_events = Vec::new();
-        for entry in self.active_connections.iter() {
+
+        for entry in self.active_connections.iter().filter(|entry| {
+            if let Some(env) = &environment {
+                entry.key().environment == *env
+            } else {
+                true
+            }
+        }) {
```
thomasheartman, line 229: This is the second part of the actual filtering logic, where we filter out active connections based on the env.

server/src/http/broadcaster.rs
```diff
@@ -221,8 +251,134 @@ impl Broadcaster {
...
+    Box::new(move |feature| {
+        if let Some(feature_project) = &feature.project {
+            projects.is_empty()
+                || projects.contains(&"*".to_string())
+                || projects.contains(feature_project)
+        } else {
+            false
+        }
+    })
+}
```
thomasheartman, line 270: This is the more general version of the `project_filter` function from `server/src/filters.rs`.

I think we can move this into the original file and either rename it and the original `project_filter` call it, or use this more generic version.

server/src/http/broadcaster.rs
```diff
@@ -221,8 +251,134 @@ impl Broadcaster {
...
+
+#[cfg(test)]
+mod test {
```
thomasheartman, line 273: I put this test in here because it tests the broadcaster, but I couldn't find a way to do it without the 1s timeout later. Waiting in unit tests isn't great, but it feels like it might be a fine tradeoff.

server/src/http/broadcaster.rs
chriswk: Makes sense to me.

chriswk (approved):
Nice, this seems simpler.

server/src/http/broadcaster.rs
```diff
@@ -1,54 +1,70 @@
...
+impl From<(&Query, &EdgeToken)> for StreamingQuery {
+    fn from((query, token): (&Query, &EdgeToken)) -> Self {
+        Self {
+            projects: token.projects.clone(),
+            name_prefix: query.name_prefix.clone(),
+            environment: match token.environment {
+                Some(ref env) => env.clone(),
+                None => token.token.clone(),
+            },
+        }
```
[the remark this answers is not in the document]
sighphyre: It's fine in my opinion

server/src/http/broadcaster.rs
sighphyre: I don't think memory shenanigans are worth it here. I'm late to this party but unless this runs every time a connected client tries to get features or the client struct holds a non Arc'd struct that's a direct chunk of memory for all the features then simpler is better.

Profile first either way

server/src/http/broadcaster.rs
sighphyre: If it makes sense and/or makes the code more readable then go for it. I wouldn't worry too much about the clone here, remember that this will get dwarfed by the memory copy of the client features JSON to the outgoing socket anyway

server/src/http/broadcaster.rs
thomasheartman: Yeah, that's fair. This runs every 30 seconds for every client. Again, probably not gonna make a massive difference, so I'd be happy to revert it to how it was 🤷🏼  I don't necessarily want to go through the rigmarole of profiling for this right away, though I'd be happy to have a look at it with you, if you want to have a look at the heaptrack output. I ... don't know what that is, so an intro might be in order 😅

server/src/http/broadcaster.rs
thomasheartman: Nah, I'm not worried about th ecloning here. That comment was just to say that I'm not replacing a previous non-clone with a new clone. The cloning happens anyway. This just clones less and more explicitly.

server/src/http/broadcaster.rs
sighphyre: Yuh, we can take a look together, Heaptrack is this beast: https://github.com/KDE/heaptrack, give you insight into the heap allocations. It's not urgent and a quick and dirty "glance at the process memory in task manager is probably good enough for now but we should know how much memory this is going to eat before we release it

thomasheartman: Ah, neat. Yeah, if that's a Linux-only tool, then I'm a little out of luck. Looks like it may be possible to compile it for macOS, but yeah, let's not start there. I'd love to have a little sesh on it if you're down, though. Just hit me up whenever.

PR #619, opened 2025-01-06
feat(1-3220): only push updates to listeners in the same environment

thomasheartman:
This PR updates the broadcaster to only push updates to listeners in the environment it got updates for. In other words, if you get an update for "development" config, you shouldn't push an update to listeners subscribed to the "production" environment.

This is the first step in more selective pushing (the next would be to do as we do in Unleash and hash the result of each query so that we only push out updates when something's actually changed).


thomasheartman:
Closed in favor of #620

PR #620, opened 2025-01-06
feat(1-3220): only push updates to listeners in the same environment

thomasheartman:
This PR updates the broadcaster to only push updates to listeners in the environment it got updates for. In other words, if you get an update for "development" config, you shouldn't push an update to listeners subscribed to the "production" environment.

This is the first step in more selective pushing (the next would be to do as we do in Unleash and hash the result of each query so that we only push out updates when something's actually changed).

### DIFF
--- a/server/src/http/broadcaster.rs
+++ b/server/src/http/broadcaster.rs
@@ -23,7 +23,7 @@ use unleash_types::client_features::{ClientFeatures, Query as FlagQuery};
 
 use crate::{
     error::EdgeError,
-    feature_cache::FeatureCache,
+    feature_cache::{FeatureCache, UpdateType},
     filters::{filter_client_features, name_prefix_filter, project_filter, FeatureFilterSet},
     tokens::cache_key,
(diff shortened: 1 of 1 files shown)


server/src/http/broadcaster.rs
```diff
@@ -194,9 +201,16 @@ impl Broadcaster {
...
 
     /// Broadcast new features to all clients.
-    pub async fn broadcast(&self) {
+    pub async fn broadcast(&self, environment: Option<String>) {
         let mut client_events = Vec::new();
-        for entry in self.active_connections.iter() {
+
+        for entry in self.active_connections.iter().filter(|entry| {
+            if let Some(env) = &environment {
+                entry.key().environment == *env
```
thomasheartman, line 209: This line fails for now because it expects the refactor in #618 to go through first. I'd like to leave this as is for now, because the extra logic required in the old setup is a little convoluted. However, it means build/tests will fail until the updated is in.

chriswk (approved):
LGTM

PR #621, opened 2025-01-06
refactor: minor cleanup after #618

thomasheartman:
This PR performs a small amount of cleanup after #618 based on the PR discussion. Specifically, it:

- Adds a new `project_filter_from_projects` function to the filters file, and makes `project_filter` fall back to that on the back end (we can discuss the names here, but doing it this way saved me touching any more code)
- Reverts the `std::mem::take` shenanigans in favor of cloning. Even though it clones one string per attached listener every 30 seconds, it's probably not going to cause a lot of memory overhead given that we drop the original shortly thereafter.


chriswk (approved):
Let's go :)

PR #622, opened 2025-01-07
fix: add meta to support new format

sjaanus:
Trying to get tip of main working, but some tests are failing.
I suspect it is related to the new meta field, I added it to all obvious places, but I think its more complex.

### DIFF
--- a/server/src/feature_cache.rs
+++ b/server/src/feature_cache.rs
@@ -99,7 +99,7 @@ fn update_client_features(
             s
         }),
         query: old.query.clone().or(update.query.clone()),
-        meta: None,
+        meta: old.meta.clone().or(update.meta.clone()),
     }
 }
 
--- a/server/src/feature_cache.rs
+++ b/server/src/feature_cache.rs
@@ -99,6 +99,7 @@ fn update_client_features(
             s
         }),
         query: old.query.clone().or(update.query.clone()),
+        meta: None,
     }
 }
 
--- a/server/src/filters.rs
+++ b/server/src/filters.rs
@@ -48,6 +48,7 @@ pub(crate) fn filter_client_features(
         segments: feature_cache.segments.clone(),
         query: feature_cache.query.clone(),
         version: feature_cache.version,
+        meta: None,
     }
 }
 
@@ -90,6 +91,7 @@ mod tests {
             }],
             query: None,
             segments: None,
+            meta: None,
         };
 
         let map: DashMap<String, ClientFeatures> = DashMap::default();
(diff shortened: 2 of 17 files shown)


sighphyre (approved):
Seems okay, I think this should work after Yggdrasil gets an update to Unleash Types

chriswk (approved):

sjaanus:
Got the build working, but I suspect it should not be None always, but it should pass this value forward from Unleash.

server/src/feature_cache.rs
```diff
@@ -99,6 +99,7 @@ fn update_client_features(
             s
         }),
         query: old.query.clone().or(update.query.clone()),
+        meta: None,
```
sjaanus, line 102: Should this be old.meta.clone()

server/src/filters.rs
```diff
@@ -48,6 +48,7 @@ pub(crate) fn filter_client_features(
         segments: feature_cache.segments.clone(),
         query: feature_cache.query.clone(),
         version: feature_cache.version,
+        meta: None,
```
sjaanus, line 51: Should this be feature_cache.meta.

server/src/types.rs
```diff
@@ -90,6 +90,7 @@ pub enum TokenType {
 }
 
 #[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
```
sjaanus, line 93: Got error

```
error: large size difference between variants
  --> server/src/types.rs:93:1
   |
93 | / pub enum ClientFeaturesResponse {
94 | |     NoUpdate(EntityTag),
   | |     ------------------- the second-largest variant contains at least 32 bytes
95 | |     Updated(ClientFeatures, Option<EntityTag>),
   | |     ------------------------------------------ the largest variant contains at least 256 bytes
96 | | }
   | |_^ the entire enum is at least 256 bytes

server/src/feature_cache.rs
chriswk: Hmm, probably `update.meta.clone().or(old.meta.clone())`

server/src/filters.rs
chriswk: yes

server/src/feature_cache.rs
sjaanus: Added

server/src/filters.rs
sjaanus: Added

chriswk (approved):
Giving you a :+1: again. So far, I don't think we use meta too much. Mateusz just confirmed that the node sdk also ignores it for now.

PR #626, opened 2025-01-08
feat: delta api implementation

sjaanus:
Now edge can understand and talk to Unleash Delta API. There is cli flag **delta** to enable this feature.


server/tests/delta_test.rs
```diff
@@ -0,0 +1,159 @@
...
+    use unleash_edge::http::unleash_client::{ClientMetaInformation, UnleashClient};
+    use unleash_edge::types::EdgeToken;
+    use unleash_types::client_features::{
+        ClientFeature, ClientFeatures, ClientFeaturesDelta, Constraint, Operator, Segment,
+    };
+    use unleash_yggdrasil::EngineState;
+
+    #[actix_web::test]
+    #[tracing_test::traced_test]
+    async fn test_delta() {
```
sjaanus, line 23: Testing whether, after the eTag is provided by Edge, Edge can successfully apply the delta when Unleash sends a new delta.

server/src/http/feature_refresher.rs
```diff
@@ -1590,4 +1702,12 @@ mod tests {
...
         assert_eq!(updated.len(), 0);
     }
+
+    // mod delta_tests {
+    //     use super::*;
+    //     #[test]
+    //     fn delta_works() {
+    //
+    //     }
+    // }
```
nunogois, line 1712: 👀

server/src/cli.rs
```diff
@@ -217,6 +217,10 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict")]
     pub streaming: bool,
 
+    /// If set to true. Edge connects to upstream using delta polling instead of normal polling.
```
sighphyre, line 220:
```suggestion
    /// If set to true, Edge connects to upstream using delta polling instead of normal polling.
```

sighphyre: Can I suggest putting something in this text to tell users that this is an experimental API and might change?

server/src/feature_cache.rs
```diff
@@ -67,6 +67,23 @@ impl FeatureCache {
         self.send_full_update(key);
     }
 
+    pub fn apply_delta(&self, key: String, delta: &ClientFeaturesDelta) {
+        let client_features = ClientFeatures {
+            version : 1,
```
sighphyre, line 72:
```suggestion
            version : 2,
```

Honestly we've been on version 2 for so long I don't even know what version 1 does

server/src/http/feature_refresher.rs
```diff
@@ -449,6 +465,36 @@ impl FeatureRefresher {
...
+        delta: ClientFeaturesDelta,
+        etag: Option<EntityTag>,
+    ) {
+        debug!("Got updated client features delta. Updating features with {etag:?}");
+        let key = cache_key(refresh_token);
+        self.features_cache.apply_delta(key.clone(), &delta);
+        self.update_last_refresh(
+            refresh_token,
+            etag,
+            self.features_cache.get(&key).unwrap().features.len(),
```
sighphyre, line 480: Is this unwrap safe?

server/tests/delta_test.rs
```diff
@@ -0,0 +1,159 @@
+mod delta_test {
```
sighphyre, line 1: This is a weird place to put tests in my opinion. Are these very long running or a pain to setup for some reason? Can they not just put alongside the code that they're testing?

sighphyre (commented):
This is looking cool!

Can I please suggest we put this behind a compiler flag while we test this? As soon as we release publicly, this becomes a huge pain to rollback or change things if we need

server/src/http/feature_refresher.rs
sjaanus: Should be, the line 476 should guarantee that cache has value with that key.

server/tests/delta_test.rs
sjaanus: Feature refresher is already so big, 1700 lines. It is super hard to follow as a newcomer.

server/src/cli.rs
sjaanus: Added

server/tests/delta_test.rs
sighphyre: I think that's a signal that feature refresher is too big and new code needs to be split out, rather than moving tests out

chriswk: Probably add a new delta_refresher.rs, which can still extend FeatureRefresher with methods by using
```rust
impl FeatureRefresher {
  /// The delta code
}

#[cfg(test)]
mod tests {
 /// our delta tests, that way we won't have to compile a separate module either.
}
```

That also means we can readd the #[cfg(test)] to the new method for unleash client.

chriswk: Doing it that way, also allows doing what Simon suggests and add a `delta` compiler flag
basically, in the http module, you'd declare
```rust
#[cfg(feature = "delta")]
mod delta_refresher;
```

chriswk: But also a good observation on the size (I'm assuming the 1700 lines includes the test part of the code as well). 

It might point towards feature_refresher needing to be a folder with multiple files rather than one file.

server/src/feature_cache.rs
FredrikOseberg: Should this eventually come from Unleash itself? If we ever change this version we'll get out of sync.

server/tests/delta_test.rs
sjaanus: Added it behind compiler flag.

server/src/cli.rs
```diff
@@ -217,6 +217,10 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict")]
     pub streaming: bool,
 
+    /// If set to true. Edge connects to upstream using delta polling instead of normal polling. This is experimental feature and might and change.
```
sighphyre, line 220:
```suggestion
    /// If set to true. Edge connects to upstream using delta polling instead of normal polling. This is experimental feature and might change.
```

sighphyre (approved):
Very nice. LGTM

server/src/http/refresher/delta_refresher.rs
```diff
@@ -0,0 +1,271 @@
...
+                    }
+                    _ => info!("Couldn't refresh features: {e:?}. Will retry next pass"),
+                }
+            }
+        }
+    }
+}
+
+
+#[cfg(feature = "delta")]
```
chriswk, line 111: Since you're always adding the impl part, I'm not sure if there's much value in using a feature flag just for tests.

server/src/http/refresher/delta_refresher.rs
```diff
@@ -0,0 +1,271 @@
...
+use tracing::{debug, info, warn};
+use unleash_types::client_features::{ClientFeaturesDelta};
+use unleash_yggdrasil::EngineState;
+
+use crate::error::{EdgeError, FeatureError};
+use crate::types::{ClientFeaturesDeltaResponse, ClientFeaturesRequest, EdgeToken, TokenRefresh};
+use crate::http::refresher::feature_refresher::FeatureRefresher;
+use crate::tokens::cache_key;
+
+impl FeatureRefresher {
```
chriswk, line 12: Either add #[cfg(feature = "delta")] here, or remove it from the tests part.

server/src/http/unleash_client.rs
```diff
@@ -210,7 +218,6 @@ impl UnleashClient {
         }
     }
 
-    #[cfg(test)]
```
chriswk, line 213: Readding this one seems useful.

PR #640, opened 2025-01-09
chore(ci): Consolidate on building with chef

chriswk:
We have way too many workflows flowing around. This one updates to use docker bake for defining multiplatform images, and uses chef to be able to cache as many layers of the docker image as possible.


sighphyre (approved):
LGTM

PR #651, opened 2025-01-10
chore(ci): Remove semver task

chriswk:
With release-plz if we want help, we can just turn on semver_check in release-plz.toml instead

This will stop having our main branch look like the build is always failing


sighphyre (approved):
Nice

PR #654, opened 2025-01-10
chore: add debug message for delta

sjaanus:


server/src/http/refresher/delta_refresher.rs
```diff
@@ -16,7 +16,13 @@ impl FeatureRefresher {
         delta: ClientFeaturesDelta,
         etag: Option<EntityTag>,
     ) {
-        debug!("Got updated client features delta. Updating features with {etag:?}");
+        let updated_len = delta.updated.len();
+        let removed_len = delta.removed.len();
+
+        debug!(
+            "Got updated client features delta. Updating features with {etag:?}, updated {updated_len:?}, removed {removed_len:?}"
+        );
```
chriswk, line 24:
```suggestion
        debug!(
            "Got updated client features delta. Updating features with {etag:?}, updated {updated_len}, removed {removed_len}"
        );
```
These are numbers you shouldn't need to use the debug output, they implement Display already.

chriswk (approved):
LGTM

PR #659, opened 2025-01-13
feat: make delta aligned with streaming, support only strict

sjaanus:
Align delta with streaming, only to work in strict mode and not conflicting.

We are not supporting legacy dynamic mode.


FredrikOseberg (approved):
LG

PR #663, opened 2025-01-14
chore(1-3244): only expose streaming endpoint if in streaming mode

thomasheartman:
Make it so that the /streaming endpoint only accepts connections if the server is in streaming mode. This effectively renders the /streaming endpoint useless if the server is not connected to an upstream stream.

This first solution uses the 403 http status code and tells the user that the endpoint is only enabled in streaming mode.

## Discussion

Is there a way to not mount this if we're not streaming? Is introducing a new edge error the right thing to do here, or is there an easier way to return arbitrary status codes?

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -1,4 +1,4 @@
-use crate::cli::EdgeMode;
+use crate::cli::{EdgeArgs, EdgeMode};
 use crate::error::EdgeError;
 use crate::feature_cache::FeatureCache;
 use crate::filters::{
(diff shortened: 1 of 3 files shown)


sighphyre (approved):
This seems okay to me. You're gonna have to fix the test though

sighphyre:
> Is there a way to not mount this if we're not streaming? 

There are a few endpoints that we conditionally mount, you can inspect the setup to see how that's done. That being said, I think what you've done is better. Its a nicer user experience to get an error message telling you what went wrong, rather than a NotFound

> Is introducing a new edge error the right thing to do here, or is there an easier way to return arbitrary status codes?

That seems fine to me, it signals the intent well enough and you need to wrap it in an error of some kind

PR #666, opened 2025-01-15
feat: start logging diff comparing with delta

sjaanus:
Now when client/features is called, it will also call delta API and verify that payload matches.

This is behind feature flag.


server/src/http/refresher/delta_refresher.rs
```diff
@@ -47,22 +47,22 @@ impl FeatureRefresher {
     }
 
     pub async fn refresh_single_delta(&self, refresh: TokenRefresh) {
-        let features_result = self
+        let delta_result = self
             .unleash_client
             .get_client_features_delta(ClientFeaturesRequest {
                 api_key: refresh.token.token.clone(),
-                etag: refresh.etag,
+                etag: refresh.etag.clone(),
```
sighphyre, line 54: Why the clone here?

sighphyre (approved):
Sure

server/src/cli.rs
```diff
@@ -221,6 +221,10 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict")]
     pub delta: bool,
 
+    /// If set to true, it compares features payload with delta payload and logs diff. This is experimental feature and might and change. Requires strict mode
```
chriswk, line 224:
```suggestion
    /// If set to true, it compares features payload with delta payload and logs diff. This is experimental feature and might change. Requires strict mode
```

server/src/http/refresher/feature_refresher.rs
```diff
@@ -470,7 +509,7 @@ impl FeatureRefresher {
             .unleash_client
             .get_client_features(ClientFeaturesRequest {
                 api_key: refresh.token.token.clone(),
-                etag: refresh.etag,
+                etag: refresh.etag.clone(),
```
chriswk, line 512: why did it suddenly need a clone?

chriswk (approved):
Some nits, but I'm :+1: here. Good job

server/src/http/refresher/feature_refresher.rs
sjaanus: It was due to 

```
 --> server/src/http/refresher/feature_refresher.rs:526:50
    |
512 |                 etag: refresh.etag,
    |                       ------------ value partially moved here
...
526 |                         self.compare_delta_cache(&refresh).await;
    |                                                  ^^^^^^^^ value borrowed here after partial move

```

PR #674, opened 2025-01-16
chore: remove compiler flag for delta

sjaanus:
Remove delta compiler flag, so we do not need to compile it manually. Feature flag is enough.

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -221,8 +221,8 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict")]
     pub delta: bool,
 
-    /// If set to true, it compares features payload with delta payload and logs diff. This is experimental feature and might change. Requires strict mode
-    #[clap(long, env, default_value_t = false, requires = "strict")]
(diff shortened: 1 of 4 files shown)


server/src/cli.rs
```diff
@@ -221,8 +221,8 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict")]
     pub delta: bool,
 
-    /// If set to true, it compares features payload with delta payload and logs diff. This is experimental feature and might change. Requires strict mode
-    #[clap(long, env, default_value_t = false, requires = "strict")]
+    /// If set to true, it compares features payload with delta payload and logs diff. This is experimental feature and might change.
+    #[clap(long, env, default_value_t = false, conflicts_with = "delta")]
     pub delta_diff: bool,
```
chriswk, line 226: so you can't have both delta_diff and delta, might be worth it to consider looking at `arg_groups`, but for now, this will work fine. :+1:

chriswk (approved):

server/src/cli.rs
sighphyre: delta_diff probably needs to come out before release anyway, that's only useful for debug

PR #698, opened 2025-01-30
feat: update edge to new delta format

sjaanus:
Make edge support the new delta format.

### DIFF
--- a/server/src/feature_cache.rs
+++ b/server/src/feature_cache.rs
@@ -68,19 +68,12 @@ impl FeatureCache {
     }
 
     pub fn apply_delta(&self, key: String, delta: &ClientFeaturesDelta) {
-        let client_features = ClientFeatures {
-            version : 2,
-            features : vec![],
-            segments: Some(vec![]),
-            query: None,
-            meta: None,
-        };
         self.features
             .entry(key.clone())
             .and_modify(|existing_features| {
-                existing_features.modify_in_place(delta);
+                existing_features.apply_delta(delta);
             })
-            .or_insert(client_features.modify_and_copy(delta));
+            .or_insert(ClientFeatures::create_from_delta(delta));
         self.send_full_update(key);
     }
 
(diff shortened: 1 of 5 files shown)


server/src/feature_cache.rs
```diff
@@ -80,7 +80,7 @@ impl FeatureCache {
             .and_modify(|existing_features| {
                 existing_features.modify_in_place(delta);
             })
-            .or_insert(client_features);
+            .or_insert(client_features.modify_and_copy(delta));
```
sjaanus, line 83: We can not put the values directly in, but need to run it through types method that parses events.

server/src/http/refresher/feature_refresher.rs
```diff
@@ -403,15 +403,24 @@ impl FeatureRefresher {
             .unleash_client
             .get_client_features_delta(ClientFeaturesRequest {
                 api_key: refresh.token.token.clone(),
-                etag: refresh.etag.clone(),
+                etag: None,
```
sjaanus, line 406: No point to pass in this etag, because etags are different for old api and delta api.

server/src/feature_cache.rs
sighphyre: Seems a bit unfortunate that we have to construct a new `ClientFeatures` just so modify_and_copy can clone it.

Running through the types code, it looks like `modify_in_place` and `modify_and_copy` are sharing the bulk of their code. Ideally the bulk of that method would be extracted to something that they could both share and a third method could be created that builds a `ClientFeatures` from a delta

sjaanus: Good point, I thought about that also at one point, but did not want to make it complex. I will take a look now.

sjaanus: Regarding the construction of new `ClientFeatures`, we need to construct it anyways at one point if the value is not in cache.

server/src/http/refresher/feature_refresher.rs
FredrikOseberg: I'm curious, don't we still need an etag here?

FredrikOseberg (approved):
LGTM

PR #705, opened 2025-02-03
feat: add configuration possibilities for CORS middleware

### LINKED ISSUE #701: can we add whitelisted domain for unleash edge instance

Currently, Unleash Edge does not support CORS configuration. We want to restrict access so that only our web apps can fetch the data as a security measure. According to the documentation, if this feature is required we should raise an issue or submit a PR. Therefore, I’ve created this feature request.

https://github.com/Unleash/unleash-edge/blob/main/migration-guide.md#not-supported

### Background

_No response_

### Solution suggestions

_No response_

chriswk:
Reported in #701 - Edge currently is set to Allow-Any-Origin  (`*`). the Issue requests the possibility to lock it down more than this.

This PR uses the same configuration variables available in Unleash Proxy: https://docs.getunleash.io/reference/unleash-proxy#configuration-options but excludes the ones that are not supported by our actix-cors middleware.

fixes #701

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -437,7 +437,6 @@ impl CorsOptions {
             .max_age(self.cors_max_age)
             .allow_any_method()
             .allow_any_header()
-            .expose_any_header();
         if let Some(origins) = self.cors_origin.clone() {
             for origin in origins {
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
LGTM

PR #713, opened 2025-02-05
feat: Add edge observability

chriswk:


server/src/metrics/actix_web_metrics.rs
```diff
@@ -121,6 +121,10 @@ impl Metrics {
             .f64_histogram(HTTP_SERVER_DURATION)
             .with_description("HTTP inbound request duration per route")
             .with_unit("ms")
+            .with_boundaries(vec![
+                1.0, 5.0, 10.0, 20.0, 30.0, 40.0, 50.0, 100.0, 200.0, 300.0, 400.0, 500.0, 750.0,
+                1000.0, 1500.0, 2000.0,
+            ])
```
chriswk, line 127: this one used 0.0 as its first bucket, which since we're operating with time, we don't have negative numbers.

server/src/http/instance_data.rs
```diff
@@ -0,0 +1,122 @@
+use chrono::Duration;
```
sighphyre, line 1: Moving this to a new file in this PR makes this really challenging to review. I'm guessing some stuff has changed here?

server/src/http/feature_refresher.rs
```diff
@@ -0,0 +1,1593 @@
+use std::collections::HashSet;
```
sighphyre, line 1: Moving this to a new file in this PR makes this really challenging to review. I'm guessing some stuff has changed here?

server/src/http/instance_data.rs
```diff
@@ -0,0 +1,122 @@
...
+                &prometheus_registry,
+                downstream_instance_data.read().await.clone(),
+                &instance_data_sender.base_path,
+            );
+            if do_the_work {
+                let status = instance_data_sender
+                    .unleash_client
+                    .send_instance_data(observed_data, &instance_data_sender.token)
+                    .await;
+                match status {
```
sighphyre, line 90: Personal taste but I think `if let Err` is a nice construct than an empty Ok match. You could replace this with

```
                if let Err(e) = status {
                    match e {
                        EdgeError::EdgeMetricsRequestError(status, message) => {
                            warn!("Failed to post instance data with status {status} and {message:?}");
                            if status == StatusCode::NOT_FOUND {
                                debug!("Upstream edge metrics not found, clearing our data about downstream instances to avoid growing to infinity (and beyond!).");
                                empty = true;
                                do_the_work = false;
                            } else if status == StatusCode::FORBIDDEN {
                                warn!("Upstream edge metrics rejected our data, clearing our data about downstream instances to avoid growing to infinity (and beyond!)");
                                empty = true;
                                do_the_work = false;
                            }
                        }
                        _ => {
                            warn!("Failed to post instance data due to unknown error {e:?}");
                            empty = false;
                        }
                    }
                }
```

server/src/client_api.rs
```diff
@@ -272,6 +275,26 @@ pub async fn post_bulk_metrics(
...
+)
+)]
+#[post("/metrics/edge")]
+#[instrument(skip(_edge_token, instance_data, connected_instances))]
+pub async fn post_edge_instance_data(
+    _edge_token: EdgeToken,
+    instance_data: Json<EdgeInstanceData>,
+    connected_instances: Data<RwLock<Vec<EdgeInstanceData>>>,
+) -> EdgeResult<HttpResponse> {
+    info!("Accepted {instance_data:?}");
```
sighphyre, line 290: Could we check if we have an InstanceDataSender here and drop this if we don't have one? It'd reduce memory and let us unnest some code later on

server/src/client_api.rs
```diff
@@ -272,6 +275,26 @@ pub async fn post_bulk_metrics(
...
+    connected_instances
```
sighphyre, line 291: Can we check if we have an instance sender here and not store this data if we don't? I'd help with memory and flattening some of the code later

server/src/http/instance_data.rs
```diff
@@ -0,0 +1,122 @@
...
+    instance_data_sender: Option<InstanceDataSender>,
+    prometheus_registry: Registry,
+    our_instance_data: Arc<EdgeInstanceData>,
+    downstream_instance_data: Arc<RwLock<Vec<EdgeInstanceData>>>,
+) {
+    let mut do_the_work = true;
+    loop {
+        let mut empty = true;
+        tokio::time::sleep(std::time::Duration::from_secs(60)).await;
+        if let Some(instance_data_sender) = instance_data_sender.clone() {
```
sighphyre, line 78: So this function is pretty nested. Again personal opinion, but if we never stored the data in if we couldn't send it, then we could use 

```

        let Some(instance_data_sender) = instance_data_sender.clone() else {
            continue;
        };
```

to flatten this out a little

server/src/http/instance_data.rs
```diff
@@ -0,0 +1,122 @@
...
+                        base_path: args.http.base_path.clone(),
+                    }
+                }))
+            }
+            _ => Ok(None),
+        }
+    }
+}
+
+pub async fn send_instance_data(
```
sighphyre, line 68: Can we give this a different name? We now have two functions called the same thing doing different things

server/src/http/unleash_client.rs
```diff
@@ -541,6 +561,48 @@ impl UnleashClient {
...
+            .observe(
+                ended_at
+                    .signed_duration_since(started_at)
+                    .num_milliseconds() as f64,
+            );
+        let r = if result.status().is_success() {
+            Ok(())
+        } else {
+            match result.status() {
+                StatusCode::BAD_REQUEST => Err(EdgeMetricsRequestError(
```
sighphyre, line 595: Why this specific check?

server/src/metrics/edge_metrics.rs
```diff
@@ -0,0 +1,329 @@
...
+
+#[derive(Debug, Default, Clone, Copy, Deserialize, Serialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct LatencyMetrics {
+    pub avg: f64,
+    pub count: f64,
+    pub p99: f64,
+}
+
+impl LatencyMetrics {
```
sighphyre, line 18: Default is already implemented for this struct, this shouldn't be needed

server/src/metrics/edge_metrics.rs
```diff
@@ -0,0 +1,329 @@
...
+        Self {
+            identifier: Ulid::new().to_string(),
+            app_name: app_name.to_string(),
+            region: std::env::var("AWS_REGION").ok(),
+            edge_version: build_info.package_version.clone(),
+            process_metrics: None,
+            started: Utc::now(),
+            traffic: InstanceTraffic::default(),
+            latency_upstream: UpstreamLatency::default(),
+            connected_edges: Vec::new(),
```
sighphyre, line 80: Not married to it but I'm so used to seeing the macro invocation over explicit new here

```suggestion
            connected_edges: vec![],
```

server/src/metrics/edge_metrics.rs
chriswk: Agreed.

server/src/metrics/edge_metrics.rs
chriswk: yeah, fixed

server/src/http/unleash_client.rs
chriswk: Because copy-pasta from earlier request. I still like it though, because it exposes full message if we receive 400 from upstream, so we know what we need to fix.

server/src/http/instance_data.rs
chriswk: yes

server/src/http/feature_refresher.rs
chriswk: TBH, I wasn't aware I did.. where was it located earlier?

chriswk: Turns out dx had moved it to the refresher module, and my branch hadn't caught that. Removed my copy and kept the one inside the refresher module.

nunogois (approved):
LGTM!

PR #723, opened 2025-02-10
feat: upgrade logging level of serialization errors to actual errors

chriswk:
We had an incident over the weekend where our sandbox instances weren't being updated due to the features response not being parsed, but since it was logged at warn we did not get any notices about this in Slack. This PR upgrades the logging level of parse errors for SSE to error so that our error alerting can trigger and we can know that our upstream is returning invalid data.


server/src/http/refresher/feature_refresher.rs
```diff
@@ -347,7 +349,7 @@ impl FeatureRefresher {
 
                                     match serde_json::from_str(&event.data) {
                                         Ok(features) => { refresher.handle_client_features_updated(&token, features, None).await; }
-                                        Err(e) => { warn!("Could not parse features response to internal representation: {e:?}");
+                                        Err(e) => { tracing::error!("Could not parse features response to internal representation: {e:?}");
```
gastonfournier, line 352: Ok, this is the relevant change, shouldn't we use the same `match serde_json::from_str` in line 424 and 425 when reading deltas?

gastonfournier (approved):
LG! Just noticed a potential improvement that can be dealt with in another PR

PR #736, opened 2025-02-14
feat: implement delta cache in edge

sjaanus:
Edge needs to start keeping all the deltas in cache. This is a cache where these deltas are kept. Tried to keep it consistant with delta cache inside Unleash https://github.com/Unleash/unleash/blob/4a72580b2477c91883680fd03adc8e31ef85f48f/src/lib/features/client-feature-toggles/delta/delta-cache.ts#L7.

### DIFF
--- a/server/src/delta_cache.rs
+++ b/server/src/delta_cache.rs
@@ -1,4 +1,4 @@
-use unleash_types::client_features::{DeltaEvent, ClientFeature, Segment, Strategy};
+use unleash_types::client_features::{DeltaEvent, ClientFeature, Segment};
 
 
 
--- a/server/src/delta_cache.rs
+++ b/server/src/delta_cache.rs
@@ -5,7 +5,7 @@ use unleash_types::client_features::{DeltaEvent, ClientFeature, Segment};
 
 #[derive(Debug, Clone)]
 pub struct DeltaHydrationEvent {
-    event_id: i32,
+    event_id: u32,
     features: Vec<ClientFeature>,
     segments: Vec<Segment>,
 }
@@ -16,22 +16,6 @@ pub struct DeltaCache {
     max_length: usize,
     events: VecDeque<DeltaEvent>,
     hydration_event: DeltaHydrationEvent,
-}
-
-trait GetEventId {
-    fn get_event_id(&self) -> i32;
-}
-
-impl GetEventId for DeltaEvent {
-    fn get_event_id(&self) -> i32 {
-        match self {
-            DeltaEvent::FeatureUpdated { event_id, .. }
-            | DeltaEvent::FeatureRemoved { event_id, .. }
-            | DeltaEvent::SegmentUpdated { event_id, .. }
-            | DeltaEvent::SegmentRemoved { event_id, .. }
-            | DeltaEvent::Hydration { event_id, .. } => *event_id,
-        }
-    }
 }
 
 impl DeltaCache {
@@ -69,7 +53,7 @@ impl DeltaCache {
     }
 
     pub fn is_missing_revision(&self, revision_id: u32) -> bool {
-        !self.events.iter().any(|event| event.get_event_id() == revision_id as i32)
+        !self.events.iter().any(|event| event.get_event_id() == revision_id)
     }
 
     pub fn get_hydration_event(&self) -> &DeltaHydrationEvent {
--- a/server/src/delta_cache.rs
+++ b/server/src/delta_cache.rs
@@ -1,4 +1,3 @@
-use std::collections::VecDeque;
 use unleash_types::client_features::{DeltaEvent, ClientFeature, Segment};
 
 
@@ -14,7 +13,7 @@ pub struct DeltaHydrationEvent {
 #[derive(Debug, Clone)]
 pub struct DeltaCache {
     max_length: usize,
-    events: VecDeque<DeltaEvent>,
+    events: Vec<DeltaEvent>,
     hydration_event: DeltaHydrationEvent,
 }
 
@@ -22,7 +21,7 @@ impl DeltaCache {
     pub fn new(hydration_event: DeltaHydrationEvent, max_length: usize) -> Self {
         let mut cache = DeltaCache {
             max_length,
-            events: VecDeque::new(),
+            events: Vec::new(),
             hydration_event: hydration_event.clone(),
         };
         cache.add_base_event_from_hydration(hydration_event);
@@ -40,15 +39,17 @@ impl DeltaCache {
 
     pub fn add_events(&mut self, events: Vec<DeltaEvent>) {
         for event in events.into_iter() {
-            self.events.push_back(event.clone());
+            self.events.push(event.clone());
             self.update_hydration_event(event);
+
             if self.events.len() > self.max_length {
-                self.events.pop_front();
+                self.events.remove(0); // O(n) operation
             }
         }
     }
 
-    pub fn get_events(&self) -> &VecDeque<DeltaEvent> {
+
+    pub fn get_events(&self) -> &Vec<DeltaEvent> {
         &self.events
     }
 
--- a/server/src/delta_cache.rs
+++ b/server/src/delta_cache.rs
@@ -105,31 +105,11 @@ fn test_update_hydration_event_and_remove_event_when_over_limit() {
         features: vec![
             ClientFeature {
                 name: "test-flag".to_string(),
-                feature_type: None,
-                description: None,
-                created_at: None,
-                last_seen_at: None,
-                enabled: false,
-                stale: None,
-                impression_data: None,
-                project: None,
-                strategies: None,
-                variants: None,
-                dependencies: None,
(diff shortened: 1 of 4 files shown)


server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+    let hydration_event = delta_cache.get_hydration_event();
+    assert_eq!(hydration_event.features.len(), 2);
+    assert_eq!(hydration_event.event_id, 7);
+    assert!(hydration_event.features.iter().any(|f| f.name == "my-feature-flag"));
+    assert!(hydration_event.features.iter().any(|f| f.name == "another-feature-flag"));
+    assert!(hydration_event.segments.iter().any(|s| s.id == 1));
+}
+
+#[test]
+fn test_prevent_mutation_of_previous_feature_updated_events() {
```
chriswk, line 226: For my understanding, you're testing that adding two FeatureUpdated events updating the same event keeps both of them?

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+            impression_data: None,
+            project: None,
+            strategies: None,
+            variants: None,
+            dependencies: None,
+        },
+    };
+    delta_cache.add_events(vec![updated_feature_event]);
+
+    assert_eq!(delta_cache.get_events()[1], initial_feature_event);
```
chriswk, line 285: We'd expect 3 events in total here?
So `delta_cache.get_events()[2]` would be `updated_feature_event`?

chriswk (approved):
Looks good (and rusty) to me.
Happy to spar next week on making sure the ETag treatment gets updated to be closer to what Unleash does.

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+
+
+#[derive(Debug, Clone)]
+pub struct DeltaCache {
+    max_length: usize,
+    events: VecDeque<DeltaEvent>,
+    hydration_event: DeltaHydrationEvent,
+}
+
+trait GetEventId {
```
sighphyre, line 21: A trait feels like massive overkill here. Can't we put this method on the DeltaEvent impl?

chriswk: You could also consider just making the fields of the structs public, so you can go `struct.field` instead of needing a `get_field()` method. After all, this isn't Java :)

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+            max_length,
+            events: VecDeque::new(),
+            hydration_event: hydration_event.clone(),
+        };
+        cache.add_base_event_from_hydration(hydration_event);
+        cache
+    }
+
+    fn add_base_event_from_hydration(&mut self, hydration_event: DeltaHydrationEvent) {
+        if let Some(last_feature) = hydration_event.features.last().cloned() {
```
sighphyre, line 49: This `if let Some` makes it looks like you always expect there to be a last feature and that that's a system constraint. If that's the case, an `unwrap` or `expect` would be more appropriate here

server/src/delta_cache.rs
kwasniew: we did the trait because the type was in another lib and was and an enum. Each variant of the enum has the same eventId. This was a workaround to do all of the work in one repo.

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    fn update_hydration_event(&mut self, event: DeltaEvent) {
+        match event {
+            DeltaEvent::FeatureUpdated { event_id, feature } => {
+                if let Some(existing) = self.hydration_event.features.iter_mut().find(|f| f.name == feature.name) {
+                    *existing = feature;
+                } else {
+                    self.hydration_event.features.push(feature);
+                }
+                self.hydration_event.event_id = event_id;
```
sighphyre, line 87: I think having this repeated event setting line in in 4 out of the 5 enum cases is a little fragile to future updates. You also went to all that trouble of creating a get_event_id method, so can I suggest we do this rather

1) Remove the calls to `self.hydration_event.event_id = event_id` in the match arms
2) Add a ```
        if !matches!(event, DeltaEvent::Hydration { .. }) {
            self.hydration_event.event_id = event.get_event_id();
        }
```

At the top of the method

File: server/src/delta_cache.rs [hunk 6]
```diff
@@ -0,0 +1,286 @@
+use std::collections::VecDeque;
+use unleash_types::client_features::{DeltaEvent, ClientFeature, Segment};
+
+
+
+#[derive(Debug, Clone)]
+pub struct DeltaHydrationEvent {
+    event_id: i32,
```
sighphyre, line 8: Should this not be a `u32` or a `usize`? Probably not a big deal just seems weird to add scope for negative numbers when they can't be negative by design in this feature

server/src/delta_cache.rs
sighphyre: Can we rather just move this method into the `impl` in the other crate? Events Ids are pretty core to the concept of events, I see no reason to scatter the logic over multiple points if we don't need to

sjaanus: Yes, we can move it. We actually did implement it in rs-types and then moved the code to edge as trait so we would not need to create new version. But we can go back easily.

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+            self.events.push_back(event.clone());
+            self.update_hydration_event(event);
+            if self.events.len() > self.max_length {
+                self.events.pop_front();
+            }
+        }
+    }
+
+    pub fn get_events(&self) -> &VecDeque<DeltaEvent> {
+        &self.events
```
sighphyre, line 68: Is the plan to kick this exact VecDeque over the HTTP boundary to downstream SDKs? I ask because if this needs to be massaged into something else first, it's going to mean doing a copy before we do a copy for sending it.

For large feature sets this can get really expensive

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+impl DeltaCache {
+    pub fn new(hydration_event: DeltaHydrationEvent, max_length: usize) -> Self {
+        let mut cache = DeltaCache {
+            max_length,
+            events: VecDeque::new(),
+            hydration_event: hydration_event.clone(),
+        };
+        cache.add_base_event_from_hydration(hydration_event);
```
sighphyre, line 44: This is more a nit than something that needs to be done (and it's good to start thinking like a Rust developer).

`add_base_event_from_hydration` currently takes an owned type but then clones the stuff it needs from that. So it doesn't _really_ need an owned type. If that took a borrow, you could invert this call with the call above it and save yourself a potentially expensive copy

server/src/delta_cache.rs
sjaanus: We did it for consistency, how other ids have been implemented in types. For example segment id, 
variant id are are i32.


https://github.com/Unleash/unleash-types-rs/blob/d782b656f6bd4887bc8251a95bbbd6b004612aeb/src/client_features.rs#L304

https://github.com/Unleash/unleash-types-rs/blob/d782b656f6bd4887bc8251a95bbbd6b004612aeb/src/client_features.rs#L266

server/src/delta_cache.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+}
+
+
+#[test]
+fn test_update_hydration_event_and_remove_event_when_over_limit() {
+    let base_event = DeltaHydrationEvent {
+        event_id: 1,
+        features: vec![
+            ClientFeature {
```
sighphyre, line 118: IIRC correctly `ClientFeature` derives `Default`, which means these big ass declarations of ClientFeature can be simplified to something like

`ClientFeature {
  name: "test-flag".to_string(),
  ///other stuff you care about
  ..ClientFeatures::default()}
}`

Makes the tests much easier to navigate and saves future maintenance work when someone adds a new field to `ClientFeature`

server/src/delta_cache.rs
sjaanus: 99% of times we take 1-5 events out of the VecDeque and push that over HTTP.

server/src/delta_cache.rs
sighphyre: Fair enough but I'd still like to see this as an unsigned type when dealing with a monotonically increasing id

Won't fuss if you don't care though

server/src/delta_cache.rs
sjaanus: Yes, added that also to test.

server/src/delta_cache.rs
sjaanus: Yes

server/src/delta_cache.rs
sjaanus: Moved

server/src/delta_cache.rs
sjaanus: Done

sighphyre (approved):
:+1:

PR #742, opened 2025-02-17
feat: start storing actual deltas in edge

sjaanus:
This is basic PoC delta caching and endpoint for /delta.

Next PRs will include:
1) Tests
2) Handling etags and event filtering
2) Solving TODO remarks (offline mode?, redis persistance?)

### DIFF
--- a/server/src/builder.rs
+++ b/server/src/builder.rs
@@ -315,7 +315,7 @@ async fn build_edge(
             .await;
     }
     Ok((
-        (token_cache, feature_cache, _delta_cache, engine_cache),
+        (token_cache, feature_cache, delta_cache, engine_cache),
         Some(token_validator),
         Some(feature_refresher),
         persistence,
--- a/server/src/builder.rs
+++ b/server/src/builder.rs
@@ -26,10 +26,12 @@ use crate::{
     http::{refresher::feature_refresher::FeatureRefresher, unleash_client::UnleashClient},
     types::{EdgeResult, EdgeToken, TokenType},
 };
+use crate::delta_cache::DeltaCache;
 
 type CacheContainer = (
     Arc<DashMap<String, EdgeToken>>,
     Arc<FeatureCache>,
+    Arc<DashMap<String, DeltaCache>>,
     Arc<DashMap<String, EngineState>>,
 );
 type EdgeInfo = (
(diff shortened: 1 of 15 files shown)


server/src/delta_cache.rs
```diff
@@ -35,72 +32,87 @@ impl DeltaCache {
...
+            .iter()
+            .any(|event| event.get_event_id() == revision_id)
     }
 
     pub fn get_hydration_event(&self) -> &DeltaHydrationEvent {
         &self.hydration_event
     }
 
-    fn update_hydration_event(&mut self, event: DeltaEvent) {
+    fn update_hydration_event(&mut self, event: &DeltaEvent) {
```
sjaanus, line 67: Passing in reference and cloning just parts of object.

server/src/builder.rs
```diff
@@ -42,16 +44,19 @@ type EdgeInfo = (
...
         Arc::new(FeatureCache::new(features_cache)),
+        Arc::new(delta_cache),
         Arc::new(engine_cache),
     )
 }
 
 async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn EdgePersistence>) {
-    let (token_cache, features_cache, engine_cache) = cache;
+    let (token_cache, features_cache, _delta_cache, engine_cache) = cache;
+    // TODO: do we need to hydrate from persistant storage for delta?
```
sighphyre, line 59: Yes I think so. Otherwise when pods spin up they'll send empty responses until they hydrate. Not sure how we intend to handle that yet but I guess that's next PR

server/src/builder.rs
```diff
@@ -135,7 +140,8 @@ pub(crate) fn build_offline_mode(
             client_features.clone(),
         )
     }
-    Ok((token_cache, features_cache, engine_cache))
+    // TODO: possibly need to resolve delta cache for offline mode?
```
sighphyre, line 143: Personal feeling but I'd skip this one for now. I don't see much value in sending deltas when that event stream is frozen

server/src/http/refresher/delta_refresher.rs
```diff
@@ -31,6 +34,32 @@ impl FeatureRefresher {
...
+        {
+            self.delta_cache.insert(
+                key.clone(),
+                DeltaCache::new(
+                    DeltaHydrationEvent {
+                        event_id,
+                        features,
+                        segments,
+                    },
+                    100,
```
sighphyre, line 54: Can we put this in a nicely named const somewhere rather?

server/src/http/refresher/delta_refresher.rs
```diff
@@ -31,6 +34,32 @@ impl FeatureRefresher {
...
+                ),
+            );
+        } else {
+            warn!(
```
sighphyre, line 58: Can this actually happen? Should this happen if the code is correct? This feels like we've broken some system constraint here

server/src/http/refresher/feature_refresher.rs
```diff
@@ -249,6 +254,33 @@ impl FeatureRefresher {
         }
     }
 
+    pub(crate) async fn delta_events_for_filter(
```
sighphyre, line 257: I think this is more or less a copy paste of `features_for_filter`. That's sometimes okay but I think in this case the subsumption and strict mode are fairly important system constraints that I think should change in lock step if they ever need to.

Is it tremendously difficult to get them to share the important stuff?

server/src/main.rs
```diff
@@ -80,6 +80,7 @@ async fn main() -> Result<(), anyhow::Error> {
     let token_validator_schedule = token_validator.clone();
     let lazy_feature_cache = features_cache.clone();
     let lazy_token_cache = token_cache.clone();
+    // TODO : do we need lazy delta cache?
```
sighphyre, line 83: Nope, if it compiles then you don't need it. The lazy stuff is to handle cases where something is used more than once later in this function and we need a reference

Might need this if we add this to the persistent store layer

```suggestion
```

sighphyre (approved):
Sure I think this looks okay. It's a bit hard to validate properly without tests and without the decisions around caching that still need to be made but this should be safe to merge

server/src/http/refresher/delta_refresher.rs
sjaanus: No, it the code is correct, it should never happen.

server/src/http/refresher/feature_refresher.rs
sjaanus: Yes

PR #749, opened 2025-02-19
feat: delta filtering and etag handling

sjaanus:
This PR handles 2 things

1. Etag handling, so we know what etag we receiving and what to send back
2. Filtering features based on etag and query params

Next PR

1. Automatic tests for client API.
2. Connect to delta cache manager, which is also being developed currently.

### DIFF
--- a/server/src/client_api.rs
+++ b/server/src/client_api.rs
@@ -49,16 +49,30 @@ pub async fn get_delta(
     filter_query: Query<FeatureFilters>,
     req: HttpRequest,
 ) -> impl Responder {
-    let requested_revision_id = req.headers().get("If-None-Match")
+    let requested_revision_id = req
+        .headers()
+        .get("If-None-Match")
         .and_then(|value| value.to_str().ok())
         .and_then(|etag| etag.trim_matches('"').parse::<u32>().ok())
         .unwrap_or(0);
 
     let current_sdk_revision_id = 100; // TODO: Read from delta_manager
 
-    match resolve_delta(edge_token, token_cache, filter_query, requested_revision_id, req).await {
+    match resolve_delta(
+        edge_token,
+        token_cache,
+        filter_query,
+        requested_revision_id,
+        req,
+    )
+    .await
+    {
         Some(Ok(delta)) => {
-            let last_event_id = delta.events.last().map(|e| e.get_event_id()).unwrap_or(current_sdk_revision_id);
+            let last_event_id = delta
+                .events
+                .last()
+                .map(|e| e.get_event_id())
+                .unwrap_or(current_sdk_revision_id);
 
             HttpResponse::Ok()
                 .insert_header(("ETag", format!("\"{}\"", last_event_id)))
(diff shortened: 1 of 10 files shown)


server/src/client_api.rs
```diff
@@ -47,12 +45,42 @@ pub async fn get_features(
...
 pub async fn get_delta(
     edge_token: EdgeToken,
-    delta_cache: Data<DashMap<String, DeltaCache>>,
     token_cache: Data<DashMap<String, EdgeToken>>,
     filter_query: Query<FeatureFilters>,
     req: HttpRequest,
-) -> EdgeJsonResult<ClientFeaturesDelta> {
-    resolve_delta(edge_token, delta_cache, token_cache, filter_query, req).await
+) -> impl Responder {
+    let requested_revision_id = req
```
sjaanus, line 52: Get etag from sdk

server/src/filters.rs
```diff
@@ -60,16 +51,6 @@ pub(crate) fn filter_client_features(
     }
 }
 
-pub(crate) fn filter_delta_events(
```
sjaanus, line 63: This is moved to delta_filters

server/src/client_api.rs
```diff
@@ -134,6 +162,27 @@ fn get_feature_filter(
...
+) -> EdgeResult<DeltaFilterSet> {
+    let validated_token = token_cache
+        .get(&edge_token.token)
+        .map(|e| e.value().clone())
+        .ok_or(EdgeError::AuthorizationDenied)?;
+
+    let query_filters = filter_query.into_inner();
+
+    let delta_filter_set = DeltaFilterSet::default().with_filter(combined_filter(
+        100,
```
sighphyre, line 178: What's 100 here?

server/src/client_api.rs
```diff
@@ -164,27 +213,52 @@ async fn resolve_features(
 
 async fn resolve_delta(
```
sighphyre, line 214: I think the return type of Option<Result<stuff>> is making your life much harder than it needs to be around error handling here and it makes semantic understanding of the function challenging. Usually this is expressed as Result<Option<stuff>> for this reason. I also don't think this function has any business converting things to Json, based on its name, that's a higher level responsibility if we can possibly help it. We can also move the check for the feature refresher earlier where it can act as a guard clause before we need to do any work for the building of fiilters. This can all be expressed as

```
async fn resolve_delta(
    edge_token: EdgeToken,
    token_cache: Data<DashMap<String, EdgeToken>>,
    filter_query: Query<FeatureFilters>,
    requested_revision_id: u32,
    req: HttpRequest,
) -> EdgeResult<Option<ClientFeaturesDelta>> {
    let Some(refresher) = req.app_data::<Data<FeatureRefresher>>() else {
        return Err(EdgeError::ClientHydrationFailed(
            "FeatureRefresher is missing - cannot resolve delta in offline mode".to_string(),
        ));
    };
    let (validated_token, filter_set, ..) =
        get_feature_filter(&edge_token, &token_cache, filter_query.clone())?;
    let delta_filter_set = get_delta_filter(&edge_token, &token_cache, filter_query.clone())?;

    let current_sdk_revision_id = requested_revision_id + 1; // TODO: get from delta manager
    if requested_revision_id >= current_sdk_revision_id {
        return Ok(None);
    }

    let delta = refresher
        .delta_events_for_filter(
            validated_token.clone(),
            &filter_set,
            &delta_filter_set,
            requested_revision_id,
        )
        .await?;

    if !delta.events.is_empty() {
        Ok(Some(delta))
    } else {
        Ok(None)
    }
}
```

sjaanus: Regarding JSON, this is following existing pattern.

https://github.com/Unleash/unleash-edge/blob/9a26011d13aceb87f146d6ebc0e6a6d9619de0c5/server/src/client_api.rs#L163

sjaanus: Also, Reworked the Option as you suggested. Now it is inside instead. I think its better now.

server/src/client_api.rs
sjaanus: Thanks, this was oversight actually. fixed.

server/src/client_api.rs
sighphyre: > Regarding JSON, this is following existing pattern.
> 
> https://github.com/Unleash/unleash-edge/blob/9a26011d13aceb87f146d6ebc0e6a6d9619de0c5/server/src/client_api.rs#L163

Yeah but that doesn't work so well with Options, which is why I suggested not using that. I think what you've done is okay though

sighphyre (approved):
Sure

PR #769, opened 2025-02-24
feat: count requests seen since last report

chriswk:
And then we leave it up to Unleash to put it in the correct bucket.


gastonfournier (approved):
LGTM!

server/src/metrics/edge_metrics.rs
```diff
@@ -76,6 +85,25 @@ impl EdgeInstanceData {
...
+        }
+    }
+
+    pub fn observe_request(&self, http_target: &str, status_code: i64) {
+        match status_code {
+            200 => {
+                self.requests_since_last_report
+                    .entry(http_target.to_string())
+                    .or_default()
+                    .requests_200 += 1
```
sighphyre, line 98: I think this isn't thread safe. Docs are a bit hard to grok here but dashmap controls request_200 but I don't think it's able to wrap access and modify in in the same lock across function calls

Can't prove it but this'd keep me up at night

sighphyre (approved):
LGTM!

PR #776, opened 2025-02-25
fix: filter segment events by revision

kwasniew:
## About the changes

We were not filtering segment events by revision so for segment events we were getting extra events

### DIFF
--- a/server/src/delta_filters.rs
+++ b/server/src/delta_filters.rs
@@ -72,7 +72,7 @@ pub(crate) fn combined_filter(
     let segment_filter = is_segment_event_filter();
 
     Box::new(move |event| {
-        segment_filter(event)
+        (segment_filter(event) && revision_filter(event))
             || (name_filter(event) && projects_filter(event) && revision_filter(event))
     })
 }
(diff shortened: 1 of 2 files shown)


server/src/delta_filters.rs
```diff
@@ -72,7 +72,7 @@ pub(crate) fn combined_filter(
     let segment_filter = is_segment_event_filter();
 
     Box::new(move |event| {
-        segment_filter(event)
+        (segment_filter(event) && revision_filter(event))
```
kwasniew, line 75: this was missing

server/src/delta_filters.rs
```diff
@@ -91,7 +91,7 @@ fn filter_deltas(
         let events = delta_cache.get_events().clone();
         events
             .iter()
-            .filter(|delta| delta_filters.apply(delta) && delta.get_event_id() >= revision)
+            .filter(|delta| delta_filters.apply(delta))
```
kwasniew, line 94: we don't need it here since we have this as part of delta filters

sighphyre (approved):
Nice catch!

PR #780, opened 2025-02-26
feat: app_name and instance_id added as label to metrics

chriswk:
In order to be able to separate metrics in grafana. this adds two more labels to the unleash_client prometheus metrics, so that we have the needed labels even in hosted edge.

### DIFF
--- a/server/src/http/unleash_client.rs
+++ b/server/src/http/unleash_client.rs
@@ -39,34 +39,34 @@ lazy_static! {
             "client_register_failures",
             "Why we failed to register upstream"
         ),
-        &["status_code"]
+        &["status_code", "app_name", "instance_id"]
     )
     .unwrap();
     pub static ref CLIENT_FEATURE_FETCH: HistogramVec = register_histogram_vec!(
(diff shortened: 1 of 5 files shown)


server/src/http/unleash_client.rs
```diff
@@ -224,19 +225,29 @@ pub struct EdgeTokens {
...
+        token_header: String,
+        backing_client: Client,
+        client_meta_information: ClientMetaInformation,
+    ) -> Self {
         Self {
             urls: UnleashUrls::from_base_url(server_url),
             backing_client,
             custom_headers: Default::default(),
             token_header,
+            meta_info: client_meta_information.clone(),
```
sighphyre, line 239: Do we need the clone? This looks like it takes an owned type

```suggestion
            meta_info: client_meta_information,
```

sighphyre (approved):
Yeah nice, LGTM

server/src/http/unleash_client.rs
chriswk: No, we didn't. Fixed.

PR #782, opened 2025-02-27
feat: No more openssl

chriswk:
No longer needs native-tls, so massively simplifies building edge. Tests in place to verify that we can load p12, p8, pem as we keep rustls and webpki up to date.


server/src/client_api.rs
chriswk, line 1: This is just clippy.

sighphyre (approved):
Yep

PR #785, opened 2025-03-03
feat: unleash-interval header

kwasniew:
## About the changes

Background:
* we're adding unleash-interval header for all fetch flags calls and send metrics calls in backend SDKs. Since edge itself can be treated as backend SDK we're also adding those headers for Edge itself

Design decisions:
* fetching flags and sending client metrics take optional interval (not every fn call has to be from a scheduler so that's why it's optional)

Doubts:
* should Edge fetch interval sync with other SDKs? (15s instead of 10s by default)
* even though we have BatchMetricsRequest similar to ClientFeaturesRequest we don't use BatchMetricsRequest at all. As a result I added a new param to send_batch_metrics since I didn't want to pollute MetricsBatch used in other places
* should we add connection-id to edge? In other backend SDKs we introduced unleash-connection-id unique identifier (Unleash controlled equivalent of instance-id new'ed up on new SDK instance) + we pass it as a header to Unleash + we add it to registration call and metrics payload calls. I'm wondering if it's worth adding to edge. Unleash-connection-id is not used for connection counting anymore but acts as a double-entry bookkeeping mechanism that can be used to check if counters were correct or if someone is not creating too many SDK instances
* I didn't find any tests for server receiving headers from the client


server/src/http/background_send_metrics.rs
```diff
@@ -87,7 +87,7 @@ pub async fn send_metrics_one_shot(
                 } else {
                     feature_refresher
                         .unleash_client
-                        .send_batch_metrics(batch.clone())
+                        .send_batch_metrics(batch.clone(), None)
```
kwasniew, line 90: first sync is not from the scheduler so no interval passed

server/src/http/background_send_metrics.rs
```diff
@@ -123,7 +123,7 @@ pub async fn send_metrics_task(
                     } else {
                         feature_refresher
                             .unleash_client
-                            .send_batch_metrics(batch.clone())
+                            .send_batch_metrics(batch.clone(), Some(interval.num_milliseconds()))
```
kwasniew, line 126: even though Edge operates on seconds we send ms on the wire as the lowest common denominator between SDKs

server/src/http/refresher/feature_refresher.rs
```diff
@@ -454,6 +454,7 @@ impl FeatureRefresher {
             .get_client_features_delta(ClientFeaturesRequest {
                 api_key: refresh.token.token.clone(),
                 etag: None,
+                interval: None,
```
kwasniew, line 457: this is from compare_data_cache which shouldn't add counts

server/src/http/unleash_client.rs
```diff
@@ -538,19 +544,24 @@ impl UnleashClient {
         }
     }
 
-    pub async fn send_batch_metrics(&self, request: MetricsBatch) -> EdgeResult<()> {
+    pub async fn send_batch_metrics(
+        &self,
+        request: MetricsBatch,
```
kwasniew, line 549: I expected a different type here (e.g. BatchMetricsRequest similarly to ClientFeaturesRequest in the fetch flags call), but since we have only MetricsBatch I am adding interval as a new param

server/src/http/unleash_client.rs
```diff
@@ -285,15 +286,20 @@ impl UnleashClient {
     }
 
     fn client_features_req(&self, req: ClientFeaturesRequest) -> RequestBuilder {
-        let client_req = self
+        let mut client_req = self
```
sighphyre, line 289: We don't need this, right?

```suggestion
        let client_req = self
```

sighphyre (approved):
Yeah looks good

sighphyre:
> should Edge fetch interval sync with other SDKs? (15s instead of 10s by default)

Yes I think that makes sense

> even though we have BatchMetricsRequest similar to ClientFeaturesRequest we don't use BatchMetricsRequest at all. As a result I added a new param to send_batch_metrics since I didn't want to pollute MetricsBatch used in other places

I believe BatchMetricsRequest is now dead code

> should we add connection-id to edge? In other backend SDKs we introduced unleash-connection-id unique identifier (Unleash controlled equivalent of instance-id new'ed up on new SDK instance) + we pass it as a header to Unleash + we add it to registration call and metrics payload calls. I'm wondering if it's worth adding to edge. Unleash-connection-id is not used for connection counting anymore but acts as a double-entry bookkeeping mechanism that can be used to check if counters were correct or if someone is not creating too many SDK instances

I think we should, yes. Edge is effectively an SDK it just SDK things a little differently from other SDKs so it should follow their behavior unless there's a strong reason not to

> I didn't find any tests for server receiving headers from the client
They probably don't exist. Headers have been an implementation detail until now. Now they're business logic, so you're free to add some if you think they're important!

chriswk (approved):
I see Simon has already commented. Just adding another :+1:

server/src/http/background_send_metrics.rs
kwasniew: I was hesitant to do this since it's not called from the scheduler (negligible one-off call as opposed to regular scheduled calls) + I'd need to drag the interval through server function calls (main -> clean_shutdown -> send_metrics_one_shot -> send_batch_metrics). I'm willing to re-visit this decision though :)

kwasniew:
Based on the feedback I will create followup PRs:
* update interval 10s -> 15s
* connection id support (+ header tests)

PR #786, opened 2025-03-03
feat: add allow-list and deny-list middlewares

chriswk:
Uses https://github.com/Unleash/actix-allow-deny-middleware to configure an allow- and a deny-list for access to the server. By default configures the allow list to allow all IPs and the deny list to be empty (deny no IPs).


server/src/http/instance_data.rs
```diff
@@ -115,7 +115,7 @@ pub async fn loop_send_instance_data(
         match instance_data_sender.as_ref() {
             InstanceDataSending::SendNothing => {
                 debug!("No instance data sender found. Doing nothing.");
-                return;
+                continue;
```
sighphyre, line 118: Just a suggestion but can we put this in another PR so we can get the last patch version in the minor range working correctly

sighphyre (approved):
LGTM, pretty cool

server/src/http/instance_data.rs
chriswk: yes. will do

PR #787, opened 2025-03-03
feat: sync refresh-interval with other SDKs

kwasniew:
## About the changes

We have an ongoing initiative to sync refresh intervals across SDKs. This PR sets fetch flags refresh interval to 15 seconds instead of 10. Also updated README


sighphyre (approved):
LGTM

PR #789, opened 2025-03-03
feat: unleash-connection-id header

kwasniew:
## About the changes

Add unleash-connection-id header to all requests we make to  upstream server (both polling and streaming).

In a followup PR I will also add connectionId to metrics payload and instance registration payload after another PR is merged (https://github.com/Unleash/unleash-types-rs/pull/81)

Another followup PR will add tests for the unleash-connection-id and unleash-interval

## Discussion points

I'm not sure about ConnectedVia (very similar to ClientMetaInformation) currently takes app_name and instance_id. Should we include connection_id in ConnectedVia?

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -1,3 +1,4 @@
+use actix_allow_deny_middleware::{AllowList, DenyList};
 use actix_middleware_etag::Etag;
 use actix_web::middleware::Logger;
 use actix_web::{web, App, HttpServer};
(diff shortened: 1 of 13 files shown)


server/src/main.rs
```diff
@@ -207,7 +209,8 @@ async fn main() -> Result<(), anyhow::Error> {
                             .start_streaming_delta_background_task(
                                 ClientMetaInformation {
                                     app_name,
-                                    instance_id,
+                                    instance_id: identifier.clone(),
```
kwasniew, line 212: it looks like streaming never respected user provided instance_id and always took our_instance_data_for_app_context.idnetifier

chriswk: indeed.

chriswk (approved):
LGTM

PR #790, opened 2025-03-04
feat: add connection-id to registration payload

kwasniew:
## About the changes

https://github.com/Unleash/unleash-types-rs/pull/81 added connection_id to registration payload and client metrics coming to Edge.
This PR fixes the types.

The connectionId will be added only to the Edge registering to Unleash since Edge does it the same as any other SDK.
However other SDKs also add connectionId to the metrics payload they send. Since Edge has a different way of sending metrics (batch metrics from multiple apps) then we could add Edge own connectionId to ConnectedVia that already has instanceId. Please let me know what you think about it?

The last PR in the series will add some tests for the new unleash-interval and unleash-connection headers.


server/src/metrics/client_metrics.rs
```diff
@@ -629,13 +630,14 @@ mod test {
     #[test_case(1, 10000, 25; "1 app 10k toggles, will be split into 25 batches")]
     #[test_case(1000, 1000, 7; "1000 apps 1000 toggles, will be split into 7 batches")]
     #[test_case(500, 5000, 15; "500 apps 5000 toggles, will be split into 15 batches")]
-    #[test_case(5000, 1, 17; "5000 apps 1 metric will be split")]
+    #[test_case(5000, 1, 19; "5000 apps 1 metric will be split")]
```
kwasniew, line 633: I don't know why but the change changed the number of batches. It looks like we did similar updates in the past.

chriswk: Probably because the size increased. the connection_id field name alone is 13 bytes extra per entry.

chriswk: 13 bytes for key,
connection-${ulid} is 47 bytes more, so 60 bytes extra per application.

chriswk (approved):
LGTM

PR #792, opened 2025-03-04
task: allow/deny middlewares only active for /api and /edge paths.

chriswk:
In order to simplify deployment and not inadvertently break health and ready checks when applying the allow or deny list middlewares, this PR only wraps "/api" and "/edge" endpoints, rather than "/*".


sastromskis (approved):
LGTM 🚀

PR #794, opened 2025-03-04
test: identification headers

kwasniew:
## About the changes

Verifying that connection-id and interval headers are sent. Modified existing custom headers test to make it more generic. If you'd prefer a separate test please let me know

Also looks like formatter has change a few lines.

### DIFF
--- a/server/src/http/unleash_client.rs
+++ b/server/src/http/unleash_client.rs
@@ -10,9 +10,9 @@ use actix_web::http::header::EntityTag;
 use chrono::Duration;
 use chrono::Utc;
 use lazy_static::lazy_static;
-use prometheus::{register_histogram_vec, register_int_gauge_vec, HistogramVec, IntGaugeVec, Opts};
+use prometheus::{HistogramVec, IntGaugeVec, Opts, register_histogram_vec, register_int_gauge_vec};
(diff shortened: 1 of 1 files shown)


server/src/http/unleash_client.rs
```diff
@@ -50,14 +50,18 @@ lazy_static! {
         "client_feature_fetch",
         "Timings for fetching features in milliseconds",
         &["status_code", "app_name", "instance_id"],
-        vec![1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 5000.0]
+        vec![
+            1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 5000.0
+        ]
```
chriswk, line 55: Is this cargo fmt changes, or is this your editor @kwasniew ?

kwasniew: cargo fmt after rust toolchain update. It did it to the entire codebase but I didn't want to pollute this PR

chriswk: Not worried about this then. :)

chriswk (approved):
LGTM

PR #802, opened 2025-03-06
fix: actix prometheus metrics should also report in milliseconds

chriswk:
Inadvertently forgot that we were using milliseconds, not seconds in the old middleware. This PR reverts back to using milliseconds rather than seconds for request durations.


sighphyre (approved):
LGTM

PR #803, opened 2025-03-07
fix: make sure we sort the list of frontend results by feature name

### LINKED ISSUE #797: Unleash Edge responds with 200 even if there is no change
### Describe the bug

I am working on a project based on React with unleash-react-proxy 4.5.0 and latest unleash edge.

When querying changes the Unleash Edge responds with 200 even if there are no changes present, which results in React state update.

If connected directly to Unleash server i am getting 304 when no changes are made to the flags.
Is this a bug or the expected result from unleash edge ? 

1. Setup Unleash with Unleash Edge as per documentation
2. Setup React application with proxy unleash client
3. create a useEffect for useFlags and check how many times the useEffect is triggered.

### Expected behavior

_No response_

### Logs, error output, etc.

```bash

```

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

Unleash Edge: 19.7.0
React Proxy: 4.4.0

### Subscription type

None

### Hosting type

Self-hosted

_No response_

chriswk:
With a great example in #797 for why they get 200 on their frontend.

We figured out that we didn't have a stable sort for our feature flags with the frontend API. 

This PR adds sorted_by_key to the list of frontend_results and a test to verify that we get 304 for multiple requests, removing the sorted_by_key causes the test to fail. 

fixes #797

### DIFF
--- a/examples/frontend-stable-sort.json
+++ b/examples/frontend-stable-sort.json
@@ -1,0 +1,76 @@
+{
+  "version": 2,
+  "features": [
+    {
+      "name": "my-projects",
+      "type": "release",
+      "enabled": true,
+      "stale": false,
+      "impressionData": false,
+      "project": "default",
+      "strategies": [
+        {
+          "name": "flexibleRollout",
+          "constraints": [],
+          "parameters": {
+            "groupId": "my-projects",
+            "rollout": "100",
+            "stickiness": "default"
+          },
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
LGTM! Still testing E2E with the sample code to triple check, but this looks good to me!

PR #811, opened 2025-03-10
fix: tls with redis

sighphyre:
Fixes an issue where TLS setup would fail if used alongside Redis in secure mode. This occurs because the Redis bindings use exactly the same code and that code can be executed at most once per process. 

Fixes an issue that we introduced with the upgrade to 19.4


gastonfournier (approved):
LGTM!

PR #829, opened 2025-03-17
feat: consumption metrics

kwasniew:
## About the changes

This PR introduces consumption tracking for both request-based and connection-based traffic in the edge service.

### Key Features
- **Dual Consumption Models**:
  - Request Consumption: Simple request counting for frontend traffic
  - Connection Consumption: Interval-based counting for backend traffic with long-lived connections
  - Special consideration for Lambda model traffic (backend APIs without long-lived connections) - TBD how to do it

- **Metrics Structure**:
  - Backend consumption metrics are bucketed by interval ranges
  - Default bucket ranges: [0, 15s] for features, [0, 60s] for metrics
  - Support for custom intervals with 5-second bucket granularity
  - Maximum bucket for intervals > 1 hour

- **Observability Integration**:
  - Thread-safe consumption metrics using `AtomicU64` and `DashMap`
  - Efficient serialization with custom implementations
  - Clear separation between runtime data structures and serialized format
  - Metrics cleared following existing patterns

### Technical Details
- Thread-safe implementation using `DashMap` and `AtomicU64`
- Custom serialization/deserialization for atomic types
- Efficient memory usage by avoiding redundant data storage
- Bucket ranges calculated based on endpoint type and interval

### Testing
- Comprehensive test coverage for bucket assignment logic
- Tests for boundary conditions in interval ranges
- Verification of metrics clearing functionality

### Important files

Currently generated structure (not set in stone):
```json
{
  "connectionConsumptionSinceLastReport": {
    "features": [{
      "meteredGroup": "default",
      "dataPoints": [
        {
          "interval": [0, 15000],
          "requests": 42
        },
        {
          "interval": [15000, 20000],
          "requests": 15
        }
      ]
    }],
    "metrics": [{
      "meteredGroup": "default",
      "dataPoints": [
        {
          "interval": [0, 60000],
          "requests": 156
        },
        {
          "interval": [60000, 120000],
          "requests": 45
        }
      ]
    }]
  },
  "requestConsumptionSinceLastReport": [{
    "meteredGroup": "default",
    "requests": 278
  }]
}
```

## Discussion points

Should we put it behind a flag or CLI arg? If so which flag/arg should I take the inspiration from?

### DIFF
--- a/server/src/metrics/edge_metrics.rs
+++ b/server/src/metrics/edge_metrics.rs
@@ -697,13 +697,11 @@ mod tests {
     fn can_observe_and_clear_consumption_metrics() {
         let instance_data = EdgeInstanceData::new("test");
 
-        // Observe some consumption
         instance_data.observe_request_consumption();
         instance_data.observe_request_consumption();
         instance_data.observe_request_consumption();
         instance_data.observe_request_consumption();
 
-        // Verify request consumption through EdgeInstanceData serialization
         let serialized = serde_json::to_value(&instance_data).unwrap();
         assert_eq!(
             serialized["requestConsumptionSinceLastReport"],
@@ -713,7 +711,6 @@ mod tests {
             })
         );
 
-        // Verify connection consumption through EdgeInstanceData serialization
         assert_eq!(
             serialized["connectionConsumptionSinceLastReport"],
             serde_json::json!({
@@ -728,10 +725,8 @@ mod tests {
             })
         );
 
-        // Clear metrics
         instance_data.clear_time_windowed_metrics();
 
-        // Verify cleared data through EdgeInstanceData serialization
         let serialized_cleared = serde_json::to_value(&instance_data).unwrap();
         assert_eq!(
             serialized_cleared["requestConsumptionSinceLastReport"],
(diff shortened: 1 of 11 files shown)


chriswk (commented):
Bit surprised that you needed a custom Serde implementation here, but looks good.

Your middleware will trigger before the request is executed, so you can't differentiate 200 - 304, is that intentional for now?

kwasniew:
@chriswk i may change the order or middleware but we count everything between 200-3xx and don't count anything >= 400. 
With serde I wanted to have easy to work with DashMap + Atomics when counting but also convenient to consume structure for Unleash that we serialize to. Originally I serialized nested DashMaps but they were too magical to parse on the Unleash side.

server/src/metrics/edge_metrics.rs
```diff
@@ -66,6 +65,234 @@ impl Clone for RequestStats {
...
+    }
+
+    pub fn as_array(&self) -> [u64; 2] {
+        [self.start, self.end]
+    }
+
+    pub fn from_array(arr: [u64; 2]) -> Self {
+        Self::new(arr[0], arr[1])
+    }
+}
```
chriswk, line 137: Hmm, did you think about using a built in std::ops::Range or std::ops::InclusiveRange here?

chriswk: I guess since you're not really doing any operations on this on the edge side, it isn't needed.

chriswk (approved):
I'm happy with this.

kwasniew:
@chriswk added code for only counting consumption of 2xx and 304 for now.

Issue #859, opened 2025-04-01
CORS configurarion for edge


synnefon:
Does edge support cors configuration settings? If not, our team would like our web apps to be able to leverage this.

I see that this was previously asked here, and marked as resolved: https://github.com/Unleash/unleash-edge/issues/701. I also see the relevant mr, and and just hoping to verify that it has been released to prod. 

Still, looking at your docs, this looks pending still: https://github.com/Unleash/unleash-edge/blob/main/migration-guide.md#not-supported.

Can you clarify the current state?

Thanks!

chriswk:
Hi @synnefon . I'll flag that we have to update the migration-guide.

If you look at the CLI instructions https://github.com/Unleash/unleash-edge/blob/main/CLI.md you can see the cors options. They are indeed functional

chriswk:
I've made #864 to track the work on the docs

synnefon:
Thanks! Super helpful

Issue #864, opened 2025-04-02
docs: Update migration guide to say that we now support CORS


chriswk:
### Describe the feature request

I want Edge to correctly state that it supports CORS.

### Background

#859 pointed out that the migration guide is still saying that CORS is not updated.

PR #865, opened 2025-04-02
docs: Remove CORS limitation from migration guide

chriswk:
Also 
fixes: #859 and #864

### DIFF
--- a/migration-guide.md
+++ b/migration-guide.md
@@ -34,9 +34,6 @@ curl --location --request GET 'http://0.0.0.0:3063/api/client/features' \
   same tokens that the front end API requires. Because of the way Edge handles API tokens, this is not a feature we're
   planning to support.
 
-- CORS configuration. Edge does not support CORS configuration, we accept any request. If you need this feature, please
-  raise an issue or submit a PR.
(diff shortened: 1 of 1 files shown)


docs/deploying.md
```diff
@@ -4,6 +4,35 @@
...
+### CORS
+
+`--cors-origin <CORS_ORIGIN>` — Sets
+  the [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)
+  header to this value
+
+To enable CORS headers, you can use the `--cors-allowed-origins` flag when starting Edge. For example:
+
+```shell
+unleash-edge edge --cors-origin "https://example.com"
```
nunogois, line 16: The above statement says `--cors-allowed-origins` but here we see `--cors-origin`. I'm guessing only one of them is correct?

Also, maybe an example with multiple values could be useful? Should they be comma separated?

chriswk: Yes, fair point. I guess one of them is AI generated. I'll have another look.

PR #877, opened 2025-04-09
fix: redact path to /{unknown} if status is not success

chriswk:
With a publicly available URL we should redact path for any status code that did not yield success.


sighphyre (approved):
Nice catch!

PR #889, opened 2025-04-11
docs: fix formatting issue in cli.md

melindafekete:
Fixing a pair of unescaped < and > in CLI.md. Without code formatting, they are treated as the start and end of HTML tags and break the rendering. The MDX parser is also unhappy and throws an error in the docs.


sighphyre (approved):
LGTM!

PR #895, opened 2025-04-15
chore: move s3 persistence behind flag

sighphyre:
Moves the S3 persister behind a flag so we can optionally not compile the AWS SDK stuff because it's really really heavy. This lets us run builds with --no-default-features. That's about 30-40% faster

Left this on by default so it's opt out


chriswk (approved):
LGTM

PR #898, opened 2025-04-17
chore: trusted token plumbing

sighphyre:
Adds a pathway to create trusted tokens in Unleash Edge. These tokens are allowed to entirely skip validation so you can have tokens that look like "haha-I'm-totally-a-valid-token-bro-trust-me". This PR does nothing on its own, it's purely the plumbing layer but this allows us to inject these tokens on startup (coming in a future PR) into the cache and use Proxy style tokens

### DIFF
--- a/server/src/auth/token_validator.rs
+++ b/server/src/auth/token_validator.rs
@@ -18,6 +18,21 @@ pub struct TokenValidator {
     pub persistence: Option<Arc<dyn EdgePersistence>>,
 }
 
+pub(crate) trait TokenRegister {
+    async fn register_token(&self, token: String) -> EdgeResult<EdgeToken>;
+}
+
+impl TokenRegister for TokenValidator {
+    async fn register_token(&self, token: String) -> EdgeResult<EdgeToken> {
+        Ok(self
+            .register_tokens(vec![token])
+            .await?
+            .first()
+            .expect("Couldn't validate token")
+            .clone())
+    }
+}
+
 impl TokenValidator {
     async fn get_unknown_and_known_tokens(
         &self,
@@ -42,15 +57,6 @@ impl TokenValidator {
             }
             tokens.into_iter().partition(|t| t.token_type.is_none())
         }
-    }
-
-    pub async fn register_token(&self, token: String) -> EdgeResult<EdgeToken> {
-        Ok(self
-            .register_tokens(vec![token])
-            .await?
-            .first()
-            .expect("Couldn't validate token")
-            .clone())
     }
 
     pub async fn register_tokens(&self, tokens: Vec<String>) -> EdgeResult<Vec<EdgeToken>> {
--- a/server/src/middleware/validate_token.rs
+++ b/server/src/middleware/validate_token.rs
@@ -231,9 +231,7 @@ mod tests {
     #[actix_web::test]
     async fn validation_allows_client_tokens_on_backend_paths() {
         let token = EdgeToken {
-            token: "*:development
-                .somesecretstring"
-                .into(),
+            token: "*:development.somesecretstring".into(),
             status: TokenValidationStatus::Validated,
             token_type: Some(TokenType::Client),
             ..Default::default()
--- a/server/src/middleware/validate_token.rs
+++ b/server/src/middleware/validate_token.rs
@@ -1,4 +1,5 @@
-use crate::auth::token_validator::TokenValidator;
+use crate::auth::token_validator::{TokenRegister, TokenValidator};
+use crate::error::EdgeError;
 use crate::types::{EdgeToken, TokenType, TokenValidationStatus};
 use actix_web::{
     HttpResponse,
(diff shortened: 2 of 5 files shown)


server/src/auth/token_validator.rs
```diff
@@ -18,6 +18,21 @@ pub struct TokenValidator {
     pub persistence: Option<Arc<dyn EdgePersistence>>,
 }
 
+pub(crate) trait TokenRegister {
```
sighphyre, line 21: I've carved this method out into its own trait so we can make the validation middleware testable, otherwise its exactly the same logic

server/src/middleware/validate_token.rs
```diff
@@ -19,76 +20,227 @@ pub async fn validate_token(
...
+    }
+}
+
+async fn validate(
+    token: &EdgeToken,
+    maybe_validator: Option<&Data<impl TokenRegister>>,
+    token_cache: &DashMap<String, EdgeToken>,
+    path: &str,
+) -> Result<(), EdgeError> {
+    if token.status == TokenValidationStatus::Trusted {
```
sighphyre, line 41: This is the magic here - pure short circuit if the token is trusted. This function is effectively the same as it was, just decomposed so we can test it

server/src/middleware/validate_token.rs
```diff
@@ -19,76 +20,227 @@ pub async fn validate_token(
         .unwrap()
         .clone()
         .into_inner();
+
+    let validation_status = validate(&token, maybe_validator, &token_cache, req.path()).await;
```
sighphyre, line 24: All of the logic here is delegated out to the private, pure validate function and only so we can test the thing without hundreds of lines of boilerplate

server/src/middleware/validate_token.rs
```diff
@@ -19,76 +20,255 @@ pub async fn validate_token(
...
+fn check_backend_path(path: &str) -> Result<(), EdgeError> {
+    if path.contains("/api/client") {
+        Ok(())
+    } else {
+        Err(EdgeError::Forbidden("".into()))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
sighphyre, line 97: New tests, because nothing was covered here

server/src/middleware/validate_token.rs
```diff
@@ -19,76 +20,255 @@ pub async fn validate_token(
...
+    struct FailValidator {}
+
+    impl TokenRegister for FailValidator {
+        async fn register_token(&self, _token: String) -> crate::types::EdgeResult<EdgeToken> {
+            Err(EdgeError::EdgeTokenError)
+        }
+    }
+
+    #[actix_web::test]
+    async fn validation_always_allows_trusted_tokens() {
```
sighphyre, line 135: Test that covers our new logic

server/src/tokens.rs
```diff
@@ -118,6 +119,15 @@ impl FromRequest for EdgeToken {
             None => "Authorization".to_string(),
         };
         let value = req.headers().get(token_header);
+
+        if let (Some(value), Some(token_cache)) =
```
sighphyre, line 123: The other piece of magic here - this function materializes a token from a request- (including all the app scoped data. If that app scoped data has a token cache and that token cache has this token stored then skip the parsing and return that value

nunogois: Because we only keep tokens that are valid in our token cache, right? I know there's no such thing for these new trusted tokens (once supplied, they're always valid), but I assume that if we have an Unleash token that becomes invalid we remove it from the cache. Which means we can trust our cache: If the token is there, it's valid.

nunogois (approved):
LGTM

server/src/tokens.rs
```diff
@@ -118,6 +119,17 @@ impl FromRequest for EdgeToken {
...
+            (value, req.app_data::<Data<DashMap<String, EdgeToken>>>())
+        {
+            if let Some(token) = token_cache.get(value.to_str().unwrap_or_default()) {
+                if token.status == TokenValidationStatus::Trusted {
+                    return ready(Ok(token.value().clone()));
```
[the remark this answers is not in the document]
sighphyre: Gosh no. This should be fine in practice because it'll result in a cache miss

server/src/tokens.rs
sighphyre: We actually keep around invalid tokens too (i.e. tokens that we've asked Unleash about and Unleash has actively said it's not valid), mostly so we can short circuit the Forbidden instead of trying to re-validate them. In practice I don't think that'd be a problem here since we never query these against Unleash itself, I'm just trying to scope this as narrowly as possible so I can reason about the different validation flows

The logic for ingesting and validating tokens needs a look over because it's way too complex for what it's doing but I don't want to touch that in this PR

nunogois: Got it, makes sense 👍 

Also, I guess there will be no way of revoking a trusted token unless you remove it from your params and restart your Edge instance, but that's probably a compromise we're okay with. Maybe it should be documented as a pro/con of Unleash tokens vs trusted tokens.

sighphyre: Yuh, works like the old trusted tokens on the Proxy. Requires a hard restart and parameter change to revoke them. Feels like an okay compromise

Yessir! This definitely needs some docs

server/src/middleware/validate_token.rs
chriswk: Love it

server/src/middleware/validate_token.rs
chriswk: I'm impressed you got the compiler to be happy here, I struggled so much with getting it to see the correct body type. Happy that you did though

chriswk (approved):
LGTM

server/src/middleware/validate_token.rs
sighphyre: I struggled too, full on wrestling match with the compiler here

PR #903, opened 2025-04-22
feat: allow trusted tokens

sighphyre:
Exposes trusted tokens on the CLI so they can be used. Requires https://github.com/Unleash/unleash-edge/pull/898 to work

### What's accepted

I've chosen to support both real tokens (because users keep asking for them) and the legacy format. The legacy format is a bit weird, you specify both the token and the environment in a single string so

"secret-123@development"

Makes them unambiguous but I can see potential confusion here. I want to tackle this with docs but I'm open to better ideas

### Delimiter

I've chosen @ for a delimiter. It's rejected by our APIs for environment and project and can't appear in the hash so it's safe from that perspective. I think it's safe from a shell perspective. Very open to a different delimiter or even a better plan here

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -173,6 +173,10 @@ pub struct EdgeArgs {
     #[clap(short, long, env, value_delimiter = ',')]
     pub tokens: Vec<String>,
 
+    /// Set a list of frontend tokens that Edge will always trust. These need to either match the Unleash token format, or they're an arbitrary string followed by an @ and then an environment, e.g. secret-123@development
+    #[clap(short, long, env, value_delimiter = ',')]
+    pub pretrusted_tokens: Option<Vec<String>>,
+
     /// Expects curl header format (-H <HEADERNAME>: <HEADERVALUE>)
     /// for instance `-H X-Api-Key: mysecretapikey`
     #[clap(short = 'H', long, env, value_delimiter = ',', value_parser = string_to_header_tuple)]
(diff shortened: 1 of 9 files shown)


server/src/cli.rs
```diff
@@ -173,6 +173,10 @@ pub struct EdgeArgs {
     #[clap(short, long, env, value_delimiter = ',')]
     pub tokens: Vec<String>,
 
+    /// Set a list of frontend tokens that Edge will always trust. These need to either match the Unleash token format, or they're an arbitrary string followed by an @ and then an environment, e.g. secret-123@development
+    #[clap(short, long, env, value_delimiter = ',')]
+    pub pretrusted_tokens: Option<Vec<String>>,
```
chriswk, line 178: Since you have a parser ready, using it through clap, so you either have successful tokens or none at all once you get to your main method?

sighphyre: Yesss! That's a great idea, I'll patch this

chriswk (approved):
Approval here, you might want to use clap and the value_parser macro so pretrusted_tokens is a Vec<EdgeToken> ?

server/src/cli.rs
sighphyre: It does end up as a vec of tuples of Vec<(String, EdgeToken)> but I think that's okay. This means you now specify the token formats correctly or get an error message, which I actually think it better

PR #904, opened 2025-04-22
Docs/pretrusted tokens

sighphyre:


README.md
```diff
@@ -113,6 +113,29 @@ For [frontend tokens](https://docs.getunleash.io/reference/api-tokens-and-client
...
+
+  `*:development.969e810bb1237ef465c42cb5d8c32c47c1377a4c0885ca3671183255`
+
+- A legacy proxy-style token, of the form:
+
+  `some-secret@development`
+
+  In this format, `some-secret` is the token your frontend client will use and `development` is the environment the token applies to.
+
+> ℹ️ **Note**: When using legacy-style tokens (`some-secret@environment`), you must also configure at least one standard token (via `TOKENS`) that covers the specified environment. This is required so that Edge knows which features to fetch for that environment.
```
gastonfournier, line 137: I think this would only work with client token, right?
```suggestion
> ℹ️ **Note**: When using legacy-style tokens (`some-secret@environment`), you must also configure at least one standard client token (via `TOKENS`) that covers the specified environment. This is required so that Edge knows which features to fetch for that environment.
```

gastonfournier (approved):
LG!

PR #919, opened 2025-04-28
feat: frontend client registration

kwasniew:
## About the changes

Frontend SDK registration inside Edge

Background:

We'd like to show connected frontend and backend applications in Unleash. We already support this for direct connections to Unleash (https://github.com/Unleash/unleash/pull/9846) and now we'd like to add edge support for frontend SDKs reporting.
Edge reports only registered backend applications and this PR adds frontend applications registration to Edge so that it can report both
Screenshot 2025-04-28 at 13 02 12

Idea:
* since frontend apps do not register explicitly we can handle frontend app registration on top of the metrics sending
* edge reports bulk metrics together with registered applications. This PR adds frontend applications to backend applications list

Screenshot 2025-04-28 at 13 16 02

Details:
* extended metadata type with SDK type: 'frontend' | 'backend' and setting it for frontend and backend respectively
* on the Unleash side we handle frontend and backend slightly differently (frontend doesn't have events emitting since frontend apps are registered on metrics interval)


server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
+            edge_token.clone(),
+            &connect_via,
+            ClientApplication {
+                app_name: metrics.app_name.clone(),
+                environment: metrics.environment.clone(),
+                instance_id: metrics.instance_id.clone(),
+                connect_via: None,
+                connection_id: None,
+                interval: 15000,
+                started: Utc::now(),
```
kwasniew, line 594: I don't think we use it

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
+        crate::metrics::client_metrics::register_client_application(
+            edge_token.clone(),
+            &connect_via,
+            ClientApplication {
+                app_name: metrics.app_name.clone(),
+                environment: metrics.environment.clone(),
+                instance_id: metrics.instance_id.clone(),
+                connect_via: None,
+                connection_id: None,
+                interval: 15000,
```
kwasniew, line 593: I don't think we use it

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
+                started: Utc::now(),
+                strategies: vec![],
```
kwasniew, line 595: It's not used

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
+                metadata: MetricsMetadata {
+                    sdk_version: Some(version),
```
kwasniew, line 597: we need to also add sdk_type so that Unleash can have lightweight processing for frontend SDKs (same as backend SDK registration but without emitting any events)

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
+                    platform_name: None,
+                    platform_version: None,
+                    yggdrasil_version: None,
+                },
+            },
+            metrics_cache.clone(),
```
kwasniew, line 603: not sure if this can get expensive but I can't find a more memory efficient workaround

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
 )]
 #[post("/client/metrics")]
 async fn post_frontend_metrics(
+    req: HttpRequest,
     edge_token: EdgeToken,
+    connect_via: Data<ConnectVia>,
```
chriswk, line 574: Why are you adding this field? You explicitly set L591 to None, so you don't currently need it.

Alternately, change L591 to `Some(connect_via.0.clone())`

server/src/frontend_api.rs
chriswk: Maybe use this PR as an excuse to drop strategies then. We used to use it to see which custom strategies a backend application had been initialised with. But with frontend, there are no strategies.

server/src/frontend_api.rs
```diff
@@ -568,10 +569,41 @@ security(
...
     metrics: Json<ClientMetrics>,
     metrics_cache: Data<MetricsCache>,
 ) -> EdgeResult<HttpResponse> {
+    if let Some(version) = req
+        .headers()
+        .get("unleash-sdk")
+        .and_then(|val| val.to_str().ok())
+        .map(str::to_owned)
+    {
```
chriswk, line 583: We might want to move this into an extractor, so you can just put it as an argument to the handler.
With an extractor, you make your self a UnleashSdkHeader type, and then write an extractor that explains how to go from a request to an UnleashSdkHeader.

Then you can simply have 
`sdk_version: UnleashSdkHeader` as an argument to the handler method.

server/src/frontend_api.rs
kwasniew: So are you saying that we can fully remove it from the type since it's no longer used?

server/src/frontend_api.rs
kwasniew: register_client_application function requires connect_via reference

server/src/frontend_api.rs
kwasniew: I relaxed backend API constraint to make this one optional but to be backwards compat with older Unleash versions we probably need to keep sending empty strategies for now

server/src/frontend_api.rs
kwasniew: Extracted

chriswk (approved):
:)

PR #964, opened 2025-06-02
feat: Added --edge-auth-header and --upstream-auth-header args

chriswk:
We have customers running behind ServiceMeshes, firewalls, API gateways that occupy the Authorization header with their authorization setup. This PR splits Edge's headers into two
- --edge-auth-header - Which header should be used to authorize access to your Edge instance
- --upstream-auth-header - Which header should Edge use to authorize against the upstream instance (typically Unleash, but in daisy-chain cases, can be another Edge, which should then make sure upstream's `--edge-auth-header` matches downstream's `--upstream-auth-header`)

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -5,6 +5,7 @@ use std::str::FromStr;
 use std::time::Duration;
 
 use crate::error;
+use crate::error::EdgeError;
 use crate::tokens::parse_trusted_token_pair;
 use crate::types::EdgeToken;
 use actix_cors::Cors;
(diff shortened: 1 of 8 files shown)


server/src/cli.rs
```diff
@@ -340,19 +337,62 @@ pub struct InternalBackstageArgs {
     pub disable_instance_data_endpoint: bool,
 }
 
-#[derive(Args, Debug, Clone)]
-pub struct TokenHeader {
-    /// Token header to use for edge authorization.
-    #[clap(long, env, global = true, default_value = "Authorization")]
-    pub token_header: String,
+#[derive(Debug, Clone, Args)]
+pub struct AuthHeaders {
```
sighphyre, line 341: I think it's unfortunate that we now propagate Option<String> throughout the system. This means doing an unwrap_or("Authorization") in several places in the heart of the system.

I'd suggest either mapping the CLI arg value to an internal non optional type like

```
struct InternalAuthHeaders { //yeah that name needs work
  //mirror fields but not optional
}
```

And then parsing to that type early in the startup sequence. That lets us push validation to the edge of the system (haha, see what I did there?)

Alternative a method on the struct itself that does that unwrap_or conversion would be a cheap way of getting something similar without future callers having to worry about resolving the Option.

sighphyre (approved):
I see this got merged faster than I can review. Nothing I'd balk at but I think there are low hanging improvements that can be added here

That being said. My tooling is adamant this requires the Rust 1.86 ABI

PR #1025, opened 2025-07-08
fix: metrics batching

sighphyre:
This fixes the broken metric batching logic that could overshoot size limits. 

## The bug

The old `cut_into_sendable_batches` naively assumes that the metrics are of uniform size in the MetricsBatch and attempts to produce a batch that splits them by count but not by size of the actual payload. When some items are significantly larger than others, this can lead to a batch that's larger then the max senable payload size to Unleash.

## The fix

This is now replaced with a [greedy chunker](https://www.reddit.com/media?url=https%3A%2F%2Fpreview.redd.it%2Fmy-10-y-o-cat-is-obese-and-wont-slim-down-he-hates-diet-v0-g7vdkbzcknfa1.jpg%3Fwidth%3D1080%26crop%3Dsmart%26auto%3Dwebp%26s%3Dc89a06d00a6cbbd3be5541cbcff78513862220d5), which does the following:

1. Initialize an empty current_batch
2.  For each item in the batch:
     Measure the serialized item's size in bytes and store it in a SizedItem
3. For each SizedItem in order:
    Calculate how big the current batch would be if we added the item
    If that size is larger than max batch size
         - Push current_batch to final_batches (if it has any items)
         - Start a new empty batch
    Add the item to current batch
4. After all items:
   If current_batch has any items, push it to final_batches.
5. Return final_batches.

 ## Considerations

Resource usage is important to reason about here since this only applies when the metrics batches are large. This is a cold path though so this is only burst usage

### Memory

The greedy approach means we have to check the size of each item. This is somewhat mitigated by using JSON serialization that doesn't allocate (ByteCounter) and because this is a clone free algorithm. Previously each item was cloned. Both implementations are O(N) but this one uses about half the memory.

### CPU

This is about twice is expensive as previously, although both are O(N). CPU is cheap for Edge, while memory can be a bit heavier under load.

### Limitations

This doesn't fix a potential issue that a packet may still fail if a single entity is too large for Unleash (can in theory happen if you have more than 100KB of variant names). But then the old code didn't handle that either and neither can Unleash itself. Might be worth another pass to sub section those as well but I don't think we've ever seen this in practice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4934,9 +4934,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.46.0"
+version = "1.46.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1140bb80481756a8cbe10541f37433b459c5aa1e727b4c020fbfebdc25bf3ec4"
+checksum = "0cc3a2344dafbe23a245241fe8b09735b521110d30fcefbbd5feb1797ca35d17"
 dependencies = [
  "backtrace",
  "bytes",
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -92,7 +92,7 @@ serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.140"
 serde_qs = { version = "0.15.0", features = ["actix4", "tracing"] }
 shadow-rs = { version = "1.2.0" }
-tokio = { version = "1.46.0", features = [
+tokio = { version = "1.46.1", features = [
     "macros",
     "rt-multi-thread",
     "tracing",


server/src/metrics/client_metrics.rs
```diff
@@ -736,83 +670,6 @@ mod test {
         )
     }
 
-    #[test_case(10, 100, 1; "10 apps 100 toggles. Will not be split")]
-    #[test_case(1, 10000, 27; "1 app 10k toggles, will be split into 27 batches")]
-    #[test_case(1000, 1000, 8; "1000 apps 1000 toggles, will be split into 8 batches")]
-    #[test_case(500, 5000, 16; "500 apps 5000 toggles, will be split into 16 batches")]
-    #[test_case(5000, 1, 20; "5000 apps 1 metric will be split")]
-    fn splits_successfully_into_sendable_chunks(apps: u64, toggles: u64, batch_count: usize) {
```
sighphyre, line 744: This is moved, not deleted

server/src/metrics/metric_batching.rs
```diff
@@ -0,0 +1,328 @@
...
+        // Check that we have no duplicates
+        let applications_sent_count = batches.iter().flat_map(|b| b.applications.clone()).count();
+
+        assert_eq!(applications_sent_count, apps.len());
+
+        let metrics_sent_count = batches.iter().flat_map(|b| b.metrics.clone()).count();
+        assert_eq!(metrics_sent_count, toggles.len());
+    }
+
+    fn execute_test(
```
sighphyre, line 253: Factored out so that prop_test regressions are copy pastable into single tests for study (see last test in the file for an example)

server/src/metrics/metric_batching.rs
```diff
@@ -0,0 +1,328 @@
...
+        || !current_batch.impact_metrics.is_empty()
+    {
+        batches.push(current_batch);
+    }
+
+    batches
+}
+
+#[instrument(skip(batch))]
+pub(crate) fn cut_into_sendable_batches(batch: MetricsBatch) -> Vec<MetricsBatch> {
```
sighphyre, line 130: This is the exposed function that got redesigned. This is a thin wrapper around `partition_batch` so that we can prop test the internals against a much smaller max bound - the prop tests are expensive and a smaller bound means faster execution

thomasheartman (approved):
Cats and prop tests; what's not to love? 😸 But yeah, this seems like a good solve to me. The tradeoff of memory to cpu seems well reasoned, and ... I'm on board 🚢

gastonfournier (approved):
Looks good! We should still try to test it manually before cutting a release, just to be extra careful :) Good job!

PR #1031, opened 2025-07-09
chore: better dx

sighphyre:
Provides a setup for a better build + test loop. This uses  [Cranelift](https://cranelift.dev/) and [LLD](https://lld.llvm.org/) to get a faster compile and skips the build + test for the integration tests. This improves the time for build + test on a hot compile from about 55 seconds to about 8 seconds on my machine. 

This is entirely opt in, so standard `cargo build` will still use the old tool chain and release builds are unaffected.

This doesn't use [Mold](https://github.com/rui314/mold), since that has some iffy support on MacOS/Windows

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -17,6 +17,52 @@ cp hooks/* .git/hooks/
 * We will adhere to [Semantic versioning](https://semver.org/)
 * We use Clippy to ensure that our code follows a consistent format, you can run Clippy using `cargo clippy --fix` once it's installed
 
+### Development Toolchain Installation
+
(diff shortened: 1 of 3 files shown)


gastonfournier (approved):
LGTM! It's a 2 way door so go for it :)

docs/development-guide.md
```diff
@@ -17,6 +17,52 @@ cp hooks/* .git/hooks/
...
+#### Clang + LLD
+
+Linux (Debian/Ubuntu-based):
+```shell
+sudo apt install clang lld
+```
+
+MacOS
+```shell
+xcode-select --install
```
kwasniew, line 41: clang from xcode doesn't install lld by default

PR #1036, opened 2025-07-11
perf: optimize impact metrics

sighphyre:
Related to https://github.com/Unleash/unleash-types-rs/pull/110. Uses the new mergetmut trait to do a swap in place of impact metrics and adds a new `Into` impl that only clones what it needs.

Other than that most of the changes here are just adapting tests to the new BTreeMap datastructure.


server/src/metrics/client_impact_metrics.rs
```diff
@@ -13,11 +13,11 @@ pub struct ImpactMetricsKey {
     pub environment: String,
 }
 
-impl From<ImpactMetricEnv> for ImpactMetricsKey {
-    fn from(value: ImpactMetricEnv) -> Self {
+impl From<&ImpactMetricEnv> for ImpactMetricsKey {
```
sighphyre, line 16: Smoother `From`. This dodges cloning the entire `ImpactMetric` and instead just clones what it needs

server/src/metrics/client_impact_metrics.rs
```diff
@@ -63,7 +63,8 @@ fn reduce_metrics_samples(metric: ImpactMetricEnv) -> ImpactMetricEnv {
         metric.environment.clone(),
     );
 
-    empty_metric.merge(metric)
+    empty_metric.merge(metric);
```
sighphyre, line 66: This now merges in place, so we need to return it on the next line

server/src/metrics/client_impact_metrics.rs
```diff
@@ -107,14 +107,12 @@ impl MetricsCache {
                 if let Some(existing_metric) =
                     aggregated_metrics.get_mut(&reduced_metric.impact_metric.name)
                 {
-                    let merged = existing_metric.clone().merge(reduced_metric);
-                    *existing_metric = merged;
+                    existing_metric.merge(reduced_metric);
```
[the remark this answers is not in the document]
kwasniew: nice

kwasniew (approved):

PR #1044, opened 2025-07-17
refactor: clean up main

sighphyre:
Extracted and refactored Actix app initialization into a dedicated setup_server function. This change removes inline setup logic from main.rs, reduces duplication, and centralizes App configuration. Mostly it makes it saner to read and reduces the 300 line main to a bunch of smaller functions that limit the amount of state sharing.

No functional changes. All runtime behavior, endpoints, and tasks remain identical.

### DIFF
--- a/server/src/metrics/edge_metrics.rs
+++ b/server/src/metrics/edge_metrics.rs
@@ -310,10 +310,10 @@ pub struct EdgeInstanceData {
 }
 
 impl EdgeInstanceData {
-    pub fn new(app_name: &str) -> Self {
+    pub fn new(app_name: &str, identifier: &Ulid) -> Self {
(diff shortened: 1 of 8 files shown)


server/src/metrics/edge_metrics.rs
```diff
@@ -310,10 +310,10 @@ pub struct EdgeInstanceData {
 }
 
 impl EdgeInstanceData {
-    pub fn new(app_name: &str) -> Self {
+    pub fn new(app_name: &str, identifier: &Ulid) -> Self {
```
sighphyre, line 313: Let's us define an instance ID in the main method and pass it in, rather than trying to get it off the struct later on

PR #1045, opened 2025-07-17
fix: shared reqwest client so keep alive timeout is not exceeded

sighphyre:


server/src/builder.rs
```diff
@@ -355,37 +347,14 @@ async fn build_edge(
     ))
 }
 
-pub async fn build_caches_and_refreshers(
```
sighphyre, line 358: This is now manually done in main so that we have a shared path for the match statement here

server/src/builder.rs
```diff
@@ -437,14 +406,27 @@ mod tests {
             consumption: false,
         };
 
+        let client_meta_information = ClientMetaInformation {
```
sighphyre, line 409: Now made once

server/src/http/instance_data.rs
```diff
@@ -27,65 +26,43 @@ pub enum InstanceDataSending {
...
-                            token: token.clone(),
-                            base_path: args.http.base_path.clone(),
-                            registry,
-                        };
-                        InstanceDataSending::SendInstanceData(instance_data_sender)
-                    })
-                    .map(Ok)
-                    .unwrap_or(Ok(InstanceDataSending::SendNothing))
-            }
+            EdgeMode::Edge(edge_args) => edge_args
```
sighphyre, line 34: We now pass in a reqwest Client and a ClientMetaInformation rather than constructing them inside the function.  Other than that this has been left more or less unchanged. It needs a bit of love probably but in the next PR

server/src/main.rs
```diff
@@ -198,15 +200,43 @@ async fn run_server(args: CliArgs) -> EdgeResult<()> {
         &edge_instance_data.clone(),
     );
 
-    let custom_headers = if let EdgeMode::Edge(edge) = &args.mode {
-        edge.custom_client_headers.clone()
-    } else {
-        vec![]
-    };
+    let (edge_info, instance_data_sender) = match &args.mode {
```
sighphyre, line 203: This was previously all done in `build_caches_and_refreshers` which matched on `mode` and chose to construct the appropriate types for Edge vs Offline. Now this is done here so we know if we need to also construct a single shared Client

PR #1047, opened 2025-07-18
feat: add an option for setting upstream keep alive timeout

sighphyre:
Adds an option for setting keep alive timeout for connections between Edge and upstream. Defaults this to Unleash's interval - 15 seconds.

Also does a bunch of surgery to fix tests :cry:

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -233,7 +233,7 @@ pub struct EdgeArgs {
     pub consumption: bool,
 
     /// Sets the keep-alive timeout for connections from Edge to upstream
-    #[clap(long, env, default_value_t = 15, requires = "strict")]
+    #[clap(long, env, default_value_t = 15)]
     pub client_keepalive_timeout: i64,
 
     /// If set to true, it compares features payload with delta payload and logs diff. This flag is for internal testing only. Do not turn this on for production configurations
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -148,7 +148,7 @@ pub enum PromAuth {
     Basic(String, String),
 }
 
-#[derive(Args, Debug, Clone)]
+#[derive(Args, Debug, Clone, Default)]
 #[command(group(
     ArgGroup::new("data-provider")
         .args(["redis_url", "backup_folder", "s3_bucket_name"]),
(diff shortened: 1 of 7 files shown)


server/src/http/unleash_client.rs
```diff
@@ -132,6 +132,36 @@ impl ClientMetaInformation {
...
+    pub skip_ssl_verification: bool,
+    pub client_identity: Option<ClientIdentity>,
+    pub upstream_certificate_file: Option<PathBuf>,
+    pub connect_timeout: Duration,
+    pub socket_timeout: Duration,
+    pub keep_alive_timeout: Duration,
+    pub client_meta_information: ClientMetaInformation,
+}
+
+#[cfg(test)]
```
sighphyre, line 146: This isn't a safe default for prod config so I've gated it behind test config

server/src/http/unleash_client.rs
```diff
@@ -1210,18 +1219,15 @@ mod tests {
...
-            Duration::seconds(5),
-            ClientMetaInformation {
+        let client = new_reqwest_client(HttpClientArgs {
+            client_identity: Some(identity),
+            client_meta_information: ClientMetaInformation {
                 app_name: "test-client".into(),
-                instance_id: "test-pkcs8".into(),
-                connection_id: "test-pkcs8".into(),
+                instance_id: "test-pem".into(),
+                connection_id: "test-pem".into(),
```
nunogois, line 1227: Nice catch!

nunogois (approved):
LG! And great job with the cleanup. I especially like the default pattern.

server/src/cli.rs
```diff
@@ -232,6 +232,10 @@ pub struct EdgeArgs {
     #[clap(long, env, default_value_t = false, requires = "strict", hide = true)]
     pub consumption: bool,
 
+    /// Sets the keep-alive timeout for connections from Edge to upstream
+    #[clap(long, env, default_value_t = 15, requires = "strict")]
```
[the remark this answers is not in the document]
sighphyre: You're right, that's overkill, I'll remove that requirement!

(We are nudging people to strict but this seems like a silly way to do it)

server/src/http/unleash_client.rs
sighphyre: Ooo, wish I could take credit. Co-pilot did this and I thought "huh, it's smarter than we are" so I kept it :laughing:

PR #1060, opened 2025-07-28
chore: add a constructor to token validator

sighphyre:
This does nothing useful, it just adds a constructor function to the token validator so that we can make some invasive internal changes without touching the entire application


chriswk (approved):
Sure. LGTM

PR #1067, opened 2025-07-30
fix: validation error should reflect status code

chriswk:
So, previously we forced all failed TokenValidationErrors to return Bad Data. This PR instead returns the status code that came from the upstream validation.

This means, if upstream is telling us 429, Edge will also tell downstream 429 (Too many requests).

We don't often get these errors, but we some suspicion that a certain customer's 400s is due to this.


sighphyre (approved):
Seems sane, Clippy has opinions here tho

PR #1071, opened 2025-07-31
fix: use types deserializer instead

sighphyre:
Won't build until https://github.com/Unleash/unleash-types-rs/pull/113 is merged

This uses the types deserializer instead of the wrapper deserializers for Context here. The types deserializer now covers all the cases, freeing us from a bunch of code we don't need

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4430,9 +4430,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
LGTM! We've discussed this online and this is how frontend API works which will soften the migration from API to edge. This makes frontend APIs behave the same in both Unleash and edge

PR #1081, opened 2025-08-05
feat: Enable consumption metrics by default

thomasheartman:
Flips the default value of the `consumption` flag to `true` instead of `false`.

In theory, this would be a breaking change, but there was a docstring that said that changes to this feature _may_ not be breaking (which is about as useless as it gets).

Still need to figure out whether this messes with Edge when you don't run strict.

Also not sure whether this makes the feature stable or whether it's still experimental.


chriswk (approved):
LRTM (Looks reasonable to me)

sjaanus (approved):
LG

PR #1086, opened 2025-08-07
feat(1-3968): send edge hosting type (self-hosted/hosted) with instance metrics data

thomasheartman:
Adds edge hosting type to the edge instance data schema. The hosting type can be set via the env var `EDGE_HOSTING`. 

In serializing, we serialize it as either `hosted` or `self-hosted`. Edge will parse any value that isn't `hosted` (regardless of capitalization) as `self-hosted`. If you don't provide a value, it will not set that value in the instance data payload.

By making the property optional, we also let Edge accept incoming edge data schemas that don't have the property set.

Adds basic tests to verify that we send the property if and only if the value is set.

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -6,7 +6,6 @@ use std::time::Duration;
 
 use crate::error;
 use crate::error::EdgeError;
-use crate::metrics::edge_metrics::Hosting;
 use crate::tokens::parse_trusted_token_pair;
 use crate::types::EdgeToken;
 use actix_cors::Cors;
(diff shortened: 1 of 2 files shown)


server/src/cli.rs
```diff
@@ -433,6 +434,10 @@ pub struct CliArgs {
     #[arg(long, hide = true, global = true)]
     pub markdown_help: bool,
 
+    /// Set this to 'hosted' to mark this Edge as an Unleash-hosted Edge instance.
+    #[arg(long, hide = true, env, global = true, value_enum)]
+    pub hosting_strategy: Option<Hosting>,
```
sjaanus, line 439: Read from env variable instead

sjaanus (approved):
The idea looks ok, without knowing all small memory optimization details about rust.

PR #1091, opened 2025-08-11
fix: work around for timeout issue

sighphyre:
Adds a middleware that forcefully terminates connections on bad requests. This is to work around https://github.com/actix/actix-web/issues/3715, which causes Actix to potentially leave bytes hanging on an open connection when hit with a 400 request that contains a body when sent by a pipelining proxy.

This isn't ideal since it always forces new connections so there's a performance hit. For a well behaved Edge setup, the happy path dominates and this shouldn't be a major issue


server/src/frontend_api.rs
```diff
@@ -75,7 +75,7 @@ pub async fn get_proxy_all_features(
         edge_token,
```
sighphyre, line 75: Everything in this file is because a new Clippy dropped

server/src/middleware/bad_request_terminator.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&self, req: ServiceRequest) -> Self::Future {
+        let service = Rc::clone(&self.service);
+
+        Box::pin(async move {
+            let mut res = service.call(req).await?;
+
+            if res.status().is_client_error() {
```
sighphyre, line 69: The interesting part. Everything else here is annoying boilerplate

sastromskis (approved):
LGTM

server/src/middleware/bad_request_terminator.rs
```diff
@@ -0,0 +1,151 @@
...
+                res.headers_mut()
+                    .insert(CONNECTION, HeaderValue::from_static("close"));
+                res.response_mut()
+                    .head_mut()
+                    .set_connection_type(ConnectionType::Close);
```
gastonfournier, line 74: If headers or the response were, for any reason, already set, this would override them. Just confirming it's intended.

gastonfournier (approved):
LG! I just left a minor comment/question

server/src/middleware/bad_request_terminator.rs
sighphyre: Yep, intentional. I really, really, really need the LB to close this connection in this case

PR #1113, opened 2025-08-21
feat: background tasks

sighphyre:
Adds background task scheduling to Edge. Chose to move away from tokio select scheduling and move to actual background tasks. This did require that I port the existing metrics send task from main, which is why this is so big.

Still needs graceful shutdown hooks but I think this is the correct place to stop and merge

### DIFF
--- a/crates/unleash-edge/src/lib.rs
+++ b/crates/unleash-edge/src/lib.rs
@@ -1,7 +1,7 @@
 use crate::edge_builder::build_edge_state;
 use crate::offline_builder::build_offline_app_state;
 use axum::Router;
-use axum::middleware::from_fn_with_state;
+use axum::middleware::{from_fn, from_fn_with_state};
 use axum::routing::get;
 use chrono::Duration;
 use std::env;
@@ -109,7 +109,8 @@ pub async fn configure_server(args: CliArgs) -> EdgeResult<Router> {
                 .layer(from_fn_with_state(
                     app_state.clone(),
                     middleware::consumption::connection_consumption,
-                )),
+                ))
+                .layer(from_fn(middleware::etag::etag_middleware)),
         );
 
     let top_router: Router = Router::new()


crates/unleash-edge-auth/src/token_validator.rs
```diff
@@ -46,6 +47,46 @@ impl TokenRegister for TokenValidator {
     }
 }
 
+pub fn create_revalidation_task(
+    validator: &Arc<TokenValidator>,
+    validation_interval_seconds: u64,
+) -> Pin<Box<dyn Future<Output = ()> + Send>> {
+    let validator_clone = validator.clone();
+    Box::pin(async move {
```
sighphyre, line 55: `schedule_validation_of_known_tokens` becomes this function now. It's no longer part of TokenValidator so we can call it nicely but it's ultimately the same function just wrapped in Box::pin(async move{}) to make it returnable without executing it

crates/unleash-edge-auth/src/token_validator.rs
```diff
@@ -46,6 +47,46 @@ impl TokenRegister for TokenValidator {
...
+            tokio::select! {
+                _ = tokio::time::sleep(sleep_duration) => {
+                    let _ = validator_clone.revalidate_known_tokens().await;
+                }
+            }
+        }
+    })
+}
+
+pub fn create_revalidation_of_startup_tokens_task(
```
sighphyre, line 67: Same thing, schedule_revalidation_of_startup_tokens is now this + Box::pin(async move{})

crates/unleash-edge-metrics/src/send_unleash_metrics.rs
```diff
@@ -0,0 +1,192 @@
+use std::sync::Arc;
```
sighphyre, line 1: This file is lifted directly from main, the only change here is that we no longer send to the legacy endpoint

crates/unleash-edge-persistence/src/lib.rs
```diff
@@ -21,24 +22,21 @@ pub trait EdgePersistence: Send + Sync {
     async fn save_features(&self, features: Vec<(String, ClientFeatures)>) -> EdgeResult<()>;
 }
 
-pub async fn persist_data(
-    persistence: Option<Arc<dyn EdgePersistence>>,
+pub fn create_persist_data_task(
```
sighphyre, line 25: Box:pin(async move{}) and we no longer need to check if we sleep forever

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -210,8 +222,213 @@ pub async fn build_edge(
...
         ),
-        Arc::new(Some(token_validator)),
-        Arc::new(Some(feature_refresher)),
+        Arc::new(token_validator),
+        Arc::new(feature_refresher),
         persistence,
     ))
 }
+
+pub async fn build_edge_state(
```
sighphyre, line 231: This is now the primary entry point into the building. Instead of assembling 8 different Arc'd structs, it now just takes what it needs, assembles the AppState and BackgroundTasks and returns those. Moving the background task assembly here means the caller doesn't need to care about whether or not you have a TokenValidator or a MetricsCache etc

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -210,8 +222,213 @@ pub async fn build_edge(
...
+        feature_refresher.clone(),
+    ));
+
+    tasks.push(create_send_instance_data_task(
+        instance_data_sender.clone(),
+        edge_instance_data.clone(),
+        instances_observed_for_app_context.clone(),
+    ));
+
+    if let Some(url) = edge.prometheus_remote_write_url {
```
sighphyre, line 379: In cases where we may or may not choose to do a background task, now we can just not do that. Previously we had to make sure the background task slept forever, otherwise Edge would exit

crates/unleash-edge-feature-refresh/src/lib.rs
```diff
@@ -463,16 +463,10 @@ impl FeatureRefresher {
...
-            loop {
-                tokio::time::sleep(Duration::from_secs(3600)).await;
-            }
-        } else {
-            loop {
-                tokio::select! {
-                    _ = tokio::time::sleep(Duration::from_secs(5)) => {
-                        self.refresh_features().await;
-                    }
+        loop {
```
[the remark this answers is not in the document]
chriswk: Nice :)

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -210,8 +222,213 @@ pub async fn build_edge(
...
+            feature_cache.clone(),
+        ));
+    } else {
+        info!("No persistence configured, skipping persistence");
+    }
+
+    tasks
+}
+
+fn create_fetch_task(
```
sighphyre, line 402: Broke this out into it's own task. This was split over 3 places, with a busy loop for fetching if you were actually streaming. Now we make one decision on how to fetch features and all the options are mutually exclusive. You can no longer theoretically be both polling and streaming

crates/unleash-edge/src/offline_builder.rs
```diff
@@ -103,3 +113,40 @@ pub fn build_offline(offline_args: OfflineArgs) -> EdgeResult<CacheContainer> {
...
+        create_offline_background_tasks(features_cache, engine_cache, offline_args);
+
+    Ok((app_state, background_tasks))
+}
+
+fn create_offline_background_tasks(
+    features_cache: Arc<FeatureCache>,
+    engine_cache: Arc<DashMap<String, EngineState>>,
+    offline_args: OfflineArgs,
+) -> Vec<Pin<Box<dyn Future<Output = ()> + Send>>> {
```
chriswk, line 146: Same thing here, since you went to the trouble of creating the type alias, why not just
```suggestion
) -> Vec<BackgroundTask> {
```

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -1,24 +1,36 @@
...
+use unleash_edge_http_client::{
+    ClientMetaInformation, HttpClientArgs, UnleashClient, new_reqwest_client,
+};
+use unleash_edge_metrics::axum_prometheus_metrics::PrometheusAxumLayer;
+use unleash_edge_metrics::metrics_pusher::create_prometheus_write_task;
+use unleash_edge_metrics::send_unleash_metrics::create_send_metrics_task;
 use unleash_edge_persistence::file::FilePersister;
 use unleash_edge_persistence::redis::RedisPersister;
-
-#[cfg(feature = "s3-persistence")]
```
nunogois, line 17: Seems like we lost this along the way. Is it intentional?

crates/unleash-edge-auth/src/token_validator.rs
```diff
@@ -46,6 +47,46 @@ impl TokenRegister for TokenValidator {
...
+    validator: &Arc<TokenValidator>,
+    tokens: Vec<String>,
+    refresher: Arc<FeatureRefresher>,
+) -> Pin<Box<dyn Future<Output = ()> + Send>> {
+    let validator = validator.clone();
```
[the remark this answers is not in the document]
sighphyre: Yep, the clone here is necessary! In this case, we're passing in a borrowed Arc<T>, we don't own that and to give this to the async function we need to own it (made necessary by the move of async move). So we clone it, giving us an owned reference. One of the pain points of async Rust really

nunogois (commented):
Cool stuff!

If this were JS land I would say something like: "Can't you create a createBackgroundTask helper where you just pass a callback?". But maybe this is more _rustic_ so I trust your experience.

Thank you for the insightful comments! Really helped navigate the changes.

crates/unleash-edge/src/edge_builder.rs
sighphyre: Nope, that was my bad. It does seem to compile without + no-default-features it but I think it's better to have it in. Fixed now

PR #1118, opened 2025-08-22
feat: shutdown hook

sighphyre:


crates/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -93,48 +159,29 @@ pub fn create_send_instance_data_task(
...
-                            _ => {
-                                warn!("Failed to post instance data due to unknown error {e:?}");
-                            }
-                        }
-                    } else {
-                        debug!("Successfully posted observability metrics.");
-                        errors = 0;
-                        downstream_instance_data.write().await.clear();
-                        our_instance_data.clear_time_windowed_metrics();
+                Err(err) => match err {
```
sighphyre, line 176: This code can now get a nice error which it can use to determine whether or not it should backoff, allow the backoff state to persist or clear the backoff state without having to deal with the http error messages

crates/unleash-edge-metrics/src/send_unleash_metrics.rs
```diff
@@ -62,6 +64,125 @@ fn get_valid_token(token_cache: Arc<TokenCache>) -> Option<EdgeToken> {
...
 
+#[derive(Debug)]
+enum MetricsSendError {
+    NoBackoff(String),
+    Backoff(String),
+    Unauthed(String),
+    Unknown(String),
+}
+
+async fn send_metrics(
```
sighphyre, line 75: Same treatment as above - we split the error handling from the backoff calculations so that we can reuse this as a background task. This has has the consequence that the background task will not correctly batch the metrics if they're too large rather than yeeting them on the floor if the final batch is too large

crates/unleash-edge-metrics/src/send_unleash_metrics.rs
```diff
@@ -62,6 +64,125 @@ fn get_valid_token(token_cache: Arc<TokenCache>) -> Option<EdgeToken> {
...
+
+    Box::pin(async move {
+        let token = get_valid_token(token_cache.clone());
+        let Some(token) = token else {
+            warn!(
+                "No valid token found for final metrics send. Shutting down without flushing metrics."
+            );
+            return;
+        };
+        let envs = get_metrics_by_environment(&metrics_cache);
```
sighphyre, line 175: Any of the metrics sending tasks can fail if they've been batched. We don't really care, we just want to display that list of fails to the user if there are any

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -15,16 +15,20 @@ use unleash_edge_delta::cache_manager::DeltaCacheManager;
 use unleash_edge_feature_cache::FeatureCache;
 use unleash_edge_feature_refresh::{FeatureRefreshConfig, FeatureRefresher};
```
sighphyre, line 16: The logic change in this file is just to assemble the background task list and bubble it to the caller

crates/unleash-edge/src/main.rs
```diff
@@ -1,13 +1,35 @@
...
 use tracing::info;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use unleash_edge::configure_server;
 use unleash_edge_cli::{CliArgs, EdgeMode};
-use unleash_edge_types::EdgeResult;
 use unleash_edge_types::errors::EdgeError;
+use unleash_edge_types::{BackgroundTask, EdgeResult};
+
+async fn shutdown_signal(address: String, shutdown_tasks: Vec<BackgroundTask>) {
```
sighphyre, line 16: The shutdown hook is immediately executed by Axum and when it completes Axum is shutdown. So we're responsible for doing the waiting here. The logic here is just to wait for SIGINT (ctrl+c) or SIGTERM (system requests shutdown). Once we get the appropriate signal we execute the shutdown tasks in parallel (they were previously executed in sequence). The exception to this is the metrics tasks which will be sequential related to one another but parallel related to the other background tasks.

crates/unleash-edge-persistence/src/lib.rs
```diff
@@ -31,14 +40,28 @@ pub fn create_persist_data_task(
...
 
+pub fn create_once_off_persist(
+    persistence: Arc<dyn EdgePersistence>,
+    token_cache: Arc<DashMap<String, EdgeToken>>,
+    features_cache: Arc<FeatureCache>,
+) -> Pin<Box<dyn Future<Output = ()> + Send>> {
+    let token_cache = token_cache.clone();
+    let features_cache = features_cache.clone();
+    let persistence = persistence.clone();
+    Box::pin(async move { persist(persistence, token_cache, features_cache).await })
```
nunogois, line 62: Same as https://github.com/Unleash/unleash-edge/pull/1118/files#r2298513725 — Trying to wrap my head around the clones.

PR #1119, opened 2025-08-22
feat: deferred validation

sighphyre:
Restores deferred validation after the move to Axum. Leverages the new task scheduling to get the job done


chriswk (approved):
Nice proof that new backgrounding is better

PR #1121, opened 2025-08-26
feat: restore streaming

sighphyre:
Restores streaming to a functioning state. This leans on Axums internal stream APIs rather than having to build it up from parts like previously. This means basically converting a channel where we receive updates about features to a stream that's connected to a downstream client

### DIFF
--- a/crates/unleash-edge-client-api/src/streaming.rs
+++ b/crates/unleash-edge-client-api/src/streaming.rs
@@ -21,7 +21,7 @@ async fn setup_streaming(
 ) -> EdgeResult<Sse<impl Stream<Item = Result<Event, axum::Error>>>> {
     let Some(delta_cache_manager) = app_state.delta_cache_manager.as_ref() else {
         return Err(EdgeError::SseError(
-            "No delta cache manager found, streaming will not work.".into(),
+            "No delta cache manager found, streaming will not work. This is likely because Edge was not started in streaming mode.".into(),
         ));
     };
 


crates/unleash-edge-streaming/src/stream_broadcast.rs
```diff
@@ -0,0 +1,194 @@
...
+                Box::pin(async move { should_continue })
+                    as Pin<Box<dyn Future<Output = bool> + Send>>
+            }
+        })
+        .filter_map({
+            let client_data = client_data.clone();
+            let delta_cache_manager = delta_cache_manager.clone();
+            move |broadcast_result| {
+                let client_data = client_data.clone();
+                let delta_cache_manager = delta_cache_manager.clone();
```
nunogois, line 79: Do we really need to do this twice? Before and after the move?

crates/unleash-edge-client-api/src/delta.rs
```diff
@@ -57,7 +57,7 @@ pub async fn get_features_delta(
     match resolve_delta(DeltaResolverArgs {
         edge_token,
         token_cache: app_state.token_cache.clone(),
-        filter_query,
+        filter_query: filter_query.0,
```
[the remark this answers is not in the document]
sighphyre: Yeeee... yeah I think so, good spot

crates/unleash-edge-client-api/src/streaming.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+async fn setup_streaming(
+    State(app_state): State<AppState>,
+    edge_token: EdgeToken,
+    Query(filter_query): Query<FeatureFilters>,
+) -> EdgeResult<Sse<impl Stream<Item = Result<Event, axum::Error>>>> {
+    let Some(refresher) = app_state.feature_refresher.as_ref() else {
+        return Err(EdgeError::SseError(
+            "No feature refresher found, streaming will not work, this is an application error, please report it.".into(),
```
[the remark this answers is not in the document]
sighphyre: Fair, good point

crates/unleash-edge-streaming/src/stream_broadcast.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    let intro_stream = once(Ok(initial_event));
+    let client_data = Arc::new(RwLock::new(ClientData {
+        revision: resolve_last_event_id(delta_cache_manager.clone(), &streaming_query),
+        streaming_query,
+    }));
+
+    let updates_stream = BroadcastStream::new(rx)
+        .take_while({
+            move |broadcast_result| {
```
sighphyre, line 65: I think this should also respect a token that gets invalidated. If your token got yanked then this should terminate

crates/unleash-edge-streaming/src/stream_broadcast.rs
sighphyre: Yep, it looks a bit hairy but yes. There's two moves here, one into the fliter_map and one into the move |broad_cast|

nunogois (approved):
LGTM

PR #1126, opened 2025-08-27
chore: terminate streaming connection if token is invalidated

sighphyre:
Terminates the streaming connection to a given client if that client's token ever becomes invalid or is removed. In practice, this should only ever occur if the upstream token is deleted from Unleash and only after Edge checks again if that token is invalid


nunogois (approved):
Nice

PR #1128, opened 2025-08-28
chore: cleanup refresher

sighphyre:
Splits the responsibilities for FeatureRefresher and DeltaRefresher into two separate structs + impls


chriswk (approved):
Was part of the pair, so tooting my own horn and giving a +1 here :)

Issue #1167, opened 2025-09-18
`unleash-edge` pod stuck in `CrashLoopBackOff` when deployed via Helm (v3.0.6)


melya:
### Describe the bug

I’m experiencing an issue where the unleash-edge pod fails to start and repeatedly goes into a CrashLoopBackOff state when deploying with the Helm chart.

**Helm Chart Details**
```
name: unleash-edge
version: 3.0.6
repository: https://docs.getunleash.io/helm-charts/
```

**values.yaml**
```
unleash-edge:
  replicaCount: 1
  fullnameOverride: unleash
  nameOverride: unleash
  edge:
    upstreamUrl: https://eu.app.unleash-hosted.com/<redacted>/
    logLevel: "trace"
  env:
    - name: LOG_LEVEL
      value: trace
    - name: WORKERS
      value: "1"
    - name: STRICT
      value: "true"
    - name: CORS_ORIGIN
      value: "*"
    - name: TOKENS
      value: "*:development.<redacted-client-token>"
```

### Expected behavior

The unleash-edge pod should initialize and remain in a healthy Running state.

### Logs, error output, etc.

```bash
{"timestamp":"2025-09-18T18:37:59.162248Z","level":"INFO","fields":{"message":"starting 1 workers"},"target":"actix_server::builder"}
{"timestamp":"2025-09-18T18:37:59.163419Z","level":"ERROR","fields":{"message":"Initialization of HTTP service error: ()"},"target":"actix_http::service"}
{"timestamp":"2025-09-18T18:37:59.163471Z","level":"ERROR","fields":{"message":"can not start worker: ()"},"target":"actix_server::worker"}
{"timestamp":"2025-09-18T18:37:59.163526Z","level":"INFO","fields":{"message":"Actix is shutting down. Persisting data"},"target":"unleash_edge"}
```

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

daveleek:
Hello and thank you for reaching out @melya!
Just writing a quick summary here in case others stumble upon this issue from what we wrote on our Slack Community:

This appears to be related to setting CORS_ORIGIN: * in your values.yaml config. Setting this parameter to * means * gets passed to Actix instead of a valid uri, and Actix will then fail to start up. In our Origin handling code in Edge we instead return AllowOrigin::any() in the event none is specified in arguments.

PR #1174, opened 2025-09-22
feat: allow explicit * in CORS origin

nunogois:
https://linear.app/unleash/issue/2-3886/allow-explicit-cors-origin-in-edge

Allows explicit `*` in CORS origin.

This should make CORS configuration more intuitive, as users can now safely specify `*` for origins without crashing on startup.

Includes some scouting.

Related: https://github.com/Unleash/unleash-edge/issues/1167

### DIFF
--- a/crates/unleash-edge-cli/src/lib.rs
+++ b/crates/unleash-edge-cli/src/lib.rs
@@ -545,8 +545,7 @@ impl CorsOptions {
                 let list = origins
                     .iter()
                     .map(|s| s.trim())
-                    .filter_map(|origin| HeaderValue::from_str(origin).ok())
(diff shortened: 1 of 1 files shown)


crates/unleash-edge-cli/src/lib.rs
```diff
@@ -533,53 +533,55 @@ impl CorsOptions {
     pub fn middleware(&self) -> CorsLayer {
         CorsLayer::new()
             .max_age(MaxAge::exact(Duration::from_secs(self.cors_max_age)))
-            .allow_headers(Any)
```
nunogois, line 536: I think this is redundant, since we call `.allow_headers(self.allowed_headers())` a few lines down.

PR #1190, opened 2025-09-30
fix: smooth out metrics load by using a bounded stream rather than concurrent loop

sighphyre:
This moves the metrics sending code to use a bounded parallel stream, rather than a single concurrent loop. This smooths out the metrics load on the Unleash side, which was previously doing its best to absolutely murder the single socket under staggered load. This in turn means very large metrics payloads (tested with 500K entities, ~100MB over a minute) now successfully upload, rather than getting yeeted on the floor. Had to make some small adjustments to add jitter and backoff & retry for network level failures


chriswk (approved):
Yeah, looks good to me.

PR #1196, opened 2025-10-01
fix: now gets first token from startup tokens for use with metrics po…

chriswk:
…sting

Debugging found that we're currently getting the validated token from tokencache, this ends up being a token that is not allowed to post metrics.

This PR sets us up to use the first passed token from startup tokens, now that we know that we're always in strict mode


sighphyre (approved):
LGTM

PR #1201, opened 2025-10-03
docs: add note that connecting SDKs should use EDGE_URL/api

chriswk:


README.md
```diff
@@ -69,17 +87,28 @@ To run Edge in **edge** mode, use the command `edge`. This is built from `HEAD`
...
 ```shell
 docker run -v ./examples:/edge/data -p 3063:3063 -e BOOTSTRAP_FILE=/edge/data/features.json -e TOKENS=<your_client_token_1,your_client_token_2> unleashorg/unleash-edge:<version> offline
 ```
 
+### Connecting SDKs
+
+Once Edge is up and running, your SDKs should connect to EDGE_URL/api.
+
+- With the default configuration that would be http://localhost:3063/api
+
```
chriswk, line 102: This is the real change. the rest is us not having a configured line width. We should look into setting up an .editorconfig for markdown files.

PR #1229, opened 2025-10-20
refactor: test config

sighphyre:
Some minor clean ups to the data flow

- Removes a bunch of behaviors that did nothing except provide seams for test code - test code now has test only methods
- ClientMetaInformation now takes Ulids and not strings for its ids. Those are cast as late as possible to strings where needed. Turns out in prod code these are only ever ulid strings so this makes it clearer what the actual expectation is
- Collapses a bunch of integration tests into main code and prunes out some duplicates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1120,9 +1120,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.49"
+version = "4.5.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4512b90fa68d3a9932cea5184017c5d200f5921df706d45e853537dea51508f"
+checksum = "0c2cfd7bf8a6017ddaa4e32ffe7403d547790db06bd171c1c53926faab501623"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -1139,9 +1139,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.49"
+version = "4.5.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0025e98baa12e766c67ba13ff4695a887a1eba19569aad00a472546795bd6730"
+checksum = "0a4c05b9e80c5ccd3a7ef080ad7b6ba7d6fc00a985b8b157197075677c82c7a0"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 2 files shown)


crates/unleash-edge-feature-refresh/src/lib.rs
```diff
@@ -154,21 +154,6 @@ pub struct FeatureRefresher {
     pub client_meta_information: ClientMetaInformation,
 }
 
-impl Default for FeatureRefresher {
```
sighphyre: This isn't safe to default. Real world usage requires the arcs here to be shared because other things populate the state here. The only place we use that is currently in test code so I'm taking it away before it does become a prod dependency

crates/unleash-edge-feature-refresh/src/lib.rs
```diff
@@ -796,4 +824,530 @@ mod tests {
         assert!(tokens_to_refresh.contains(&etag_but_last_refreshed_ten_seconds_ago));
         assert!(tokens_to_refresh.contains(&no_etag_so_is_due_for_refresh));
     }
+
+    async fn client_api_test_server(
```
sighphyre: All of these tests are just harvested from the integration tests. They're doing basically the same patterns and the only reason I can see that there were broken out was to reduce the LOC here

crates/unleash-edge-http-client/src/lib.rs
```diff
@@ -103,34 +104,15 @@ lazy_static! {
     .unwrap();
 }
 
+#[cfg_attr(test, derive(Default))]
```
sighphyre: Also not safe to default, so I've moved the default to a compiler gate behind test. Gives us test code that can mess with this but not prod code

crates/unleash-edge-http-client/src/lib.rs
```diff
@@ -142,25 +124,7 @@ pub struct HttpClientArgs {
     pub client_meta_information: ClientMetaInformation,
 }
 
-impl Default for HttpClientArgs {
```
sighphyre: You guessed it, not safe to default

crates/unleash-edge-http-client/src/lib.rs
```diff
@@ -351,46 +321,19 @@ impl UnleashClient {
         }
     }
 
-    pub fn from_url(server_url: Url, instance_id_opt: Option<String>) -> Result<Self, EdgeError> {
-        UnleashClient::new(server_url.as_str(), instance_id_opt)
-    }
-
-    pub fn new(server_url: &str, instance_id_opt: Option<String>) -> Result<Self, EdgeError> {
```
sighphyre: Also only used in tests. And this does a lot, which is worrying

crates/unleash-edge-integration-tests/src/delta_refresher.rs
```diff
@@ -1,181 +0,0 @@
-#[cfg(test)]
```
sighphyre: This is a straight up duplicate of the existing test in the main delta refresher code. It's getting dropped

PR #1230, opened 2025-10-20
chore: isolate app state

sighphyre:
Reduces the scope of AppState to only production code. Tests are now able to mock out the data they need at will without going via the builder. First step in making sure Edge cannot construct itself in invalid ways

### DIFF
--- a/crates/unleash-edge-client-api/src/lib.rs
+++ b/crates/unleash-edge-client-api/src/lib.rs
@@ -13,19 +13,6 @@ pub mod metrics;
 pub mod register;
 pub mod streaming;
 
-
-
-
-
-
-
-
-
-
-
-
-
-
 pub fn router_for<S>() -> Router<S>
 where
     S: Clone + Send + Sync + 'static,


chriswk (approved):
Again, this seems reasonable. Painful to clean up, but once in place makes our lives better. Have a :+1:

crates/unleash-edge-appstate/src/lib.rs
```diff
@@ -76,21 +62,17 @@ impl AppStateBuilder {
...
                 instance_id: instance_id.to_string(),
             },
-            instance_data_sending: Arc::new(InstanceDataSending::SendNothing),
             edge_instance_data: Arc::new(EdgeInstanceData::new(
                 app_name,
                 &instance_id,
                 Some(hosting),
             )),
             connected_instances: Arc::new(RwLock::new(Vec::new())),
-            edge_persistence: None,
```
sighphyre: Handled purely through background tasks

crates/unleash-edge-frontend-api/src/querystring_extractor.rs
```diff
@@ -2,22 +2,19 @@ use axum::extract::FromRequestParts;
...
 use serde::de::DeserializeOwned;
 use serde_qs::Config;
-use unleash_edge_appstate::AppState;
 use unleash_edge_types::errors::EdgeError;
 
 /// Wrapper type like QsQuery, but with custom config.
 pub struct QsQueryCfg<T>(pub T);
 
-impl<T> FromRequestParts<AppState> for QsQueryCfg<T>
+impl<S, T> FromRequestParts<S> for QsQueryCfg<T>
```
sighphyre: Similar dance to earlier, this is now just generic over anything Send+Sync so we no longer need AppState here

chriswk (commented):
Keeping my approval, this still looks like a better way to go. the partial state extractors also look neat.

PR #1233, opened 2025-10-22
feat: add enterprise self hosted type

sighphyre:


crates/unleash-edge/src/edge_builder.rs
```diff
@@ -265,6 +271,13 @@ pub async fn build_edge(
     ))
 }
 
+pub fn hosting_type() -> Hosting {
```
sighphyre: The into/map/from_str dance really isn't necessary here. We literally care about a single case here - if the env var is set and it's hosted

crates/unleash-edge-cli/src/lib.rs
```diff
@@ -483,6 +484,18 @@ pub struct CliArgs {
 
     #[clap(flatten)]
     pub otel_config: OpenTelemetryConfig,
+
+    // Internal usage only. Do not set this
+    #[clap(env = "EDGE_HOSTING", value_parser = hosting_type, hide = true)]
+    pub hosting_type: Option<Hosting>,
+}
```
thomasheartman: mfw this is the original implementation in the PR that introduced it, but I was [told to change it](https://github.com/Unleash/unleash-edge/pull/1086#discussion_r2262527786) 💀 

But yea, this is my preference too, and I like the hosting type function you've got under here 🙌🏼

PR #1235, opened 2025-10-23
chore: remove shutdown signal, no longer needed

sighphyre:


crates/unleash-edge-enterprise-integration-tests/Cargo.toml
```diff
@@ -38,8 +37,8 @@ unleash-edge-feature-refresh = { workspace = true }
 unleash-edge-http-client = { workspace = true }
 unleash-edge-types = { workspace = true }
 
-
 [dev-dependencies]
+unleash-edge = { path = "../unleash-edge", version = "20.0.0", features = ["enterprise"]}
```
sighphyre: Very uncool and subtle thing that was burning us without me realizing. This causes the whole production build to enable the enterprise feature without the ability to turn it off

PR #1236, opened 2025-10-23
chore: add docker build and workflow for enterprise edge

sighphyre:


chriswk (approved):
LGTM, but let's double check that the features flag works.

PR #1243, opened 2025-10-27
feat: refresh state in streaming

nunogois:
https://linear.app/unleash/issue/2-3997/support-refresh-state-in-streaming

Implements support for the new refresh state in streaming.

Every attempt at a surgical change required more changes than I ideally wanted, so I fully leaned into refactoring this with the help of Geppetto. Hopefully it's a bit more readable as a side effect of implementing the feature?

Also tried to keep the logs the same for consistency / easier reviewing.

### DIFF
--- a/crates/unleash-edge-feature-refresh/src/delta_refresh.rs
+++ b/crates/unleash-edge-feature-refresh/src/delta_refresh.rs
@@ -91,12 +91,10 @@ async fn handle_sse(
         eventsource_client::SSE::Event(event)
             if event.event_type == "unleash-connected" || event.event_type == "unleash-updated" =>
         {
-            match event.event_type.as_str() {
-                "unleash-connected" => {
-                    debug!("Connected to unleash! Populating my flag cache now.")
-                }
-                "unleash-updated" => debug!("Got an unleash updated event. Updating cache."),
-                _ => unreachable!(),
+            if event.event_type == "unleash-connected" {
+                debug!("Connected to unleash! Populating flag cache now.");
+            } else {
+                debug!("Got an unleash updated event. Updating cache.");
             }
 
             match serde_json::from_str(&event.data) {


crates/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -32,129 +34,207 @@ pub type Environment = String;
...
+        eventsource_client::SSE::Connected(_) => {
+            debug!("SSE Connection established");
+        }
+        eventsource_client::SSE::Comment(_) => {
+            // purposefully left blank.
+        }
+    }
+}
+
+async fn run_stream_task(
```
sighphyre: I think the sleeps here are probably quite dangerous. I don't know what happens to a stream when it backs up but I'd rather not find out the hard way. Either way we can use use a tokio select to give us reactive responses to our futures here, so we can do something like this

``` rust

async fn run_stream_task(
    delta_refresher: Arc<DeltaRefresher>,
    token: EdgeToken,
    streaming_url: String,
    client_meta_information: ClientMetaInformation,
    custom_headers: Vec<(String, String)>,
    mut refresh_state_rx: Receiver<RefreshState>,
) {
    let mut stream: Option<SseStream> = None;

    loop {
        let state = *refresh_state_rx.borrow_and_update();
        if matches!(state, RefreshState::Paused) {
            if refresh_state_rx.changed().await.is_err() {
                info!("Control channel closed; stopping SSE stream task");
                return; // dunno if this is correct for the flow, but we should do something here
            }
            continue;
        }

        if stream.is_none() {
            let s = match build_sse_stream(
                &streaming_url,
                &token,
                &client_meta_information,
                &custom_headers,
            ) {
                Ok(s) => s,
                Err(e) => {
                    warn!(
                        "SSE misconfiguration detected; cannot build stream: {e:?}. Exiting stream task."
                    );
                    return; // dunno, maybe a panic is better here
                }
            };

            stream = Some(s);
            info!(
                "SSE connected (app: {}, instance: {})",
                client_meta_information.app_name, client_meta_information.instance_id
            );
        }

        if let Some(s) = stream.as_mut() {
            // now the select drives forwards both futures, if we get a changed event while we're waiting for the SSE stream
            // then we can immediately react to it without a sleep and fallback to the entry point of the loop - which is to
            // sleep until we're unpaused
            tokio::select! {
                _ = refresh_state_rx.changed() => {
                    continue;
                }

                next = s.next() => {
                    match next {
                        Some(Ok(sse)) => {
                            handle_sse(sse, &delta_refresher, &token).await;
                        }
                        Some(Err(e)) => {
                            info!("SSE stream error: {e:?}; reconnecting immediately");
                            stream = None;
                            continue;
                        }
                        None => {
                            info!("SSE stream ended; reconnecting immediately");
                            stream = None;
                            continue;
                        }
                    }
                }
            }
        }
    }
}

```

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -32,129 +34,207 @@ pub type Environment = String;
...
+    streaming_url: &str,
+    token: &EdgeToken,
+    client_meta_information: &ClientMetaInformation,
+    custom_headers: &[(String, String)],
+) -> anyhow::Result<SseStream> {
+    use anyhow::Context;
+
+    let mut es_client_builder = eventsource_client::ClientBuilder::for_url(streaming_url)
+        .context("Failed to create EventSource client for streaming")?
+        .header("Authorization", &token.token)?
```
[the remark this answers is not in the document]
nunogois: It should be using a logic similar to what we have in the client, right? https://github.com/Unleash/unleash-edge/blob/e0a0963b634e2613d8da0693e3428726fa4894ad/crates/unleash-edge-http-client/src/lib.rs#L376-L387

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -32,129 +35,199 @@ pub type Environment = String;
...
+        eventsource_client::SSE::Event(event)
+            if event.event_type == "unleash-connected" || event.event_type == "unleash-updated" =>
+        {
+            match event.event_type.as_str() {
+                "unleash-connected" => {
+                    debug!("Connected to unleash! Populating my flag cache now.")
+                }
+                "unleash-updated" => debug!("Got an unleash updated event. Updating cache."),
+                _ => unreachable!(),
+            }
```
nunogois: Just to keep the original logs, but I'm also okay with dropping this if you think we don't need it. We can add a single log that is generic for both "connecting" and "updating" if you prefer.

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -32,129 +34,207 @@ pub type Environment = String;
...
-                eventsource_client::ReconnectOptions::reconnect(true)
-                    .retry_initial(true)
-                    .delay(Duration::from_secs(5))
-                    .delay_max(Duration::from_secs(30))
-                    .backoff_factor(2)
-                    .build(),
-            )
-            .build();
+            .as_str()
+            .to_owned();
```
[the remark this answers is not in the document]
nunogois: Oh nice, thanks!

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
nunogois: Thanks, implemented your suggestion.

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -32,129 +34,207 @@ pub type Environment = String;
...
+                }
+                Err(e) => {
+                    warn!("Could not parse features response to internal representation: {e:?}");
+                }
+            }
+        }
+        eventsource_client::SSE::Event(event) if event.event_type == "unleash-updated" => {
+            debug!("Got an unleash updated event. Updating cache.",);
+
+            match serde_json::from_str(&event.data) {
```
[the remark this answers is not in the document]
sighphyre: Agreed, it should be an easier tackle after this is merged

crates/unleash-edge-feature-refresh/src/delta_refresh.rs
sighphyre: Yeah this is totally fine, if you want something slightly more elegant than the match + unreachable, an if statement would probably be my go to here:

``` rust
    if event.event_type == "unleash-connected" {
        debug!("Connected to unleash! Populating flag cache now.");
    } else {
        debug!("Got an unleash updated event. Updating cache.");
    }
```

nunogois: Of course 🤦 in my attempt to write idiomatic Rust I sometimes end up overcomplicating.

PR #1245, opened 2025-10-28
feat: edge license api

sighphyre:
Adds a licensing API for daisy chaining to downstream instances

This responds immediately with validation status based off what Edge currently believes about its own validation status - this is important for resilience reasons - an upstream Edge should be able to do validations without a running Unleash instance

Currently validation status is hardcoded - coming in a future PR


crates/unleash-edge-client-api/src/heartbeat.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+impl FromRef<AppState> for HeartbeatState {
+    fn from_ref(app_state: &AppState) -> Self {
+        let client = app_state
+            .token_validator
+            .as_ref()
+            .as_ref()
+            .map(|v| v.unleash_client.clone())
+            .expect("Not running in Edge mode but enterprise Edge requires this for licensing");
```
sighphyre: Not a huge fan of this, I think this is a hack. But without reworking the app state flow to have a client and/or build the token validator on the fly from a client, there's no other way to get a client right now. Leaving this because I think our app state is already too broad and this is the lesser of two evils but I'm open to thoughts here

crates/unleash-edge-client-api/src/heartbeat.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        HeartbeatState {
+            client,
+            license_state: LicenseStateResponse::Valid, //TODO: get this from either the API call or Backup. This probably needs to be set at the AppState level
```
sighphyre: Needs a future PR to make this work

crates/unleash-edge-enterprise/src/lib.rs
```diff
@@ -1,15 +1,17 @@
 use std::{pin::Pin, sync::Arc};
 use tokio::sync::watch::Sender;
 use tracing::{debug, warn};
+use ulid::Ulid;
 use unleash_edge_http_client::UnleashClient;
 use unleash_edge_types::{RefreshState, enterprise::LicenseStateResponse, tokens::EdgeToken};
 
 async fn send_heartbeat(
```
sighphyre: Sending heartbeats now requires a connection id so we can send the downstream instance's connection id. Probably not worth it for this to have a whole different method from our existing heartbeat method

crates/unleash-edge-client-api/Cargo.toml
```diff
@@ -34,5 +35,6 @@ utoipa = { workspace = true }
 [dev-dependencies]
 axum-test = { workspace = true }
 maplit = { workspace = true }
+reqwest = { workspace = true }
```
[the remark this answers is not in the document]
nunogois: Yep, only imported in tests 👍

crates/unleash-edge-http-client/src/lib.rs
```diff
@@ -634,6 +632,7 @@ impl UnleashClient {
                 )
             })?;
 
+        println!("Received heartbeat response: {:?}", response);
```
nunogois: I assume this was just for debug purposes and should be cleaned up? Otherwise make it a `debug!` maybe?


```suggestion

```

PR #1247, opened 2025-10-28
feat: global refresh state bound to db

sighphyre:
Bit of a WIP but this allows us to mount the current license state from either the db  or the API into a variable that we can use to inform downstream instances of the license state

### DIFF
--- a/crates/unleash-edge-enterprise/src/lib.rs
+++ b/crates/unleash-edge-enterprise/src/lib.rs
@@ -1,9 +1,14 @@
 use std::{pin::Pin, sync::Arc};
 use tokio::sync::watch::Sender;
+use tracing::warn;
 use ulid::Ulid;
 use unleash_edge_http_client::UnleashClient;
 use unleash_edge_persistence::EdgePersistence;
-use unleash_edge_types::{RefreshState, enterprise::ApplicationLicenseState, tokens::EdgeToken};
+use unleash_edge_types::{
+    RefreshState,
+    enterprise::{ApplicationLicenseState, LicenseState},
+    tokens::EdgeToken,
+};
 
 pub fn create_enterprise_heartbeat_task(
     unleash_client: Arc<UnleashClient>,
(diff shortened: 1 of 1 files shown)


crates/unleash-edge-client-api/src/heartbeat.rs
```diff
@@ -65,14 +66,14 @@ async fn heartbeat(
...
-            edge_license_state: app_state.license_state,
+            edge_license_state: *app_state.license_state.read().await,
         }),
     )
 }
 
 #[derive(Clone)]
 pub struct HeartbeatState {
-    license_state: LicenseState,
+    license_state: Arc<RwLock<LicenseState>>,
```
sighphyre: Another very annoying thing here - in theory we shouldn't need the Arc<RwLock wrapping at this point. We should be able to transform to a copy of LicenseState but between axum marking this as non async and tokio being a bit wild with locks I can't find a way to clone this here

...wouldn't be a problem if this was an AtomicU8 :laughing:

crates/unleash-edge/src/edge_builder.rs
```diff
@@ -605,3 +611,47 @@ fn create_stream_task(
...
+            if let Some(p) = persistence {
+                p.load_license_state().await.map_err(|_| {
+                    EdgeError::HeartbeatError(
+                        "Could not load license from either persistence or API".into(),
+                        StatusCode::SERVICE_UNAVAILABLE,
+                    )
+                })
+            } else {
+                Err(EdgeError::HeartbeatError(
+                    "Could not reach unleash API and no cached license found".into(),
```
nunogois:
```suggestion
                    "Could not reach upstream API and no cached license found".into(),
```

crates/unleash-edge-enterprise/src/lib.rs
```diff
@@ -1,60 +1,36 @@
...
 
-async fn send_heartbeat(
-    unleash_client: Arc<UnleashClient>,
-    token: EdgeToken,
-    refresh_state_tx: &Sender<RefreshState>,
-    connection_id: &Ulid,
-) {
-    match unleash_client.send_heartbeat(&token, connection_id).await {
-        Ok(response) => match response {
-            LicenseState::Valid => {
```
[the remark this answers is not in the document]
sighphyre: Kinda sorta not really. Do we care about this output though? Do we need a user to change something in Edge if this happens? I want to say not really

PR #1256, opened 2025-10-30
chore: feature gate cli args

sighphyre:
Streaming + delta are now intentionally enterprise only

Opens a path to gate off more cli args via compiler flag


nunogois (approved):
You shall not pass! 🧙‍♂️

PR #1261, opened 2025-10-31
fix: stop clippy complaining

chriswk:
And since I run a pre-commit hook that runs rustfmt and clippy, I rustified the heartbeat in the integration tests.


sighphyre (approved):
LGTM!

PR #1269, opened 2025-11-04
chore: notify pipeline about enterprise edge build

sastromskis:
Notifying the pipeline about enterprise edge build


sighphyre (approved):
Yes this looks good!

PR #1274, opened 2025-11-05
fix: normalize path now also replaces double/multiple slashes

chriswk:
So if you make `/api/frontend/` as your base URL and your SDK adds /client/metrics to give you `/api/frontend//client/metrics` this PR will now convert that to `/api/frontend/client/metrics`.

Came in as a request from a customer.


gastonfournier (approved):
LGTM!

PR #1276, opened 2025-11-05
task: workflow to replace release plz

chriswk:
We've loved release-plz, but it doesn't work very well with multiple crates due to how cargo has rate limits on publishing and there are some crates in this we don't want to release.

This PR attempts to reproduce some of release-plz's functionality (check for changes since last tag, make PR with a new release, create tag on main if version has been bumped and there are no other changes)


sighphyre (approved):
Chaos. Love it. I'm here for it. Let's see if it works

PR #1279, opened 2025-11-06
fix: make edge able to listen on ipv6 again

### LINKED ISSUE #1275: Listening to IPv6 addresses no longer possible
### Describe the bug

The recently released version 20 broke listening to IPv6 addresses. Before we were able to set `INTERFACE=::` and we could connect to unleash-edge via IPv6 just fine. Now it fails booting with a `AddrParseError(Ipv4)`, because apparently it [explicitly only accepts IPv4 addresses](https://github.com/Unleash/unleash-edge/blob/unleash-edge-v20.0.0/crates/unleash-edge-cli/src/lib.rs#L708).

We're running unleash-edge in an IPv6-only Kubernetes cluster, so this is essential to us.

I would also like to suggest choosing `::` over `0.0.0.0` as the default in general, because it will create a dual stack socket, so you're not limiting it to either IPv4 or IPv6.

1. Set environment variable `INTERFACE=::` or use the corresponding CLI flag
2. Start unleash-edge

### Expected behavior

unleash-edge starts normally and responds to IPv6 connections.

### Logs, error output, etc.

```sh
thread 'main' panicked at crates/unleash-edge-cli/src/lib.rs:709:57:
called `Result::unwrap()` on an `Err` value: AddrParseError(Ipv4)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

20.1.1

### Subscription type

None

### Hosting type

None

_No response_

chriswk:
This also makes us dual stack compatible, if you bind an unspecified ipv6 address ("::", "::1") we will also turn off ipv6 only on windows to ensure we're listening on both

fixes #1275

### DIFF
--- a/crates/oss/unleash-edge-cli/src/lib.rs
+++ b/crates/oss/unleash-edge-cli/src/lib.rs
@@ -597,7 +597,7 @@ pub struct HttpServerArgs {
     /// Which port should this server listen for HTTP traffic on
     #[clap(short, long, env, default_value_t = 3063)]
     pub port: u16,
-    /// Which interface should this server listen for HTTP traffic on. Listens on "::" by default for dual stack
+    /// Which interface should this server listen for HTTP traffic on. Listens on "::" by default for dual-stack
     #[clap(short, long, env, default_value = "::")]
     pub interface: String,
     /// Which base path should this server listen for HTTP traffic on


sighphyre (approved):
Seems good. I think co-pilot may have a point about the error handling though

PR #1287, opened 2025-11-12
chore: give enterprise edge endpoint its own home

sighphyre:
Just a bit of clean up. Since we now have an endpoint that is now enterprise only, I'm moving it into its own crate so future endpoints can all get a nice home together. This doesn't solve the problem of router composition but that sounds like a problem to deal with when we have more than 1 endpoint that's only enterprise


nunogois (approved):
Nice

PR #1308, opened 2025-11-25
chore: remove license notes

nunogois:
https://linear.app/unleash/issue/2-3969/add-new-code-license-to-enterprise-edge

GH license detection didn't like our notes at the top of the licenses, so maybe we should just remove them and rely on the information on the README.

Follow-up to: #1307 

image


gastonfournier (approved):
LOL I was going to ask if it was ok modifying the default licenses :D

PR #1309, opened 2025-11-27
chore: build enterprise artifacts and include them in gh releases

nunogois:
https://linear.app/unleash/issue/2-4046/build-enterprise-artifacts-and-include-them-in-gh-releases

Builds enterprise artifacts and includes them in GH releases.


sighphyre (approved):
Can't promise it'll work but it seems like a safe enough experiment

PR #1317, opened 2025-12-03
task: added packer workflow and packer build using amazon ebs to buil…

chriswk:
and packer build using amazon ebs to build edge binary

This uses a minimal ubuntu image, installs the needed dependencies for building a rust project, then clones the repo from EDGE_VERSION, builds for its architecture, sets up a service unit for the binary produced.

We should check that the ami produced is suitable for deployment.


sastromskis (approved):
LG

PR #1371, opened 2026-01-13
feat: added revision id and last update to prometheus

chriswk:
These metrics will help debug hosted edge. Ensuring that at the very least, the hosted nodes are all in agreement on which revision each of the keys are synced to.


sastromskis (approved):
LG!

PR #1372, opened 2026-01-13
feat: add Opentelemetry tracing support to enterprise

chriswk:
This PR adds a tracing crate, which when the enterprise feature is enabled and OTEL_EXPORTER_OTLP_ENDPOINT is set, will setup opentelemetry for traces, logs and metrics and export it to the OTLP_ENDPOINT. 

Since datadog, sentry, jaeger and quite a few other monitoring tools all have OTLP support this allows us to have platform independent tracing support and leave it up to the user/collector to decide which system to integrate with.

For metrics we still use prometheus, so we won't have much metrics to export yet. A future improvement might be to replace prometheus with opentelemetry metrics. Since our infrastructure for metrics is still focused on Prometheus, it won't add value for us.

### DIFF
--- a/crates/enterprise/unleash-edge-tracing/src/lib.rs
+++ b/crates/enterprise/unleash-edge-tracing/src/lib.rs
@@ -1,8 +1,5 @@
-use opentelemetry::{KeyValue, global};
-use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
-use opentelemetry_otlp::{
-    Compression, LogExporter, MetricExporter, SpanExporter, WithExportConfig, WithHttpConfig,
-};
+use opentelemetry::KeyValue;
+use opentelemetry_otlp::{WithExportConfig, WithHttpConfig};
 use opentelemetry_sdk::Resource;
 use opentelemetry_sdk::logs::SdkLoggerProvider;
 use opentelemetry_sdk::metrics::SdkMeterProvider;
@@ -12,8 +9,7 @@ use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
-use unleash_edge_cli::{CliArgs, LogFormat, OtelExporterProtocol};
-use unleash_edge_types::errors::EdgeError;
+use unleash_edge_cli::{CliArgs, LogFormat};
 use unleash_edge_types::{BackgroundTask, EdgeResult};
 
 #[derive(Debug, Clone)]
@@ -61,18 +57,18 @@ fn resource(app_id: String) -> Resource {
 #[cfg(feature = "enterprise")]
 fn init_otel(
     endpoint: &str,
-    mode: &OtelExporterProtocol,
+    mode: &unleash_edge_cli::OtelExporterProtocol,
     app_id: String,
 ) -> anyhow::Result<(SdkTracerProvider, SdkMeterProvider, SdkLoggerProvider)> {
     let res = resource(app_id);
     // --- Traces ----
     let span_exporter = match mode {
-        OtelExporterProtocol::Http => SpanExporter::builder()
+        unleash_edge_cli::OtelExporterProtocol::Http => opentelemetry_otlp::SpanExporter::builder()
             .with_http()
             .with_endpoint(endpoint)
-            .with_compression(Compression::Gzip)
+            .with_compression(opentelemetry_otlp::Compression::Gzip)
             .build(),
-        OtelExporterProtocol::Grpc => SpanExporter::builder()
+        unleash_edge_cli::OtelExporterProtocol::Grpc => opentelemetry_otlp::SpanExporter::builder()
             .with_tonic()
             .with_endpoint(endpoint)
             .build(),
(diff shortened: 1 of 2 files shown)


crates/oss/unleash-edge-cli/src/lib.rs
```diff
@@ -648,7 +653,9 @@ pub struct DatadogConfig {
 #[derive(Args, Debug, Clone)]
 pub struct OpenTelemetryConfig {
     #[arg(long, env, global = true)]
-    pub otel_collector_url: Option<String>,
+    pub otel_exporter_otlp_endpoint: Option<String>,
+    #[arg(long, env, global = true, default_value = "grpc")]
+    pub otel_exporter_otlp_protocol: String,
```
sighphyre: Agree with Co-Pilot here, I think this really wants to be an enum

crates/enterprise/unleash-edge-tracing/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+) -> anyhow::Result<(SdkTracerProvider, SdkMeterProvider, SdkLoggerProvider)> {
+    let res = resource(app_id);
+    // --- Traces ----
+    let span_exporter = if mode.starts_with("http") {
+        SpanExporter::builder()
+            .with_http()
+            .with_endpoint(endpoint)
+            .with_compression(Compression::Gzip)
+            .build()
+    } else {
```
sighphyre: What happens here is someone specifies a type that isn't http or grpc? An enum would make that case go away correctly imo

crates/enterprise/unleash-edge-tracing/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+        ))
+        .with_attribute(KeyValue::new("service_instance_id", app_id))
+        .build()
+}
+
+fn init_otel(
+    endpoint: &str,
+    mode: &str,
+    app_id: String,
+) -> anyhow::Result<(SdkTracerProvider, SdkMeterProvider, SdkLoggerProvider)> {
```
sighphyre: I also think the tuple return type is weird here. This gets immediately ingested into an OtelHolder. We can just return that, surely?

crates/enterprise/unleash-edge-tracing/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+        simple_logging(args)
+    }
+}
+
+fn simple_logging(args: &CliArgs) -> EdgeResult<OtelHolder> {
+    init_logging(args);
+    Ok(OtelHolder::empty())
+}
+
+fn init_tracing_subscriber(logger_provider: &SdkLoggerProvider, cli_args: &CliArgs) {
```
sighphyre: This definitely wants to return a result. If we want to keep the shape where the holder has a bunch of options, this would let us partially initialize it. Either way, the caller should get to make a decision about what to do here, not this thing failing silently

crates/oss/unleash-edge/src/lib.rs
```diff
@@ -58,19 +58,30 @@ pub type EdgeInfo = (
...
+            )
+        })
+        .clone();
     let instances_observed_for_app_context: Arc<RwLock<Vec<EdgeInstanceData>>> =
         Arc::new(RwLock::new(Vec::new()));
     let metrics_middleware =
         PrometheusAxumLayer::new(&args.app_name.clone(), &app_id.clone().to_string());
 
-    let (app_state, background_tasks, shutdown_tasks) = match &args.mode {
+    let (app_state, background_tasks, mut shutdown_tasks) = match &args.mode {
```
sighphyre: This `mut` is a smell. Background tasks are generated elsewhere so we can keep server setup separate from task handling. Having to mutate this to inject a task we create here means something has gone wrong. I don't see anything in this PR that should prevent the creation of a shutdown task here that fits the existing pattern. Doesn't really seem unreasonable to limit this to Edge mode either, not sure why we'd want this for offline

crates/enterprise/unleash-edge-tracing/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            logger_provider: None,
+        }
+    }
+}
+
+fn log_filter() -> EnvFilter {
+    EnvFilter::try_from_default_env()
+        .or_else(|_| EnvFilter::try_new("info"))
+        .unwrap()
+}
```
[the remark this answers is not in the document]
chriswk: Ooh, I couldn't find a `new(s: &str)` when I originally wrote this, this is copied from the old Edge log configuration. So didn't look if that existed now.

crates/oss/unleash-edge/src/lib.rs
chriswk: fair point.

crates/enterprise/unleash-edge-tracing/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use clap::Parser;
+    use unleash_edge_cli::{CliArgs, LogFormat};
+
+    #[test]
+    fn test_otel_holder_empty() {
```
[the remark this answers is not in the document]
chriswk: Agreed, dropped it.

Issue #1141, opened 2025-09-03
Edge should not retain segments not being used by any feature


chriswk:
### Describe the feature request

I would like Edge to not return segments that aren't being used by any features.

### Background

We had an issue arise where a customer had multiple edge instances running, all synced to the same key with the same etag and same feature count, but varying segment counts. This confused the customer who thought Edge might not be evaluating consistently.

PR #1400, opened 2026-01-29
fix: filter out unnecessary segments when responding to features requests

sighphyre:
This prunes out unnecessary segments when building the response to `api/client/features`. This does two important things

1) It trades off a bit of CPU for network
2) It ensures that segments being sent downstream are controlled by Edge and not by Unleash - this is important to ensure that streaming connections send identical responses to polling connections, since Unleash may choose to send segments differently to streaming clients at different times in the streamed connection

The filter loop here is linear on time based on the number of referenced segments and linear on space for the same. In practice I believe this is probably a small number but bench marking with a 1K feature set, 10 strategies, 10 segments, so a total of 100K referenced segments yields about a 10% reduction in throughput. That data set is approximately 10MB so already an excessive amount. Some exploration suggests that datasets that do actively get pruned often result in better times. I think this is an acceptable tradeoff

If you're asking why an expensive loop like this has such a small impact on the system, heaptrack analysis points to the two large memory allocations present in JSON serialized responses. One of which I think is unavoidable, so this should never be a dominant factor

Edit: This may not be the heart of the bug that we are facing. I'm going to keep this open a while longer while we hunt a bit more but this is probably still worth it for reducing data transfer

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -820,7 +820,7 @@ dependencies = [
  "hyper-util",
  "mime",
  "pretty_assertions",
- "reqwest 0.13.1",
+ "reqwest 0.13.2",
  "reserve-port",
  "rust-multipart-rfc7578_2",
  "serde",
@@ -3469,7 +3469,7 @@ checksum = "441b28c02014ba0e458838548d5fe770fd0238f024d250f5512cf8540ab48b7a"
 dependencies = [
  "prometheus",
  "prost",
- "reqwest 0.13.1",
+ "reqwest 0.13.2",
  "snap",
 ]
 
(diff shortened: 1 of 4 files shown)


chriswk:
So I guess, this https://github.com/Unleash/unleash-edge/issues/1141 was more important than we thought back in Sep

chriswk (commented):
This makes sense to me.

gastonfournier (approved):
This looks good to me. I was trying to find a way of getting rid of the sequentiality of first filtering out features and then segments, but I couldn't find a good way. So 👍

sighphyre:
> This looks good to me. I was trying to find a way of getting rid of the sequentiality of first filtering out features and then segments, but I couldn't find a good way. So 👍

We could precompute a bunch of stuff. Would probably mean dropping support for tag/project/prefix filtering (golly gosh I would like to do that). That would mean the work in etagging becomes free (currently very expensive). But it's a big and scary change

PR #1494, opened 2026-03-16
test: add multi-node streaming contract coverage

gastonfournier:
## About these changes

This PR adds initial integration coverage for downstream streaming on `GET /api/client/streaming`, with a focus on multi-node Edge behavior.

The main contract being exercised is that a client may disconnect from one Edge node and reconnect to another, and the new node may either:
- replay incrementally from `Last-Event-ID`, or
- fall back to a fresh hydration payload

**The new tests cover:**
- staggered node startup with equivalent effective state
- cross-node reconnect where one node can replay and another falls back to hydration
- project-scoped reconnect behavior across nodes
- a stale-node control case to distinguish history differences from genuinely stale state

### DIFF
--- a/crates/oss/unleash-edge-integration-tests/src/streaming_refresher.rs
+++ b/crates/oss/unleash-edge-integration-tests/src/streaming_refresher.rs
@@ -3,7 +3,7 @@ mod tests {
     use axum::{Router, extract::FromRef};
     use axum_test::TestServer;
     use eventsource_stream::Eventsource;
-    use std::{sync::Arc, time::Duration};
+    use std::{collections::BTreeMap, sync::Arc, time::Duration};
     use tokio::time::timeout;
     use tokio_stream::StreamExt as _;
     use unleash_edge_appstate::edge_token_extractor::AuthState;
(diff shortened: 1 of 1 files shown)


crates/oss/unleash-edge-integration-tests/src/streaming_refresher.rs
```diff
@@ -820,4 +948,265 @@ mod tests {
...
+        let token_cache_b = Arc::new(TokenCache::default());
+        let delta_cache_manager_a = Arc::new(DeltaCacheManager::new());
+        let delta_cache_manager_b = Arc::new(DeltaCacheManager::new());
+
+        token_cache_a.insert(token.token.clone(), token.clone());
+        token_cache_b.insert(token.token.clone(), token.clone());
+
+        delta_cache_manager_a.insert_cache(
+            "development",
+            delta_cache_with_state(0, vec![feature("Inigo Montoya", "project-a")]),
```
daveleek: This test might fail eventually if we add some validation on the feature name (which we might never do here)

daveleek:
```suggestion
            delta_cache_with_state(0, vec![feature("Inigo-Montoya", "project-a")]),
```

sighphyre (commented):
Let's talk a bit about this. From reading this over, we've mocked the HTTP layers, the last event id handling, the ordering of the ClientFeatures response and the way deltas are applied. I think that's too much mocking for what these tests are trying to achieve. I want to either go down to unit tests without HTTP and be thorough around the behavior of the DeltaCacheManager, or I want to go up and do proper e2e tests in containers and use Bash/Python to drive the state forward.

PR #1578, opened 2026-05-06
chore: fix expect log to show correct url

thomasheartman:
fixes an issue where we `expect`ed the wrong URL for edge instance data. Should hopefully clear up some confusion if someone saw this and thought the actual url was `/instance-data`.

Did a grep through the repo, couldn't find any other instances that should be fixed.


chriswk (approved):
Strange that it fails to create this, but yeah, the expect should at least be correct. Thanks for the fix :)

PR #1601, opened 2026-05-18
fix: allow backups to correctly hydrate in streaming mode

sighphyre:
Allows backups to hydrate the delta cache on startup. This is necessary so Edge can survive restarts without upstream in streaming mode

### DIFF
--- a/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
+++ b/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
@@ -36,7 +36,7 @@ use unleash_yggdrasil::EngineState;
 
 pub type Environment = String;
 
-const DELTA_CACHE_LIMIT: usize = 100;
+pub const DELTA_CACHE_LIMIT: usize = 100;
 const SDK_TYPE_HEADER: &str = "Unleash-Sdk-Type";
 const SSE_IDLE_TIMEOUT: Duration = Duration::from_secs(90);
 const SSE_IDLE_RECONNECT_DELAY: Duration = Duration::from_secs(5);
--- a/crates/oss/unleash-edge/src/edge_builder.rs
+++ b/crates/oss/unleash-edge/src/edge_builder.rs
@@ -22,10 +22,11 @@ use unleash_edge_auth::token_validator::{
 use unleash_edge_cli::{
     AuthHeaders, EdgeArgs, LogFormat, OtelExporterProtocol, RedisArgs, RedisMode, S3Args,
 };
+use unleash_edge_delta::cache::{DeltaCache, DeltaHydrationEvent};
 use unleash_edge_delta::cache_manager::{DeltaCacheManager, create_terminate_sse_connections_task};
 use unleash_edge_feature_cache::FeatureCache;
 use unleash_edge_feature_refresh::delta_refresh::{
-    DeltaRefresher, start_streaming_delta_background_task,
+    DELTA_CACHE_LIMIT, DeltaRefresher, start_streaming_delta_background_task,
 };
 use unleash_edge_feature_refresh::{
     FeatureRefreshConfig, FeatureRefresher, HydratorType, start_refresh_features_background_task,
@@ -140,8 +141,7 @@ async fn get_data_source(args: &PersistenceArgs) -> Option<Arc<dyn EdgePersisten
 }
 
 async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn EdgePersistence>) {
-    let (token_cache, features_cache, _delta_cache, engine_cache) = cache;
-    // TODO: do we need to hydrate from persistent storage for delta?
+    let (token_cache, features_cache, delta_cache, engine_cache) = cache;
     let tokens = storage.load_tokens().await.unwrap_or_else(|error| {
         warn!("Failed to load tokens from cache {error:?}");
         vec![]
@@ -160,11 +160,23 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
         features_cache.insert(key.clone(), features.clone());
         let mut engine_state = EngineState::default();
 
-        let warnings = engine_state.take_state(UpdateMessage::FullResponse(features));
+        let warnings = engine_state.take_state(UpdateMessage::FullResponse(features.clone()));
         if let Some(warnings) = warnings {
             warn!("Failed to hydrate features for {key:?}: {warnings:?}");
         }
         engine_cache.insert(key.clone(), engine_state);
+
+        delta_cache.insert_cache(
+            key.as_str(),
+            DeltaCache::new(
+                DeltaHydrationEvent {
+                    event_id: 0,
+                    features: features.features.clone(),
+                    segments: features.segments.unwrap_or_else(|| vec![]).clone(),
+                },
+                DELTA_CACHE_LIMIT,
+            ),
+        );
     }
 }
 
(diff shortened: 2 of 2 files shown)


crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -36,7 +36,7 @@ use unleash_yggdrasil::EngineState;
 
 pub type Environment = String;
 
-const DELTA_CACHE_LIMIT: usize = 100;
+pub const DELTA_CACHE_LIMIT: usize = 100;
```
sighphyre, line 39: Don't love this, creates a new dependency across crates for something that should be built into the streaming guarantees. Hard to patch this nicely without re-architecting streaming though and I'm not prepared to do that right now. And it's also a small issue (right now)

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -140,8 +141,7 @@ async fn get_data_source(args: &PersistenceArgs) -> Option<Arc<dyn EdgePersisten
 }
 
 async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn EdgePersistence>) {
-    let (token_cache, features_cache, _delta_cache, engine_cache) = cache;
-    // TODO: do we need to hydrate from persistent storage for delta?
```
sighphyre, line 144: Prescient comment 😅

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -934,6 +946,67 @@ pub async fn resolve_license(
 
 #[cfg(test)]
 mod tests {
```
sighphyre, line 954: I've just broken this out into its own module. The other tests are really just enterprise tests and they're doing a silly dance to try (and failing) to prevent unused imports - they can have their own home with proper cfg gating

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -934,6 +946,67 @@ pub async fn resolve_license(
...
+        let delta_cache = delta_cache
+            .get(&key)
+            .expect("delta cache should be hydrated from persisted features");
+        let hydration_event = delta_cache.get_hydration_event();
+        assert_eq!(hydration_event.features, vec![feature]);
+        assert_eq!(hydration_event.segments, vec![segment]);
+    }
+}
+
+#[cfg(all(test, feature = "enterprise"))]
```
sighphyre, line 1014: Just lifted all the cfg gates to the module level. No need to fiddle with this inside the test modules, it just leads to a horrible mess imo

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -160,11 +160,23 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
...
         if let Some(warnings) = warnings {
             warn!("Failed to hydrate features for {key:?}: {warnings:?}");
         }
         engine_cache.insert(key.clone(), engine_state);
+
+        delta_cache.insert_cache(
+            key.as_str(),
+            DeltaCache::new(
+                DeltaHydrationEvent {
+                    event_id: 0,
```
sighphyre: We can and should just write this to backup. Fixed in coming PRs

crates/oss/unleash-edge/src/edge_builder.rs
chriswk: Nostradamus would be proud :)

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -160,11 +160,23 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
...
+                    features: features.features.clone(),
+                    segments: features.segments.unwrap_or_else(std::vec::Vec::new).clone(),
```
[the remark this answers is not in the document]
chriswk: I think it has a point here, no?

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
chriswk: I think this is fine for now, since we no longer publish the crates, it isn't the end of the world. A rearchitecture might be coming sooner or later, but no problem approving a visibility modifier as long as we're still free from circular deps. :)

crates/oss/unleash-edge/src/edge_builder.rs
chriswk: Nice

crates/oss/unleash-edge/src/edge_builder.rs
chriswk: yeah, agreed. Thanks for bothering enough to fix it.

chriswk (approved):
The Copilot commands are not a deal breaker, but might save us an allocation?

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -160,11 +160,23 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
...
         let mut engine_state = EngineState::default();
 
-        let warnings = engine_state.take_state(UpdateMessage::FullResponse(features));
+        let warnings = engine_state.take_state(UpdateMessage::FullResponse(features.clone()));
         if let Some(warnings) = warnings {
             warn!("Failed to hydrate features for {key:?}: {warnings:?}");
         }
         engine_cache.insert(key.clone(), engine_state);
+
+        delta_cache.insert_cache(
```
gastonfournier: We are also writing to features_cache above (line 160). I wonder if we should just have a delegate function that knows how to store whatever we loaded from the storage into the corresponding caches, otherwise we risk forgetting about one or the other, or the two drifting apart. Maybe a refactor for later. But it's also a bit confusing that we're loading from cache (aka persistent storage?) and then writing that to multiple caches...

sighphyre: Mmm, I think we might have to talk offline here if you have ideas, this is too detailed to do in text. But there's basically only two paths in the application to do these writes. Not sure that justifies an abstraction here. The current abstractions are also... for want of a better word... janky and fixing them requires a scalpel and a chainsaw

crates/oss/unleash-edge/src/edge_builder.rs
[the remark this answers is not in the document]
sighphyre: This comment is insane and wrong. Yes, this is a problem, but this cannot be solved with this approach because this is startup hydration flow and that id doesn't exist yet. The attached comment in the PR states that it's getting fixed in the next revision

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -160,11 +160,23 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
...
+            key.as_str(),
+            DeltaCache::new(
+                DeltaHydrationEvent {
+                    event_id: 0,
+                    features: features.features.clone(),
+                    segments: features.segments.unwrap_or_default(),
+                },
+                DELTA_CACHE_LIMIT,
+            ),
+        );
```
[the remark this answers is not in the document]
sighphyre: I hate this. I hate that the underlying abstractions force this check because it's stupid and it should never happen. Whatever, fine, we can do this and fix it in the future

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -934,6 +946,67 @@ pub async fn resolve_license(
 
 #[cfg(test)]
 mod tests {
+    use crate::edge_builder::build_caches;
+    use std::env::temp_dir;
+    use std::sync::Arc;
+    use ulid::Ulid;
+    use unleash_edge_persistence::EdgePersistence;
+    use unleash_edge_persistence::file::FilePersister;
+    use unleash_types::client_features::{ClientFeature, ClientFeatures, Segment};
```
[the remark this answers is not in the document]
sighphyre: It is used and this build treats warnings as errors. False negative

sighphyre:
> The Copilot commands are not a deal breaker, but might save us an allocation?

24 bytes on the stack during startup flow. Copilot is being insane. Sure, accepted, I guess in 2026, every byte of RAM counts

crates/oss/unleash-edge/src/edge_builder.rs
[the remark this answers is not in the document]
sighphyre: Dangerously wrong

crates/oss/unleash-edge/src/edge_builder.rs
gastonfournier: yes, I was thinking talking offline / online actually, remotely... ok, gather? I'll go find you after lunch

PR #1603, opened 2026-05-19
feat: allow backup to persist and restore streaming ids

sighphyre:
This allows Edge to restart without line of sight to Unleash and still serve api/client/features when in streaming mode. Should also allow streaming clients to get their initial hydration event and continue when Unleash comes online

### DIFF
--- a/crates/enterprise/unleash-edge-delta/Cargo.toml
+++ b/crates/enterprise/unleash-edge-delta/Cargo.toml
@@ -9,6 +9,7 @@ publish = true
 license-file = "../../../LICENSE-ENTERPRISE.md"
 
 [dependencies]
+ahash = { workspace = true }
 dashmap = { workspace = true }
 prometheus = { workspace = true }
 tokio = { workspace = true }
--- a/crates/enterprise/unleash-edge-delta/src/cache.rs
+++ b/crates/enterprise/unleash-edge-delta/src/cache.rs
@@ -47,6 +47,13 @@ impl DeltaCache {
             .any(|e| e.get_event_id() == revision)
     }
 
+    pub fn latest_revision(&self) -> u32 {
+        self.get_events()
+            .last()
+            .map(|e| e.get_event_id())
+            .unwrap_or(self.hydration_event.event_id)
+    }
+
     pub fn add_events(&mut self, events: &[DeltaEvent]) {
         for event in events.iter() {
             self.events.push(event.clone());
--- a/crates/enterprise/unleash-edge-delta/src/cache_manager.rs
+++ b/crates/enterprise/unleash-edge-delta/src/cache_manager.rs
@@ -1,4 +1,5 @@
 use crate::cache::DeltaCache;
+use ahash::HashMap;
 use dashmap::DashMap;
 use prometheus::{IntCounter, IntGauge, register_int_counter, register_int_gauge};
 use std::sync::{Arc, LazyLock};
@@ -87,6 +88,13 @@ impl DeltaCacheManager {
         let _ = self
             .update_sender
             .send(DeltaCacheUpdate::Deletion(env.to_string()));
+    }
+
+    pub fn get_last_event_ids(&self) -> HashMap<String, u32> {
+        self.caches
+            .iter()
+            .map(|entry| (entry.key().clone(), entry.value().latest_revision()))
+            .collect()
     }
 }
 
--- a/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
+++ b/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
@@ -36,7 +36,7 @@ use unleash_yggdrasil::EngineState;
 
 pub type Environment = String;
 
-const DELTA_CACHE_LIMIT: usize = 100;
+pub const DELTA_CACHE_LIMIT: usize = 100;
 const SDK_TYPE_HEADER: &str = "Unleash-Sdk-Type";
 const SSE_IDLE_TIMEOUT: Duration = Duration::from_secs(90);
 const SSE_IDLE_RECONNECT_DELAY: Duration = Duration::from_secs(5);
--- a/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
+++ b/crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
@@ -20,7 +20,6 @@ use unleash_edge_feature_cache::FeatureCache;
 use unleash_edge_feature_filters::FeatureFilterSet;
 use unleash_edge_feature_filters::delta_filters::{DeltaFilterSet, filter_delta_events};
 use unleash_edge_http_client::{ClientMetaInformation, UnleashClient};
-use unleash_edge_persistence::EdgePersistence;
 use unleash_edge_types::errors::{EdgeError, FeatureError};
 use unleash_edge_types::headers::{
     UNLEASH_APPNAME_HEADER, UNLEASH_CLIENT_SPEC_HEADER, UNLEASH_CONNECTION_ID_HEADER,
@@ -352,7 +351,6 @@ pub struct DeltaRefresher {
     pub delta_cache_manager: Arc<DeltaCacheManager>,
     pub engine_cache: Arc<DashMap<String, EngineState>>,
     pub refresh_interval: chrono::Duration,
-    pub persistence: Option<Arc<dyn EdgePersistence>>,
     pub streaming: bool,
     pub client_meta_information: ClientMetaInformation,
     pub edge_instance_data: Arc<EdgeInstanceData>,
@@ -710,7 +708,6 @@ mod tests {
             features_cache: features_cache.clone(),
             engine_cache: engine_cache.clone(),
             refresh_interval: Duration::seconds(6000),
-            persistence: None,
             streaming: false,
             client_meta_information: ClientMetaInformation::test_config(),
             edge_instance_data: Arc::new(edge_instance_data_for_delta_refresher_test()),
@@ -876,7 +873,6 @@ mod tests {
             features_cache,
             engine_cache,
             refresh_interval: Duration::seconds(6000),
-            persistence: None,
             streaming: true,
             client_meta_information: ClientMetaInformation::test_config(),
             edge_instance_data: Arc::new(edge_instance_data_for_delta_refresher_test()),
--- a/crates/oss/unleash-edge-persistence/src/file.rs
+++ b/crates/oss/unleash-edge-persistence/src/file.rs
@@ -53,6 +53,12 @@ impl FilePersister {
         license_path
     }
 
+    pub fn last_event_id_path(&self) -> PathBuf {
+        let mut last_event_id_path = self.storage_path.clone();
+        last_event_id_path.push("unleash_last_event_id.json");
+        last_event_id_path
+    }
+
     pub fn new(storage_path: &Path) -> Self {
         let _ = std::fs::create_dir_all(storage_path);
         FilePersister {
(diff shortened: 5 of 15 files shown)


crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -166,12 +171,14 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
         }
         engine_cache.insert(key.clone(), engine_state);
 
-        if !features.features.is_empty() {
+        if !features.features.is_empty()
+            && let Some(last_event_id) = last_event_ids.get(&key)
+        {
```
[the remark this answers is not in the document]
sighphyre: Bruh. You need to pick a lane. Last PR you said that was a terrible idea.

This is as good a compromise as possible in my opinion

crates/oss/unleash-edge-persistence/src/lib.rs
```diff
@@ -50,6 +52,11 @@ pub fn create_persist_data_task(
                         token_cache.clone(),
                         features_cache.clone()
                     ).await;
+
+                    if let Some(delta_cache_manager) = delta_cache_manager.as_ref()
+                        && let Err(e) = persistence.save_last_event_ids(delta_cache_manager.get_last_event_ids()).await {
+                            warn!("Could not persist last event ids: {e:?}");
```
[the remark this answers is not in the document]
sighphyre: Cheers, this is great feedback, thank you Copilot, fixed in latest patch

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -352,7 +351,6 @@ pub struct DeltaRefresher {
     pub delta_cache_manager: Arc<DeltaCacheManager>,
     pub engine_cache: Arc<DashMap<String, EngineState>>,
     pub refresh_interval: chrono::Duration,
-    pub persistence: Option<Arc<dyn EdgePersistence>>,
     pub streaming: bool,
```
sighphyre, line 354: For whatever reason delta refresh has a dependency on persistence. Which it doesn't use. Need to invert that dependency to do this feature, so that needed me to break this dependency, otherwise cycle 😢

crates/enterprise/unleash-edge-delta/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = true
 license-file = "../../../LICENSE-ENTERPRISE.md"
 
 [dependencies]
+ahash = { workspace = true }
```
sighphyre, line 12: Not really new. Because we have a dependency on ahash elsewhere in this project, it's now more or less impossible to use hashmaps without importing this

crates/oss/unleash-edge-persistence/src/file.rs
```diff
@@ -200,7 +200,7 @@ impl EdgePersistence for FilePersister {
         .map(|_| ())
     }
 
-    async fn save_last_event_ids(&self, event_ids: HashMap<String, u64>) -> EdgeResult<()> {
+    async fn save_last_event_ids(&self, event_ids: HashMap<String, u32>) -> EdgeResult<()> {
```
sighphyre, line 203: For whatever reason the caching uses a 32 bit data type here. Whatever, doesn't really matter. This saves us a weird cast later. Not sure that matters because there's already a bunch of weird casts on this data type

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -180,7 +187,7 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
             );
         } else {
             debug!(
```
sighphyre, line 187: This should be super rare and only really occur if the storage borked. Is debug appropriate here? Do we want to raise this up to warn/info?

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -612,6 +619,7 @@ fn create_shutdown_tasks(
             persistence,
             token_cache.clone(),
             feature_cache,
+            delta_cache_manager.clone(),
         ));
```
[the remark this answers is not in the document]
sighphyre: Yes. I suppose. That being said, this requires you to change polling modes on Edge from streaming to polling and then back again, while allowing persistence to occur on each change. And even then, it only happens on the first time after that flipping and only if Unleash isn't available on startup.

That seems like an edge case I'm willing to let slide. I agree in a perfect world we'd fix that, however, this points to a deeper architectural smell in the delta design that I don't want to tackle in this fix

crates/enterprise/unleash-edge-delta/src/cache.rs
```diff
@@ -47,6 +47,13 @@ impl DeltaCache {
             .any(|e| e.get_event_id() == revision)
     }
 
+    pub fn latest_revision(&self) -> u32 {
+        self.get_events()
+            .last()
+            .map(|e| e.get_event_id())
+            .unwrap_or(self.hydration_event.event_id)
+    }
```
gastonfournier, line 55: Is this per token? I think it doesn't matter, if you send a larger revision id than the one that unleash has for your token, you'll get a hydration event.

When this happens? You can have a development token at revision 123 and a prod token at revision 100 because you mostly made changes in development. Keeping only 123 in the backup should be enough, and the last event should also be fine, because despite having different visible revisions, the ids are global and monotonically increasing.

crates/enterprise/unleash-edge-delta/src/cache_manager.rs
```diff
@@ -88,6 +89,13 @@ impl DeltaCacheManager {
             .update_sender
             .send(DeltaCacheUpdate::Deletion(env.to_string()));
     }
+
+    pub fn get_last_event_ids(&self) -> HashMap<String, u32> {
+        self.caches
+            .iter()
+            .map(|entry| (entry.key().clone(), entry.value().latest_revision()))
+            .collect()
+    }
```
gastonfournier, line 98: Ok, regarding my previous comment, I think each cache is exclusively linked to one environment, and here's where we collect the latest revision id per environment. So, all good

crates/oss/unleash-edge-persistence/src/lib.rs
```diff
@@ -109,6 +123,18 @@ async fn save_features(features_cache: &FeatureCache, persister: &Arc<dyn EdgePe
...
 
+async fn save_last_event_ids(
+    delta_cache_manager: &Arc<DeltaCacheManager>,
+    persister: &Arc<dyn EdgePersistence>,
+) {
+    if let Err(save_error) = persister
+        .save_last_event_ids(delta_cache_manager.get_last_event_ids())
+        .await
+    {
+        warn!("Could not persist last event ids: {save_error:?}");
```
gastonfournier, line 134: When we try loading it, if we failed storing, will it panic and fail starting up?

crates/oss/unleash-edge-persistence/src/s3.rs
```diff
@@ -211,10 +211,10 @@ pub mod s3_persister {
                 .send()
                 .await
                 .map_err(|_| {
-                    EdgeError::PersistenceError("Failed to load last event id".to_string())
+                    EdgeError::PersistenceError("Failed to load last event ids".to_string())
```
gastonfournier, line 214: Should we fail or just return zero?

crates/oss/unleash-edge-persistence/src/s3.rs
```diff
@@ -211,10 +211,10 @@ pub mod s3_persister {
...
                 })?;
             let data = response.body.collect().await.map_err(|_| {
-                EdgeError::PersistenceError("Failed to read last event id data".to_string())
+                EdgeError::PersistenceError("Failed to read last event ids data".to_string())
```
gastonfournier, line 217: Same, cause existing caches will not have it, and future caches can potentially fail to write it

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -155,6 +155,11 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
         token_cache.insert(token.token.clone(), token);
     }
 
+    let last_event_ids = storage.load_last_event_ids().await.unwrap_or_else(|error| {
+        warn!("Failed to load last event ids from cache {error:?}");
+        Default::default()
```
gastonfournier, line 160: Ok, I think we're handling it gracefully

crates/oss/unleash-edge/src/edge_builder.rs
gastonfournier: I think I'd go with a default value for the event id which should give us a hydration

crates/oss/unleash-edge/src/edge_builder.rs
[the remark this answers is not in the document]
gastonfournier: Yeah, I guess, I can't get my mind around this

crates/enterprise/unleash-edge-delta/src/cache.rs
sighphyre: This particular struct is stored by environment. So we call this once per environment. We drop this kind of data structure into the store and load it back

{
   "development": 123,
   "production": 100,
}

Given we're already yeeting all the features at the store per env this doesn't feel like a big extension

crates/enterprise/unleash-edge-delta/src/cache_manager.rs
sighphyre: Yessir!

crates/oss/unleash-edge-persistence/src/lib.rs
sighphyre: Nah, it'll still load but there will be no data to serve so existing hydrated SDKs will get nothing useful and non hydrated SDKs will get nothing

crates/oss/unleash-edge-persistence/src/s3.rs
sighphyre: On this layer I want to report failure here. We should trap that and make a decision around what to do later. Not sure I'm making the correct decision right now though

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -171,7 +176,7 @@ async fn hydrate_from_persistent_storage(cache: CacheContainer, storage: Arc<dyn
                 key.as_str(),
                 DeltaCache::new(
                     DeltaHydrationEvent {
-                        event_id: 0,
+                        event_id: *last_event_ids.get(&key).unwrap_or(&0),
```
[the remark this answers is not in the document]
gastonfournier: From gipity:
> The problem is in the later streaming update path: crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs:390. If a delta cache already exists, the code always calls update_cache(&key, &delta.events). So when upstream sends a full DeltaEvent::Hydration, it is treated like an incremental update.
> 
> That is bad because DeltaCache::update_hydration_event updates the event_id, but explicitly ignores the Hydration payload: crates/enterprise/unleash-edge-delta/src/cache.rs:124. So the cache can end up with:
> - event_id: fresh from upstream hydration
> - features / segments: old persisted snapshot
> 
> The regular features_cache is probably fine, because ClientFeatures::apply_delta does replace features on hydration. The bug is specifically the downstream delta cache used to answer delta/hydration requests. New downstream clients may receive a hydration event built from stale DeltaCache.get_hydration_event().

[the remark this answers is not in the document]
gastonfournier: https://github.com/Unleash/unleash-edge/pull/1607 what'd be the suggested "fix"

gastonfournier (approved):
Unblocking to go hunting the bug on a different stream (pun intended)

PR #1609, opened 2026-05-20
fix: allow streaming to survive mid stream hydrations

sighphyre:


crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -387,25 +387,14 @@ impl DeltaRefresher {
...
-                        segments,
-                    },
-                    DELTA_CACHE_LIMIT,
-                ),
-            );
+            self.delta_cache_manager.insert_cache(&key, delta_cache);
+        }
+
+        if self.delta_cache_manager.get(&key).is_some() {
+            self.delta_cache_manager.update_cache(&key, &delta.events);
```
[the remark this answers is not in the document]
sighphyre: Awesome feedback, didn't even notice that. Thank you

crates/enterprise/unleash-edge-delta/src/cache.rs
```diff
@@ -56,6 +72,9 @@ impl DeltaCache {
 
     pub fn add_events(&mut self, events: &[DeltaEvent]) {
         for event in events.iter() {
+            if matches!(event, DeltaEvent::Hydration { .. }) {
+                self.events.clear();
```
sighphyre, line 76: We're making the assumption that if you're getting a hydration event, everything that came before is now nonsense

crates/enterprise/unleash-edge-delta/src/cache.rs
```diff
@@ -28,6 +28,22 @@ impl DeltaCache {
         cache
     }
 
+    pub fn from_events(events: &[DeltaEvent], max_length: usize) -> Option<Self> {
```
sighphyre, line 31: More or less the same as before, but now the cache module knows how to assemble a cache, instead of the delta refresh having opinions on what a valid cache can be built from

crates/enterprise/unleash-edge-delta/src/cache_manager.rs
```diff
@@ -63,6 +63,10 @@ impl DeltaCacheManager {
         self.caches.get(env).map(|entry| entry.value().clone())
     }
 
+    pub fn contains(&self, env: &str) -> bool {
```
sighphyre, line 66: Previous code would just clone the entire thing when it wanted to check if it existed. That seems insanely wasteful so we can just not do that

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -387,25 +387,14 @@ impl DeltaRefresher {
...
-                ),
-            );
+            self.delta_cache_manager.insert_cache(&key, delta_cache);
+        }
+
+        if self.delta_cache_manager.contains(&key) {
+            self.delta_cache_manager.update_cache(&key, &delta.events);
         } else {
             warn!(
                 "Warning: No hydrationEvent found in delta.events, but cache empty for environment"
```
sighphyre, line 400: I haven't changed this but in the old code this was more or less giving up - no downstream clients can use this. I haven't altered it but this might want to be a very loud error

crates/enterprise/unleash-edge-delta/src/cache.rs
```diff
@@ -121,8 +140,20 @@ impl DeltaCache {
...
+            } => {
+                self.hydration_event = DeltaHydrationEvent {
+                    event_id: *event_id,
+                    features: features.clone(),
+                    segments: {
+                        let mut segments = segments.clone();
+                        segments.sort_by_key(|segment| segment.id);
+                        segments
+                    },
+                };
```
sighphyre, line 156: Old comment definitely wasn't true with the recovery code we've added to Unleash around sending hydrations when it detects things are off. I think this fixes the original issue and introduces fewer foot guns

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -387,25 +387,14 @@ impl DeltaRefresher {
...
-                    },
-                    DELTA_CACHE_LIMIT,
-                ),
-            );
+            self.delta_cache_manager.insert_cache(&key, delta_cache);
+        }
+
+        if self.delta_cache_manager.contains(&key) {
+            self.delta_cache_manager.update_cache(&key, &delta.events);
         } else {
```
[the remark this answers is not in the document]
sighphyre: Fair enough. Low risk but sane change

crates/enterprise/unleash-edge-delta/src/cache.rs
```diff
@@ -28,6 +28,22 @@ impl DeltaCache {
...
+                features,
+                segments,
+            } => Some(DeltaHydrationEvent {
+                event_id: *event_id,
+                features: features.clone(),
+                segments: segments.clone(),
+            }),
+            _ => None,
+        })?;
+        Some(Self::new(hydration_event, max_length))
```
gastonfournier, line 44: If we have multiple events, are we dropping all and only keeping the hydration one?

crates/enterprise/unleash-edge-delta/src/cache.rs
gastonfournier: I was thinking if this code would be clearer if it's inside the update_hydration_event method

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -387,25 +387,16 @@ impl DeltaRefresher {
...
-                        segments,
-                    },
-                    DELTA_CACHE_LIMIT,
-                ),
-            );
+            self.delta_cache_manager.insert_cache(&key, delta_cache);
+        }
+
+        if cache_populated {
+            self.delta_cache_manager.update_cache(&key, &delta.events);
```
gastonfournier, line 397:
```suggestion
            self.delta_cache_manager.insert_cache(&key, delta_cache);
            // are we considering this same as an already populated cache? If so, shouldn't we do:
            cache_populated = true;
        }

        if cache_populated {
            self.delta_cache_manager.update_cache(&key, &delta.events);
```

gastonfournier (approved):
I think it looks nice, just a couple of cosmetic comments

crates/enterprise/unleash-edge-delta/src/cache.rs
sighphyre: For creation yeah we only use the first hydration event in the list. But the function call that calls this also calls update events with the same list so the rest of the events will get played onto the base event directly afterwards.

This is a behavior change, previously we'd drop all other events. In our current implementation it doesn't matter but I think this has fewer footguns

crates/enterprise/unleash-edge-delta/src/cache.rs
sighphyre: Yeah I tried it, it ends up getting buried in a weird and unclear place, so I did this instead. Don't think t here's a nice way to make this clear without a restructure - stabilize first, we can make pretty later

I worry less here since the function chain is one public method and a rest are privates, so it's harder to screw up via the public API

gastonfournier: Yes, I noticed that the function is not public so it's just cosmetics

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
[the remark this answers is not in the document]
sighphyre: D'oh. This suggestion is a bug. Rolled back

PR #1634, opened 2026-05-29
chore: upgrade rustls-webpki

sighphyre:
Closes a few vulnerabilities present in rustls-webpki 0.101.7 by moving to 0.101.13. Requires a bit of surgery in the AWS and event source clients

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,13 @@ ahash = { version = "0.8.12" }
 anyhow = { version = "1.0.102" }
 async-trait = { version = "0.1.89" }
 aws-config = { version = "1.8.17", features = ["behavior-version-latest"] }
-aws-sdk-s3 = { version = "1.134.0", default-features = false, features = ["behavior-version-latest"] }
+aws-sdk-s3 = { version = "1.134.0", default-features = false, features = [
+    "behavior-version-latest",
+    "sigv4a",
+    "http-1x",
+    "default-https-client",
+    "rt-tokio",
+] }
 axum = { version = "0.8.9", features = [
     "macros",
     "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ ahash = { version = "0.8.12" }
 anyhow = { version = "1.0.102" }
 async-trait = { version = "0.1.89" }
 aws-config = { version = "1.8.17", features = ["behavior-version-latest"] }
-aws-sdk-s3 = { version = "1.133.0", features = ["behavior-version-latest"] }
+aws-sdk-s3 = { version = "1.133.0", default-features = false, features = ["behavior-version-latest"] }
 axum = { version = "0.8.9", features = [
     "macros",
     "tokio",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ ahash = { version = "0.8.12" }
 anyhow = { version = "1.0.102" }
 async-trait = { version = "0.1.89" }
 aws-config = { version = "1.8.17", features = ["behavior-version-latest"] }
-aws-sdk-s3 = { version = "1.134.0", features = ["behavior-version-latest"] }
+aws-sdk-s3 = { version = "1.133.0", default-features = false, features = ["behavior-version-latest"] }
```
sighphyre: Turns out we don't actually need rusttls here, we can just disable everything we don't need

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -129,7 +129,12 @@ fn build_sse_stream(
         es_client_builder = es_client_builder.header(key, value)?;
     }
 
-    let client = es_client_builder.reconnect(reconnect_opts()).build();
+    let _ = rustls::crypto::ring::default_provider().install_default();
+    let transport = launchdarkly_sdk_transport::HyperTransport::new_https()
+        .context("Failed to create EventSource HTTPS transport")?;
+    let client = es_client_builder
+        .reconnect(reconnect_opts())
+        .build_with_transport(transport);
```
sighphyre, line 137: New LD client setup wants an explicit setup with transport

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -129,7 +129,12 @@ fn build_sse_stream(
         es_client_builder = es_client_builder.header(key, value)?;
     }
 
-    let client = es_client_builder.reconnect(reconnect_opts()).build();
+    let _ = rustls::crypto::ring::default_provider().install_default();
```
sighphyre, line 132: Don't care if this fails - it only fails if it's already been called earlier and it's cheap to do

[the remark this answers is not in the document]
sighphyre: I think there's some reasonble ways to solve this problem and this isn't one of them. I'm willing to leave this alone and refactor this out in another PR. Right now, what we have here is safe and cheap, I don't want to start layering in LazyLocks to deal with an error message that no one ever cares about

Cargo.toml
```diff
@@ -25,7 +25,7 @@ ahash = { version = "0.8.12" }
 anyhow = { version = "1.0.102" }
 async-trait = { version = "0.1.89" }
 aws-config = { version = "1.8.17", features = ["behavior-version-latest"] }
-aws-sdk-s3 = { version = "1.133.0", features = ["behavior-version-latest"] }
+aws-sdk-s3 = { version = "1.133.0", default-features = false, features = ["behavior-version-latest"] }
```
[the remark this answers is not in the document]
sighphyre: Okay fair, I'll patch

Cargo.toml
```diff
@@ -25,7 +25,13 @@ ahash = { version = "0.8.12" }
 anyhow = { version = "1.0.102" }
 async-trait = { version = "0.1.89" }
 aws-config = { version = "1.8.17", features = ["behavior-version-latest"] }
-aws-sdk-s3 = { version = "1.134.0", features = ["behavior-version-latest"] }
+aws-sdk-s3 = { version = "1.134.0", default-features = false, features = [
+    "behavior-version-latest",
+    "sigv4a",
+    "http-1x",
+    "default-https-client",
+    "rt-tokio",
```
sighphyre, line 33: I think this is the minimal set of features that don't change behavior

gastonfournier (approved):
LG!

chriswk (approved):

PR #1639, opened 2026-06-01
feat: add single client feature endpoint

sighphyre:
Adds an endpoint that resolves a single client feature - this is so that we can match the Unleash API.

This does a few things differently because there's no real path to make them match identically:

- Edge does not include the description of the flag, think that's okay. That's already happening on the features endpoint anyway
- Edge does not respect tag filtering, because Edge does not have tags present, same on features endpoint
- Environment and project filtering are ignored, we're moving towards deprecating that on Unleash, no point in adding them in here. Environment + project are respected via the token anyway
- Name prefix filtering is ignored, in Unleash the prefix filter is set to the flag name so behavior matches

### DIFF
--- a/crates/oss/unleash-edge-client-api/src/features.rs
+++ b/crates/oss/unleash-edge-client-api/src/features.rs
@@ -84,20 +84,16 @@ pub async fn get_feature(
     AuthToken(edge_token): AuthToken,
     Path(feature_name): Path<String>,
 ) -> EdgeJsonResult<ClientFeature> {
+    let filters = FeatureFilters { name_prefix: None };
 
-    let cache_entry = app_state
-        .features_cache
-        .get(&cache_key(&edge_token))
-        .ok_or(EdgeError::ClientCacheError)?;
-
-    let feature = cache_entry
+    let Json(features) = resolve_features(&app_state, edge_token, filters).await?;
+    let feature = features
         .features
-        .iter()
-        .find(|f| f.name == feature_name)
-        .cloned()
+        .into_iter()
+        .find(|feature| feature.name == feature_name)
         .ok_or(EdgeError::FeatureNotFound(feature_name))?;
 
-    Ok(Json(feature.clone()))
+    Ok(Json(feature))
 }
 
 #[instrument(skip(app_state, edge_token, filter_query))]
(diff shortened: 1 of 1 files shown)


crates/oss/unleash-edge-client-api/src/features.rs
```diff
@@ -62,6 +62,44 @@ pub async fn post_features(
...
+    State(app_state): State<FeatureState>,
+    AuthToken(edge_token): AuthToken,
+    Path(feature_name): Path<String>,
+) -> EdgeJsonResult<ClientFeature> {
+
+    let cache_entry = app_state
+        .features_cache
+        .get(&cache_key(&edge_token))
+        .ok_or(EdgeError::ClientCacheError)?;
+
```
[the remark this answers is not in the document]
sighphyre: Cheers, resolved

crates/oss/unleash-edge-client-api/src/features.rs
```diff
@@ -62,6 +62,40 @@ pub async fn post_features(
...
+        ("Authorization" = [])
+    )
+)]
+#[instrument(skip(app_state, edge_token))]
+pub async fn get_feature(
+    State(app_state): State<FeatureState>,
+    AuthToken(edge_token): AuthToken,
+    Path(feature_name): Path<String>,
+) -> EdgeJsonResult<ClientFeature> {
+    let filters = FeatureFilters { name_prefix: None };
```
gastonfournier, line 87: Do we accept filters? It's kind of silly:
1. If the filter doesn't filter out the feature name, the filter is pointless.
2. If the filter does exclude the feature name, it's a YOU problem 🗡️

gastonfournier: Or is it actually I don't know how to read rust (yet) and it's actually defining this as no filter?

gastonfournier (approved):
I think it looks good, and maybe I answered myself

crates/oss/unleash-edge-client-api/src/features.rs
sighphyre: On Unleash, yeah we accept filters, use them and then immediately make them the user's problem if they don't make sense :shrug:. On this, nah, there's no path to pass them via the endpoint, I've chosen to yeet them because there's nothing you can do with them except make yourself sad for no reason. The filter thing you're seeing here is internal - I'm just using it so that we get nice errors if you goof on the API token and so that you can't access a feature your token doesn't allow

PR #1672, opened 2026-06-17
chore: Add more debug logging for S3

krzychukula:
## About the changes

Trying to log more details about S3 error to make debugging easier

## Discussion points

This is generated. I don't know Rust 🙈 

I tried to guide it to work around some of the things I could think of, but this needs a nitpicky review.

### DIFF
--- a/crates/oss/unleash-edge-persistence/src/s3.rs
+++ b/crates/oss/unleash-edge-persistence/src/s3.rs
@@ -144,8 +144,8 @@ pub mod s3_persister {
                 .send()
                 .await
                 .map_err(|e| EdgeError::PersistenceError(format!("Failed to load license state: {e:?}")))?;
-            let data = response.body.collect().await.map_err(|_| {
-                EdgeError::PersistenceError("Failed to read license state data".to_string())
+            let data = response.body.collect().await.map_err(|e| {
+                EdgeError::PersistenceError(format!("Failed to read license state data: {e:?}"))
             })?;
             serde_json::from_slice(&data.to_vec()).map_err(EdgeError::from)
         }
@@ -198,8 +198,8 @@ pub mod s3_persister {
                 .send()
                 .await
                 .map_err(|e| EdgeError::PersistenceError(format!("Failed to load last event ids: {e:?}")))?;
-            let data = response.body.collect().await.map_err(|_| {
-                EdgeError::PersistenceError("Failed to read last event ids data".to_string())
+            let data = response.body.collect().await.map_err(|e| {
+                EdgeError::PersistenceError(format!("Failed to read last event ids data: {e:?}"))
             })?;
             serde_json::from_slice(&data.to_vec()).map_err(EdgeError::from)
         }
--- a/crates/oss/unleash-edge-persistence/src/s3.rs
+++ b/crates/oss/unleash-edge-persistence/src/s3.rs
@@ -11,9 +11,23 @@ pub mod s3_persister {
         self as s3,
         primitives::{ByteStream, SdkBody},
     };
+    use tracing::debug;
     use unleash_edge_types::errors::EdgeError;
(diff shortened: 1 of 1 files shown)


crates/oss/unleash-edge-persistence/src/s3.rs
```diff
@@ -197,10 +223,13 @@ pub mod s3_persister {
...
-                    "Failed to save last event ids: {}",
-                    s3_err.into_service_error()
-                ))),
+                Err(s3_err) => {
+                    debug_s3_error("Failed to save last event ids", &s3_err);
+                    Err(EdgeError::PersistenceError(format!(
+                        "Failed to save last event ids: {}",
+                        s3_err.into_service_error()
+                    )))
+                }
```
sighphyre: I think instead of trying to examine the error and extract bits, we should be able to just do this with each of the error handling calls in this module. From a bit of reading `into_service_error()` is probably discarding a bunch of useful information here, instead we can just lean on the errors underlying debug implementation via `format(err:?)` which should dump the whole chain. I see we're logging that already at warning/info level in the right places so this is probably enough. Needs a bit of testing though to see what the S3 crate actually decides to do here

```suggestion
                Err(s3_err) => {
                    Err(EdgeError::PersistenceError(format!(
                        "Failed to save last event ids: {s3_err:?}",
                    )))
                }
```

crates/oss/unleash-edge-persistence/src/s3.rs
```diff
@@ -11,10 +11,24 @@ pub mod s3_persister {
...
+        if !tracing::enabled!(tracing::Level::DEBUG) {
+            return;
+        }
+        let mut parts = vec![err.to_string()];
+        let mut source = err.source();
+        while let Some(e) = source {
+            parts.push(e.to_string());
+            source = e.source();
+        }
+        debug!("{context}: {}", parts.join(": "));
```
sighphyre: I think this is overkill, I would prefer to just get the errors working correctly. I've provided a different path below. If you really want to go this path though, `debug!(err:?)` should handle the source recursion for you without having to lean on &dyn and poking the internals with a stick

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -90,6 +90,46 @@ impl From<&EdgeArgs> for PersistenceArgs {
     }
 }
 
+const SENSITIVE_QUERY_PARAMS: &[&str] = &[
+    "X-Amz-Security-Token",
+    "X-Amz-Signature",
+    "X-Amz-Credential",
+];
+
+fn redact_url_credentials(url: &str) -> String {
```
sighphyre: I'd prefer to not do this, if possible. I don't think printing the URL tells us much and redacting sensitive info is a surprisingly tough thing to do properly, especially with AWS stuff where they tend to do some pretty custom stuff

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -121,14 +161,27 @@ async fn get_data_source(args: &PersistenceArgs) -> Option<Arc<dyn EdgePersisten
...
     if let Some(s3_args) = args.s3.clone() {
-        let s3_persister = S3Persister::new_from_env(
-            &s3_args
-                .s3_bucket_name
-                .clone()
-                .expect("Clap is confused, there's no bucket name"),
-            s3_args.s3_force_path_style,
-        )
-        .await;
+        let bucket = s3_args
```
sighphyre: I don't actually think this change has much value. This can basically just be read from the env vars that you provide to Edge at startup. I think let's focus on the error information bubbling, which should contain the information we're actually looking for

sighphyre (commented):
I think we can simplify this massively just by just trapping the error debug information in the strings passed to the error type that we bubble

sighphyre (approved):
Awesome!

PR #1674, opened 2026-06-19
Decrease error logging for first run a bit

krzychukula:
It seems that logging this error was confusing for customers - nothing is really wrong on the first run. 

I added error in https://github.com/Unleash/unleash-edge/pull/1672/changes#diff-9ae328e49f01398e440d9ca76601c00c41644c241f2aab403ef3802d9da07785L217

I think it still makes sense to log weird issues, but missing key already is handled differently in `load_features` so it probably should here too. 

But, the other option is to just go back to the previous message.

### DIFF
--- a/crates/oss/unleash-edge-persistence/src/s3.rs
+++ b/crates/oss/unleash-edge-persistence/src/s3.rs
@@ -372,5 +372,20 @@ mod tests {
             let loaded_event_ids = persister.load_last_event_ids().await.unwrap();
             assert_eq!(loaded_event_ids, event_ids);
         }
+
+        #[tokio::test]
+        async fn test_s3_persister_load_last_event_ids_returns_friendly_error_on_first_run() {
+            let (_localstack, persister) = setup_s3_persister().await;
+
+            let result = persister.load_last_event_ids().await;
+
+            let Err(EdgeError::PersistenceError(msg)) = result else {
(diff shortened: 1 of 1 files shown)


krzychukula:
```
#32 0.083 + PKG_CONFIG_SYSROOT_DIR=/ CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc cargo chef cook --target=aarch64-unknown-linux-gnu --release --recipe-path recipe.json
#32 0.167     Updating crates.io index
#32 0.546 error: failed to get `matchers` as a dependency of package `tracing-subscriber v0.3.23`
#32 0.546     ... which satisfies dependency `tracing-subscriber = "^0.3.23"` (locked to 0.3.23) of package `unleash-edge v0.0.1 (/app/crates/oss/unleash-edge)`
#32 0.546 
#32 0.546 Caused by:
#32 0.546   failed to load source for dependency `matchers`
#32 0.546 
#32 0.546 Caused by:
#32 0.546   unable to update registry `crates-io`
#32 0.546 
#32 0.546 Caused by:
#32 0.546   download of ma/tc/matchers failed
```

sighphyre:
Going to close this one for now. Ideally what we'd do here is downgrade this depending on whether or not we're running streaming mode - or probably just not call this function in the first place.

PR #1709, opened 2026-07-15
feat: send token usage downstream

gastonfournier:
## Summary

Propagates Edge-observed SDK token usage to Unleash through the existing client bulk metrics pipeline.

Edge records tokens that successfully pass token validation and includes them in upstream bulk metrics as `seenTokens`. Unleash can then update `seenAt` for SDK tokens that only connect through Edge, which restores token-cleanup visibility for deployments where SDKs do not connect directly to Unleash.

## Design notes

- Uses the existing metrics payload instead of introducing a new endpoint or background channel.
- Sends token secrets in the request body as `seenTokens`. We considered hashing/HMAC, but kept this simple because Edge already sends token secrets in request bodies for validation, and these metrics requests use the same authenticated TLS path.
- Seen tokens are batched and split with the normal metrics batching logic, so payload size is bounded by the existing upstream metrics request limits.
- This depends on Edge upstream metrics delivery. Edge does not expose a general switch to disable upstream client metrics sending; the internal/backstage metrics flags only disable inspection/Prometheus endpoints. As long as Edge is configured to send metrics upstream normally, token last-seen propagation rides along with that path.

## Payload estimation

  For ~2500 tokens:

  Typical token string payload:
  2500 * ~70-80 chars = ~175-200 KB raw string content

  JSON array overhead:

  - quotes around each token: 2 * 2500 = 5 KB
  - commas: ~2.5 KB
  - field name / brackets: negligible, <100 B

  So typical uncompressed JSON for seenTokens is roughly:

  ~185-210 KB

  If projects/environments are unusually long, maybe:

  2500 * ~150 chars ≈ 375 KB + JSON overhead


crates/oss/unleash-edge-metrics/src/send_unleash_metrics.rs
```diff
@@ -163,18 +165,18 @@ async fn send_metrics(
     let mut results = Vec::new();
 
     for (env, batch) in envs {
-        let slices = get_appropriately_sized_env_batches(&metrics_cache, &batch);
+        let tok = startup_tokens
+            .iter()
+            .find(|t| t.environment.as_ref() == Some(&env))
+            .map(|t| t.token.as_str())
+            .expect("Unable to determine token to use for metrics sending");
```
gastonfournier: Fails faster than previous version and uses this token to hash the tokens before sending them

gastonfournier: This changed with the SHA-256 decision: we no longer use the selected upstream token as an HMAC key. The value in keeping this path strict is still that Edge must have an upstream token for the environment it is sending metrics for, but hashing itself is now deterministic SHA-256 over the observed SDK token.

crates/oss/unleash-edge-types/src/metrics/mod.rs
```diff
@@ -345,6 +345,7 @@ pub struct MetricsCache {
     pub applications: DashMap<ApplicationKey, ClientApplication>,
     pub metrics: DashMap<MetricsKey, ClientMetricsEnv>,
     pub impact_metrics: DashMap<ImpactMetricsKey, Vec<ImpactMetricEnv>>,
+    pub seen_tokens: DashMap<String, String>,
 }
```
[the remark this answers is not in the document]
gastonfournier: Agreed. The cache key is now the SHA-256 hash rather than the plaintext token. I kept the historical `seen_tokens` field name to keep the patch smaller, but added comments clarifying that it is `SHA-256 token hash -> environment` and that raw API tokens are not retained there. This also lets downstream `seenTokenHashes` be cached/forwarded without reconstructing the original token.

crates/oss/unleash-edge-metrics/src/client_metrics.rs
```diff
@@ -90,6 +102,8 @@ pub fn get_appropriately_sized_env_batches(
     cache: &MetricsCache,
     batch: &MetricsBatch,
 ) -> Vec<MetricsBatch> {
+    let batch = batch.clone();
```
gastonfournier, line 112: We were cloning anyway so it's safer to do it earlier

crates/oss/unleash-edge-metrics/src/metric_batching.rs
sighphyre: Can we include an item here for tokens? This is probably the most important test in this file, it proves that the batches are segmented in a way that lets us pass Unleash's size validation

crates/oss/unleash-edge-metrics/src/metric_batching.rs
```diff
@@ -234,7 +247,10 @@ mod tests {
             );
         }
         sink_metrics(&cache, &toggles);
-        let batches = get_appropriately_sized_batches(&cache);
```
gastonfournier, line 237: Batches are now different by environment. This might be a good thing, but the best is that we stop needing a different method

crates/oss/unleash-edge-metrics/src/client_metrics.rs
```diff
@@ -212,6 +185,24 @@ pub fn register_application(cache: &MetricsCache, application: ClientApplication
...
 
+pub fn register_seen_token(cache: &MetricsCache, token: &str) {
+    if EdgeToken::from_str(token)
+        .ok()
+        .and_then(|token| token.environment)
+        .is_some()
+    {
+        cache.seen_tokens.insert(token.to_string());
+    } else {
+        tracing::warn!("Ignoring seen token without a parseable environment");
```
gastonfournier, line 201: Same assumption we had: all tokens come with the environment as metadata

crates/oss/unleash-edge-metrics/src/client_metrics.rs
```diff
@@ -53,10 +52,24 @@ pub fn get_metrics_by_environment(cache: &MetricsCache) -> HashMap<String, Metri
...
+        .seen_tokens
+        .iter()
+        .filter_map(|entry| {
+            let token = entry.key();
+            EdgeToken::from_str(token)
+                .ok()
+                .and_then(|edge_token| edge_token.environment)
+                .map(|environment| (environment, token.clone()))
+        })
+        .into_group_map();
```
gastonfournier: We could argue about doing this split by environment on the ingestion side, but this is off the critical path, so It might be better this way.

crates/oss/unleash-edge-metrics/src/send_unleash_metrics.rs
```diff
@@ -163,18 +164,27 @@ async fn send_metrics(
...
+            .map(|t| t.token.as_str())
+        else {
+            let slices = get_appropriately_sized_env_batches(&metrics_cache, &batch);
+            results.extend(slices.into_iter().map(|slice| {
+                Err(MetricsSendError::NoBackoff(format!(
+                    "No upstream token configured for environment {env}. Dropping metrics batch of {} bytes",
+                    size_of_batch(&slice)
+                )))
+            }));
+            continue;
```
sighphyre, line 179: I think this is just noise. There's no way for metrics to get into Edge if there's no matching env token, because they'll get rejected at the API layer

Guess it's not harmful though, might one day catch a screw up

crates/oss/unleash-edge-metrics/src/client_metrics.rs
sighphyre: I think it's fine to do this here, I care more that it's doing similar things to the rest of the function (and it is). That being said, this little mapping dance to extract the environment could be avoided if we just stored a DashSet<EdgeToken> instead of a DashSet<String>. Not a big deal but it'd help understand the intent of the DashSet and concrete types are easier to reason about than strings.

crates/oss/unleash-edge-metrics/src/client_metrics.rs
```diff
@@ -212,6 +185,24 @@ pub fn register_application(cache: &MetricsCache, application: ClientApplication
         .insert(ApplicationKey::from(application.clone()), application);
 }
 
+pub fn register_seen_token(cache: &MetricsCache, token: &str) {
+    if EdgeToken::from_str(token)
+        .ok()
+        .and_then(|token| token.environment)
+        .is_some()
```
sighphyre: This function is called once and at the callsite we have an EdgeToken. Seems a shame to convert it to a str then on the first line here immediately convert it back to an EdgeToken to do some checking. Might be worth checking the middleware that does the parsing of the environment too - I'm not sure environment can actually be None here

sighphyre (approved):
Very nice. Left some thoughts, nothing pressing, if you merge as is I think this is good

crates/oss/unleash-edge-metrics/src/send_unleash_metrics.rs
gastonfournier: Agree, it's probably just noise but harmless IMO, just defensive programming

crates/oss/unleash-edge-metrics/src/client_metrics.rs
gastonfournier: Agree, this is something that was bugging me, I think [06bb06f](https://github.com/Unleash/unleash-edge/pull/1709/commits/06bb06f19ba44f7a3b935e284786dd728eec8458) makes this flow better

PR #1755, opened 2026-07-29
feat: metrics for discarded feature state and failed background refreshes

gastonfournier:
## Summary

This closes the incident observability gap where Edge could keep serving stale in-memory feature state while a detached background refresh task stopped applying new state. The previous behavior only logged discarded feature-state warnings, so the node could remain healthy while its applied revision froze until restart moved the same poisoned state into the synchronous hydration path.

The metric registration lives in `unleash-edge-feature-refresh`, which already owns polling/delta refresh metrics and already depends on `prometheus`. This keeps the dependency graph unchanged while still using the default Prometheus registry gathered by Edge's existing `/metrics` endpoint.

## Metrics

- `edge_feature_state_warnings_total{environment,source}`
  - source values: `hydration`, `delta`, `full`, `offline`
  - increments by the number of yggdrasil warnings returned when feature state compilation discards toggles
- `edge_feature_refresh_errors_total{environment,kind}`
  - kind values: `parse`, `stream`, `fetch`
  - increments when the background delta/streaming refresh path cannot parse events, hits stream errors, or delta fetches fail
- `edge_last_applied_revision_id{environment}`
  - set when Edge successfully observes an applied full/delta/persisted revision id

## Intended Alerts

```promql
increase(edge_feature_state_warnings_total[15m]) > 0
increase(edge_feature_refresh_errors_total[15m]) > 0
```

Focused coverage added:

- `delta_warnings_increment_feature_state_warnings_metric`
- `hydration_warnings_increment_feature_state_warnings_metric`


crates/oss/unleash-edge-feature-refresh/src/feature_state.rs
```diff
@@ -0,0 +1,70 @@
+use prometheus::{IntCounterVec, IntGaugeVec, register_int_counter_vec, register_int_gauge_vec};
```
sighphyre, line 1: Bit of a weird name for this file/ refresh_metrics.rs maybe?

crates/oss/unleash-edge-feature-refresh/src/feature_state.rs
```diff
@@ -0,0 +1,70 @@
...
+static FEATURE_REFRESH_ERRORS: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    match register_int_counter_vec!(
+        "edge_feature_refresh_errors_total",
+        "Total number of background feature refresh errors",
+        &[ENVIRONMENT_LABEL, KIND_LABEL]
+    ) {
+        Ok(counter) => counter,
+        Err(error) => panic!("failed to register edge_feature_refresh_errors_total: {error}"),
+    }
+});
```
sighphyre, line 33: Nit: Error handling is a little weird here imo. I probably wouldn't bother trying to surface the error here, it's a pretty limited set and it's more or less bounded to bad names or duplicate registration. So I kinda would have just done

``` rust
static LAST_APPLIED_REVISION_ID: LazyLock<IntGaugeVec> = LazyLock::new(|| {
    register_int_gauge_vec!(
        "edge_last_applied_revision_id",
        "Last feature revision ID successfully applied by Edge",
        &[ENVIRONMENT_LABEL],
    )
    .expect("failed to register edge_last_applied_revision_id")
});
```

Which is consistent with the other things but if you really want the error message, this is probably more idiomatic Rust

``` rust
static LAST_APPLIED_REVISION_ID: LazyLock<IntGaugeVec> = LazyLock::new(|| {
    register_int_gauge_vec!(
        "edge_last_applied_revision_id",
        "Last feature revision ID successfully applied by Edge",
        &[ENVIRONMENT_LABEL],
    )
    .unwrap_or_else(|e| panic!("failed to register edge_last_applied_revision_id: {e}"))
});
```

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -235,6 +243,10 @@ async fn run_stream_task_with_idle_timeout(
             ) {
                 Ok(s) => s,
                 Err(e) => {
+                    observe_feature_refresh_error(
+                        &token.environment.clone().unwrap_or("*".to_string()),
```
sighphyre, line 247: Trying to get people to be a bit more careful about blind clones, congrats, you're on my list today. 

I realize this is probably copying an existing pattern but I'm trying to get enough people to change behavior so we get a critical mass in this project. 

In this case, you don't need to do this, the receiver already expects a borrow.

Be cool if we could patch all the places in this PR where we're doing this - let's be the change we (I) want to see

```suggestion
                        token.environment.as_deref().unwrap_or(""),
```

sighphyre (approved):
I think the logic is fine, I have some thoughts on structure and hygiene that would make me happy if we could tackle before merging but leaving that decision up to you

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
```diff
@@ -1022,6 +1082,38 @@ mod tests {
         }
     }
 
+    fn feature_with_invalid_single_value_constraint(name: &str) -> ClientFeature {
```
ppicom, line 1085: **question:** What is the reason this constraint is invalid?

crates/oss/unleash-edge-feature-refresh/src/feature_state.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::sync::LazyLock;
+
+const ENVIRONMENT_LABEL: &str = "environment";
+const SOURCE_LABEL: &str = "source";
+const KIND_LABEL: &str = "kind";
+
+pub const HYDRATION_SOURCE: &str = "hydration";
+pub const DELTA_SOURCE: &str = "delta";
+pub const FULL_SOURCE: &str = "full";
+pub const OFFLINE_SOURCE: &str = "offline";
```
ppicom, line 11: **question:** What are the different sources of hydration? Does it make sense to "docstring" these constants?

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -965,9 +972,43 @@ mod tests {
...
+        HYDRATION_SOURCE, feature_state_warnings_total,
+    };
     use unleash_edge_persistence::EdgePersistence;
     use unleash_edge_persistence::file::FilePersister;
-    use unleash_types::client_features::{ClientFeature, ClientFeatures, Segment};
+    use unleash_types::client_features::{
+        ClientFeature, ClientFeatures, Constraint, Operator, Segment, Strategy,
+    };
+
+    fn client_features_with_invalid_single_value_constraint() -> ClientFeatures {
```
ppicom, line 984: **question:** What makes the single value constraint invalid?

ppicom (commented):
Hey there! 👋 I left a few newbie questions about things I haven't grasped.

crates/oss/unleash-edge-feature-refresh/src/refresh_metrics.rs
```diff
@@ -0,0 +1,70 @@
+use prometheus::{IntCounterVec, IntGaugeVec, register_int_counter_vec, register_int_gauge_vec};
```
gastonfournier: Agree: refresh_metrics.rs describes its responsibility better than feature_state.rs

crates/oss/unleash-edge-feature-refresh/src/refresh_metrics.rs
```diff
@@ -0,0 +1,70 @@
...
+static FEATURE_REFRESH_ERRORS: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    match register_int_counter_vec!(
+        "edge_feature_refresh_errors_total",
+        "Total number of background feature refresh errors",
+        &[ENVIRONMENT_LABEL, KIND_LABEL]
+    ) {
+        Ok(counter) => counter,
+        Err(error) => panic!("failed to register edge_feature_refresh_errors_total: {error}"),
+    }
+});
```
gastonfournier: Sure! `.expect("failed to register …")` is simpler and also includes the underlying error's debug representation.

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
gastonfournier: Yes! Thanks! I've also been fighting against unnecessary clones in the past. Thank you for being that change now :)

crates/oss/unleash-edge-feature-refresh/src/delta_refresh.rs
gastonfournier: `NumEq` requires a numeric operand in `value`; this test sets it to `None` (both value and values) deliberately triggering compilation warnings.

crates/oss/unleash-edge-feature-refresh/src/refresh_metrics.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::sync::LazyLock;
+
+const ENVIRONMENT_LABEL: &str = "environment";
+const SOURCE_LABEL: &str = "source";
+const KIND_LABEL: &str = "kind";
+
+pub const HYDRATION_SOURCE: &str = "hydration";
+pub const DELTA_SOURCE: &str = "delta";
+pub const FULL_SOURCE: &str = "full";
+pub const OFFLINE_SOURCE: &str = "offline";
```
gastonfournier: hydration means persisted-state restoration; delta covers delta polling/SSE; full covers full polling responses; offline covers bootstrap loading/reloading. We can add rustdoc to clarify this distinction.

crates/oss/unleash-edge/src/edge_builder.rs
gastonfournier: It's the same NumEq requiring a value and having: 
```
values: None,
value: None,
```

PR #1763, opened 2026-07-30
feat: add SDK flavor to Edge's frontend client metadata

irstavr:
## About the changes

[Now](https://github.com/Unleash/unleash-types-rs/pull/155) that "unleash-types" carry`sdkFlavor`, `sdkFlavorVersion` on `MetricsMetadata`.

We bump to v0.16.2 of "unleash-types" lib to point to latest schema changes.

Backend SDKs are already fine - they send flavor in their registration.

Note: 

1. no flavour -> the fields are simply absent (`skip_serializing_if` on the type), so old clients and old Unleash behave exactly as before (fully backwards-compatible)
2. fix commit: omit :arm64 nothing installs this package anymore - to fix build workflow

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6153,9 +6153,9 @@ dependencies = [
 
 [[package]]
 name = "unleash-types"
-version = "0.16.1"
+version = "0.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75aa5b25bab4826d8d9ec316332f0fa07326cf969d56589a2992ba04aa39793d"
+checksum = "ecf1d95c8d28530377cb1d80e006cb8144f859fc36962268bef88b2526666bfc"
 dependencies = [
  "base64 0.22.1",
  "chrono",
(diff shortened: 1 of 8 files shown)


crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -127,12 +129,31 @@ pub async fn frontend_post_enabled_features(
...
+    if let Some(application) = frontend_application_with_sdk_flavor_headers(
+        &headers,
+        &edge_token,
+        &app_state.connect_via,
+        &metrics.app_name,
+        metrics.instance_id.clone(),
+    ) {
+        // Without this, frontend SDKs never reach the upstream bulk
+        // `applications[]` and never appear in `client_sdk_versions`
+        unleash_edge_metrics::client_metrics::register_application_if_absent(
```
sighphyre: So let's roll back the change in this function. I don't think metrics are ever going to work for frontend SDKs unless we make some biiiig changes to Unleash itself

sighphyre (approved):
Ye

PR #1779, opened 2026-08-06
Spike/scriptable enrichers

sighphyre:
Adds scriptable context enrichers. This is very much a spike and probably should not be merged. If you want to run it, you can do this:

``` sh
cargo build -r

./target/release/unleash-edge edge -u http://localhost:4242 -t ${TOKEN} --context-enricher-script examples/context-enrichers/01-sync-user-id.js --features-refresh-interval-seconds 1
```


sighphyre:
This was never meant to be merged, only seen. It's done its job, going to close

PR #1810, opened 2026-08-21
chore: core node worker enricher

sighphyre:
Implements the core of the Context Enricher system - a small Node worker child process managed by a Rust module. 

This works by spawning a Node process and immediately invoking a JS script inside that worker. The JS script ingests messages in a loop over std in, processes then asynchronously and writes back responses over stdout. The Rust worker manages this whole process, piping in requests over std in and reading the responses back at a later time over std out. Because this work is inherently asynchronous, the protocol works over message ids - e.g. Rust sends job 7, later expecting a response with that same id

This is as small as I can make this without losing the context for what I'm trying to do, all of this code is currently unused so I've marked the relevant unused places with expect(dead_code) for now to get the compiler off my case, when the code is used those expects will fail the lint and CI so we don't forget them

### DIFF
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -10,9 +10,9 @@ pub(crate) struct EnrichmentRequest {
 }
 
 pub(crate) struct EnrichmentResponse {
-    #[expect(dead_code)]
+    #[cfg_attr(not(test), expect(dead_code))]
     pub(crate) id: i64,
-    #[expect(dead_code)]
+    #[cfg_attr(not(test), expect(dead_code))]
     pub(crate) outcome: Result<Context, String>,
 }
 
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -1,0 +1,99 @@
+use serde::{Deserialize, Deserializer};
+use std::collections::HashMap;
+use unleash_types::client_features::Context;
+
+pub(crate) struct EnrichmentRequest {
+    pub(crate) id: i64,
+    pub(crate) context: Context,
+    pub(crate) headers: HashMap<String, String>,
+}
+
+pub(crate) struct EnrichmentResponse {
+    pub(crate) id: i64,
+    pub(crate) outcome: Result<Context, String>,
+}
+
+pub(crate) enum StateMessage {
+    Ready,
+    Error(String),
+    Closing,
+}
+
+#[derive(Deserialize)]
+struct ProtocolEnrichmentResponse {
+    pub(crate) id: i64,
+    pub(crate) context: Option<Context>,
+    pub(crate) error: Option<String>,
+}
+
+impl<'de> Deserialize<'de> for EnrichmentResponse {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // yeah life sucks here - this is borked on the wire, but importantly this is not
+        // a worker telling us about an error state, this is a worker who's unable to communicate with
+        // us effectively for some reason - most likely a bug in the protocol itself. Either way, this is
+        // a different type of error from a worker reporting a protocol enrichment error
+        let protocol_response = ProtocolEnrichmentResponse::deserialize(deserializer)?;
+
+        match (protocol_response.context, protocol_response.error) {
+            (Some(context), None) => Ok(EnrichmentResponse {
+                id: protocol_response.id,
+                outcome: Ok(context),
+            }),
+            (None, Some(error)) => {
+                // errors here are happy path - user defined script is having a moment, but imporatntly, nothing is wrong with our system
+                Ok(EnrichmentResponse {
+                    id: protocol_response.id,
+                    outcome: Err(error),
+                })
+            }
+            (Some(_), Some(_)) => Err(serde::de::Error::custom(
+                "Context enricher protocol error: both context and error are set",
+            )),
+            (None, None) => Err(serde::de::Error::custom(
+                "Context enricher protocol error: response must contain either context or error",
+            )),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_deserializing_happy_path_enrichment_response_works() {
+        let json = r#"{"id": 1, "context": {"userId": "123"}, "error": null}"#;
+        let response: EnrichmentResponse = serde_json::from_str(json)
+            .expect("Failed to deserialize happy path enrichment response");
+
+        assert_eq!(response.id, 1);
+        assert!(response.outcome.is_ok());
+        assert_eq!(response.outcome.unwrap().user_id.as_deref(), Some("123"));
+    }
+
+    #[test]
+    fn test_deserializing_error_path_enrichment_response_works() {
+        let json = r#"{"id": 2, "context": null, "error": "Some error occurred"}"#;
+        let response: EnrichmentResponse = serde_json::from_str(json)
+            .expect("Failed to deserialize error path enrichment response");
+
+        assert_eq!(response.id, 2);
+        assert!(response.outcome.is_err());
+        assert_eq!(response.outcome.unwrap_err(), "Some error occurred");
+    }
+
+    #[test]
+    fn broken_protocol_message_errors() {
+        let json = r#"{"id": 3, "context": {"userId": "123"}, "error": "Some error occurred"}"#;
+        let both_fields_set_response: Result<EnrichmentResponse, _> = serde_json::from_str(json);
+
+        let json = r#"{"id": 4, "context": null, "error": null}"#;
+        let neither_field_set_response: Result<EnrichmentResponse, _> = serde_json::from_str(json);
+
+        assert!(both_fields_set_response.is_err());
+        assert!(neither_field_set_response.is_err());
+    }
+}
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -2,6 +2,7 @@ use serde::{Deserialize, Deserializer};
 use std::collections::HashMap;
 use unleash_types::client_features::Context;
 
+#[expect(dead_code)]
 pub(crate) struct EnrichmentRequest {
     pub(crate) id: i64,
     pub(crate) context: Context,
@@ -9,14 +10,16 @@ pub(crate) struct EnrichmentRequest {
 }
 
 pub(crate) struct EnrichmentResponse {
+    #[expect(dead_code)]
     pub(crate) id: i64,
+    #[expect(dead_code)]
     pub(crate) outcome: Result<Context, String>,
 }
 
-pub(crate) enum StateMessage {
-    Ready,
-    Error(String),
-    Closing,
+#[derive(Deserialize)]
+pub(crate) struct ReadyMessage {
+    #[serde(rename = "messageType")]
+    pub _message_type: String,
 }
 
 #[derive(Deserialize)]
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -99,4 +99,19 @@ mod tests {
         assert!(both_fields_set_response.is_err());
         assert!(neither_field_set_response.is_err());
     }
+
+    #[test]
+    fn malformed_enrichment_response_errors() {
+        let response: Result<EnrichmentResponse, _> = serde_json::from_str("not-json");
+
+        assert!(response.is_err());
+    }
+
+    #[test]
+    fn ready_message_deserializes() {
+        let ready: ReadyMessage = serde_json::from_str(r#"{"messageType":"ready"}"#)
+            .expect("Failed to deserialize ready message");
+
+        assert_eq!(ready._message_type, "ready");
+    }
 }
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -46,7 +46,7 @@ impl<'de> Deserialize<'de> for EnrichmentResponse {
                 outcome: Ok(context),
             }),
             (None, Some(error)) => {
-                // errors here are happy path - user defined script is having a moment, but imporatntly, nothing is wrong with our system
+                // errors here are happy path - user defined script is having a moment, but importantly, nothing is wrong with our system
                 Ok(EnrichmentResponse {
                     id: protocol_response.id,
                     outcome: Err(error),
--- a/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
@@ -243,16 +243,6 @@ mod tests {
     use tokio::io::AsyncWriteExt;
 
     fn node_is_available() -> bool {
-        // very stupid hack - right now CI doesn't have a Node runtime
-        // available and I don't want to add it. So we'll skip these on CI
-        // but I'm a banana who forgets things so I'm putting a time-bomb on this
-        // if this check is still here Aug 26 we fail so someone (me) needs to fix it
-
-        assert!(
-            std::time::SystemTime::now()
-                < std::time::SystemTime::UNIX_EPOCH + std::time::Duration::from_secs(1787695200)
-        );
-
         StdCommand::new(resolve_node_path())
             .arg("--version")
             .output()
--- a/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
@@ -268,6 +268,7 @@ mod tests {
         command
     }
 
+    #[cfg(unix)]
     #[tokio::test]
     async fn worker_sends_and_reads_happy_path_messages_to_child() {
         let command = fake_child_command(
--- a/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
@@ -214,3 +214,149 @@ fn resolve_node_path() -> PathBuf {
         .find(|path| path.is_file())
         .unwrap_or(node)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{
+        fs,
+        io::{BufRead, BufReader as StdBufReader, Read, Write},
+        path::Path,
+        process::{Child as StdChild, Command as StdCommand},
+    };
+    use tempfile::{NamedTempFile, TempPath};
+
+    fn node_is_available() -> bool {
+        StdCommand::new(resolve_node_path())
+            .arg("--version")
+            .output()
+            .is_ok_and(|output| output.status.success())
+    }
+
+    fn write_temp_enricher(source: &str) -> TempPath {
+        let file = NamedTempFile::with_suffix(".cjs").expect("failed to create temp enricher file");
+        let path = file.into_temp_path();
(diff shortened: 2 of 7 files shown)


crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    command
+        .arg(format!("--max-old-space-size={}", CHILD_MEMORY_CEILING_MB))
+        .arg("--eval")
+        .arg(WORKER_SCRIPT)
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .env_clear()
+        .kill_on_drop(true);
```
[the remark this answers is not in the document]
sighphyre: This is true but currently out of scope for this PR because there's nowhere to pipe in that script from and wiring it up would make it harder for humans to understand what's going on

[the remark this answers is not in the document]
sighphyre: Okay I changed my mind - we can just pass this as a parameter where we expect it

crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
```diff
@@ -0,0 +1,455 @@
...
+        );
+
+        StdCommand::new(resolve_node_path())
+            .arg("--version")
+            .output()
+            .is_ok_and(|output| output.status.success())
+    }
+
+    fn write_temp_enricher(source: &str) -> TempPath {
+        let file = NamedTempFile::with_suffix(".cjs").expect("failed to create temp enricher file");
```
daveleek, line 253: .expect("failed to create temp enricher file");

This takes some getting used to. "huh, why are they expecting this to fai-oooooh right."

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[derive(Deserialize)]
+pub(crate) struct ReadyMessage {
+    #[serde(rename = "messageType")]
+    pub _message_type: String,
+}
+
+#[derive(Deserialize)]
+struct ProtocolEnrichmentResponse {
```
ppicom, line 26: **thought:** When I hear "protocol", my mind goes to a structure that is generic enough to support a wide array of messages.

HTTP comes to mind, with headers and body. HTTP does not force a specific content in the body, it just tells you "Hey, here's where you put your things and I don't care if it's JSON, binary data or plain text as long as you set the corresponding header".

It's not clear to me if that is what we want to build, or if we will will just be exchanging one kind of messages: enrichment requests.

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub(crate) id: i64,
+    pub(crate) context: Option<Context>,
+    pub(crate) error: Option<String>,
```
ppicom, line 29: **question:** Going back to my HTTP train of thought, if we're building a protocol maybe we want to structure them in a way that errors can be communicated in the same way as normal errors? 

My rationale is that, in HTTP, any kind of response (success, server error, client error, etc) is returned using the same structure: headers + (optionally) a body.

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_deserializing_happy_path_enrichment_response_works() {
+        let json = r#"{"id": 1, "context": {"userId": "123"}, "error": null}"#;
```
ppicom, line 71: **question:** Have we considered an alternative to text-based comms? Maybe protobuff or flatbuffers?

It makes it hard to introduce breaking changes to the protocol in the future + the performance gainz of not parsing text.

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
ppicom: On that topic, evolvability also comes to mind. Is it worth adding `metadata` and/or versions to the protocol messages, so that we can make version based changes in the future?

crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
```diff
@@ -0,0 +1,455 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{fs, process::Command as StdCommand};
+    use tempfile::{NamedTempFile, TempPath};
+    use tokio::io::AsyncWriteExt;
+
+    fn node_is_available() -> bool {
+        // very stupid hack - right now CI doesn't have a Node runtime
```
ppicom: If it's stoopid bit it works, it's not stoopid.

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+// We use stdout for communication with the parent process, so anything that touches that will
+// break the protocol. Easy enough to just capture console output and redirect it to stderr instead.
+// The Rust worker runner expects us to do this and pipes it all out to logs
+function captureConsole() {
+    console.log = (...parts) => toStdErr("log", parts);
+    console.info = (...parts) => toStdErr("info", parts);
+    console.warn = (...parts) => toStdErr("warn", parts);
+    console.error = (...parts) => toStdErr("error", parts);
+    console.debug = (...parts) => toStdErr("debug", parts);
+}
```
ppicom, line 39: If we used sockets I think we would not need to have these redirections. I think it can be a bit confusing for users if we send INFO level texts to StdErr

ppicom (commented):
Hey! The code looks badass 🚀 I left a few thoughts around IPC and the protocol itself, I have no comments regarding Rust _per se_

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
sighphyre: Ah! I think you're using the word protocol wrong! A protocol is a set of rules that allows two systems to transmit information between them. The part of HTTP that's the protocol doesn't really have any specificity on what can be transmitted - rather it's the set of rules that must be followed for both systems to be mutually intelligible - how the data must be structured and the format of the messages that can be sent.

In this case, our protocol is the messaging format - which specifies the rules for the two systems to communicate; what each side must respect in terms of structure so the other can understand it. The message being transmitted is a Context, in HTTP that would be a String.

> On that topic, evolvability also comes to mind. Is it worth adding metadata and/or versions to the protocol messages, so that we can make version based changes in the future?

Nah, that's not necessary here. That's important when you don't own both sides of the pipe and one side can change out of lock step with the other. In this setup, the event that changes the protocol is a code and version change and a restart of the parent process and therefore the child process. A child can never be spawned by a parent on a different version

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
sighphyre: So I think this is the same as the previous comment. You're talking about REST here, not actually HTTP. HTTP errors aren't actually a thing that you'll likely ever see - they're handled on the server layer, probably logged and probably disbanded, they never actually become an HTTP response that can be parsed.

It's largely true of this as well. We're handling both of the error types you're talking about here - failures in the other system that are a valid message and also failures of the system to send a valid message. If you take a peek at this beast:

```
impl<'de> Deserialize<'de> for EnrichmentResponse
```
It's goal is to transform the protocol format to either a EnrichmentResponse message that the handler can understand - that's a valid message that both sides can and do understand - or a an Error type - in that case the message itself cannot be well understood.

The dance around the pair of options is because there's no really nice way to specify this in JSON (discriminated unions are horrible as a wire format, fight me.

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
sighphyre: Nah, didn't really spend much time thinking about that here. Those low level binary formats are a huge pain in the ass to use - we're already using FlatBuffers in part of Yggdrasil. For a spike that I need to feel out customer fit, I'm not planning on optimizing this right now, not at least until I understand where the performance is going in a real world scenario - it's very, very easy to optimize the wrong things. All I want from the performance conversation right now is the correct architecture - protocol format can be swapped pretty easily in a patch version

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: Users aren't ever going to know :sweat_smile:

The standard pipes aren't for other people's eyes here, We trap all the streams and do things with them. For all console out we just log it at info level on the Rust side.

Can argue that we should disambiguate the streams but importantly that is a protocol level decision not a transport level decision. If we pipe this over TCP sockets we'd still need to encode that disambiguation in the message format. And we can do that already by changing the message protocol - I've just chosen not to :sweat_smile: 

Open to talking about sockets here if you think there's some value but from experience dealing with low level sockets, they're an even bigger pain in the butt than std pipes and a lot slower. I'd rather avoid them unless there solving a killer problem

crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
```diff
@@ -0,0 +1,445 @@
...
+            }
+            Err(error) => {
+                let _ = event_tx
+                    .send(WorkerEvent::BrokenPipe(format!(
+                        "child stdout read failed: {error}"
+                    )))
+                    .await;
+                break;
+            }
+            Ok(None) => break,
```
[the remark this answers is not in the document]
sighphyre: Eh. Nah, this is a work in progress, it's addressed in later PRs, don't want to change this here

crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
sighphyre: Well it was stoopid and it worked - I forgot about this and I got caught out by the tests. Its fixed and removed now haha

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+            );
+        }
+    });
+}
+
+function main() {
+    try {
+        captureConsole();
+        const enricherScript = loadEnricherScript(process.argv);
+        if (typeof enricherScript !== "function") {
```
gastonfournier, line 101: I think a single function is all the user need... I was just thinking about it, and I believe it gives all the flexibility for them to decide inside the function... whether they're supporting one ore multiple workflows this should be enough. Maybe we need to consider more parameters but that'd be additive in the future

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
gastonfournier: I think stdout is a reasonable tradeoff for this stage. One small concern: patching `console.*` doesn't prevent `process.stdout.write()` from corrupting the protocol.

We could capture the real stdout writer before loading the enricher, use that exclusively for protocol messages, and then redirect/reject subsequent direct writes. Something along these lines:
```ts
const protocolWrite = process.stdout.write.bind(process.stdout);

function send(message) {
    protocolWrite(`${JSON.stringify(message)}\n`);
}

// this can be at script loading time not inside the main function
process.stdout.write = (...parts) => {
    process.stderr.write(
        `[process.stdout.write] ${parts.map(formatConsolePart).join(" ")}\n`,
    );
    return true;
};
```

Not a security boundary of course, since user code could still deliberately write to **fd 1 via fs** (TIL), but it would protect us from the most likely accidental protocol corruption without complicating the IPC setup.

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+            id: message.id,
+            error: error instanceof Error ? error.message : String(error),
+        });
+    }
+}
+
+function loadEnricherScript(args) {
+    const index = args.indexOf("--enricher-script");
+    if (index === -1 || args.length <= index + 1) {
+        throw new Error("enricher script path not provided");
```
gastonfournier, line 64: Will this put the process in a crash loop? It might be fine... cause it'd be a configuration/deployment issue

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+    console.log = (...parts) => toStdErr("log", parts);
+    console.info = (...parts) => toStdErr("info", parts);
+    console.warn = (...parts) => toStdErr("warn", parts);
+    console.error = (...parts) => toStdErr("error", parts);
+    console.debug = (...parts) => toStdErr("debug", parts);
+}
+
+async function handle(message, enrich) {
+    try {
+        const result = await enrich(message.context);
```
gastonfournier, line 51: I'd do something like this:
```suggestion
        const result = await enrich(message);
```

And then the enrich function becomes something like this:
```ts
function enrich({ context }) {
    // ...
}
```
this would make it easier to extend in the future, cause it's still one argument but it'd make it easier to send in other details (maybe headers or something)

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+        send({ id: message.id, context: result });
```
gastonfournier, line 52: Should we validate result is an object, not undefined, etc? Maybe as simple as:

```suggestion
        if (
            result === null ||
            typeof result !== "object" ||
            Array.isArray(result)
        ) {
            throw new Error(
                "context enricher must return an object",
            );
        }
        send({ id: message.id, context: result });
```

crates/enterprise/unleash-edge-context-enrichers/src/worker.rs
```diff
@@ -0,0 +1,445 @@
...
+        }
+    }
+}
+
+// The intent is to pipe user script's std out and std err to the process std err. That way when someone leaves a bunch
+// of console logs in the script, it doesn't break our IPC protocol. So this is just a polite way of making sure
+// that stream doesn't fill its own buffers and block while still giving the user a way to see what their script is doing.
+// We also really don't care about this from our system's perspective so it gets a lazy background task to just OMNOMNOMNOM the stream
```
gastonfournier, line 197: 🤣

gastonfournier (approved):
Overall looks good, just some potential improvements, but not blockers for an MVP

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+// The Rust worker runner expects us to do this and pipes it all out to logs
+function captureConsole() {
+    console.log = (...parts) => toStdErr("log", parts);
+    console.info = (...parts) => toStdErr("info", parts);
+    console.warn = (...parts) => toStdErr("warn", parts);
+    console.error = (...parts) => toStdErr("error", parts);
+    console.debug = (...parts) => toStdErr("debug", parts);
+}
+
+async function handle(message, enrich) {
```
gastonfournier, line 49: This might be AI slop (i.e. not verified by a human but seems to make sense), here it goes: When a request times out, `expire_pending_responses()` removes it from the map and therefore frees a slot.

The JavaScript operation hasn't been cancelled. So this can happen:

```
32 requests sent to Node
↓
all 32 call slow/hanging async code
↓
Rust deadlines expire
↓
pending_responses becomes empty
↓
Rust sends another 32
↓
the first 32 are STILL running in Node
↓
repeat
```

Eventually you can have hundreds/thousands of outstanding promises in Node while Rust believes there are only 32 in flight. Something like giving each handle function a timeout might help
```ts
function timeoutAfter(ms) {
    return new Promise((_, reject) => {
        setTimeout(
            () => reject(new Error(`Context enrichment timed out after ${ms}ms`)),
            ms,
        );
    });
}

async function handle(message, enrich) {
    try {
        const result = await Promise.race([
            enrich(message.context),
            timeoutAfter(1000),
        ]);

        send({ id: message.id, context: result });
    } catch (error) {
        send({
            id: message.id,
            error: error instanceof Error ? error.message : String(error),
        });
    }
}
```

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: It could, yeah. Right now it's not called anywhere so this is never executed. The way I'm intending this to be used is that this is done on the startup flow early in Edge's lifecycle, if the script isn't there we error out and tell the user they goofed. Doesn't stop someone deleting the script during execution which would trigger a crash loop + angry logs though

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: Yeah, fair comment, I'll patch.

Not really possible to make this completely bullet proof I think, and I'm not sure anyone is actually doing raw std out writes in scripts these days, but like you say, it's cheap to add.

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: Good suggestion. I'm also not passing in the headers here right now either and this suggestion would solve this out the box so I'll patch!

Edit: huh, actually I think your suggestion just works out the box - I'll accept!

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: Nah, don't think it's necessary. This gets sent to the Rust side and validation occurs in the read + deserialize, which would already catch this case

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,106 @@
...
+        try {
+            const message = JSON.parse(line);
+            // awaiting here would block our message handler and prevent inputs from the parent process being read.
+            // When the task to process the message is complete, we'll write it back over stdout so it can be fire and forget
+            void handle(message, enricherScript);
+        } catch (error) {
+            console.error(
+                `invalid JSON request: ${error instanceof Error ? error.message : String(error)}`,
+            );
+        }
```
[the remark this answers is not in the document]
sighphyre: No it doesn't, the job times out upstream

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -0,0 +1,116 @@
...
+    console.error = (...parts) => toStdErr("error", parts);
+    console.debug = (...parts) => toStdErr("debug", parts);
+}
+
+function captureProcessStdout() {
+    process.stdout.write = (...parts) => {
+        const message = parts.map(formatConsolePart).join(" ");
+        process.stderr.write(`[process.stdout.write] ${message}\n`);
+        return true;
+    };
```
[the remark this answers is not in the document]
sighphyre: I'm not doing this. This is a safety net fallback. I'm not adding more chaotic complexity here

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
sighphyre: It's... uh... the usual AI thing... wait wait wait... lemme quote Claude "you're half right but the half that's wrong is the load bearing seam" :laughing: 

Yeah, this is a problem, we can leave jobs lying around if they timeout. The bigger problem is this solution catches the very polite case - promises that are lingering. That is not the case I'm worried about, I'm more worried about sync work poisoning the event loop. The second issue with this is that the promise continues on in the background, it just never resolves in an observable way.

The existing plan was to detect poisoned instances and respin them but I haven't got there yet, needs a bit of careful thought because it's non trivial

gastonfournier: > sync work poisoning the event loop

I'm not too worried about this, and I'd blame the user if this happens :D but it's possible that they do some `while (true) {}` in a more subtle way... although for this spike I think it's an acceptable risk...

PR #1822, opened 2026-08-27
chore: worker pool for enrichment

sighphyre:
Adds a basic round robin worker pool for jobs so we can scale this out rather than being pinned to a single Node instance


crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,398 @@
...
+        let mut workers = vec![];
+
+        for worker_id in 0..worker_count.get() {
+            let worker =
+                NodeWorkerController::start(worker_id, &as_absolute_path(&script_path)?).await?;
+            workers.push(WorkerSlot {
+                jobs_inflight: AtomicUsize::new(0),
+                worker,
+            });
+        }
```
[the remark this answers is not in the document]
sighphyre: This comment is stupid.

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,398 @@
...
+        timeout: Duration,
+    ) -> Result<Context, EnricherError> {
+        let deadline = Instant::now() + timeout;
+
+        let _permit = timeout_at(deadline, self.inner.job_slots.acquire())
+            .await
+            .map_err(|_| EnricherError::IOError("Worker response timed out".to_string()))?
+            .map_err(|_| {
+                EnricherError::UnexpectedShutdown("Worker pool is shutting down".to_string())
+            })?;
```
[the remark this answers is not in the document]
sighphyre: Yes fair, fixed

crates/enterprise/unleash-edge-context-enrichers/src/command.rs
```diff
@@ -11,6 +11,7 @@ pub(crate) enum EnricherError {
     ProtocolError(String),
     ScriptError(String),
     IOError(String),
+    Timeout(String),
```
sighphyre, line 14: So we can bubble timeouts to the original caller - seems nicer to tell them timeouts happened rather than everything else in this list which is much much worse

crates/enterprise/unleash-edge-context-enrichers/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 mod child;
 mod command;
 mod driver;
+mod pool;
 mod protocol;
 mod worker;
+
+const MAX_SCHEDULED_JOBS: usize = 32;
```
sighphyre, line 8: Moved this out to a constant - the Worker defends itself by blocking more than this but the pool will also not hand over more jobs than it believes the worker can handle

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,394 @@
...
+                }
+            }
+        }
+
+        if let Some((selected_worker_index, _)) = idle_worker {
+            self.inner.worker_slots[selected_worker_index]
+                .jobs_inflight
+                .fetch_add(1, Ordering::SeqCst);
+
+            *next_worker_index = (selected_worker_index + 1) % total_workers;
```
gastonfournier, line 135: Do we need next_worker_index if every time we request a next worker we iterate on the whole set and pick the least busy one?

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,394 @@
...
+        // This is a bit subtle and I don't know that I've made the right choice here, only that one needed to be made
+        // This algorithm spreads load equally across workers by always rotating out the next worker index on selection.
+        // Under high load this doesn't matter because we'll pick the lowest load worker, but under low load this means
+        // that we spread the jobs around the whole pool rather than ping ponging between worker 1 and worker 2. Does this
+        // matter? Dunno, might be net negative for cache locality, might be net positive for GC pressure
+        let mut next_worker_index = self.inner.next_worker_index.lock().await;
+        let total_workers = self.inner.worker_slots.len();
+
+        let start = *next_worker_index;
+        let mut idle_worker = None;
```
gastonfournier: It's not precisely idle, might be the less busy one

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,394 @@
...
+            *next_worker_index = (selected_worker_index + 1) % total_workers;
+
+            Ok(selected_worker_index)
+        } else {
+            Err(EnricherError::Timeout("All workers are busy".to_string()))
+        }
+    }
+}
+
+fn as_absolute_path(path: &std::path::Path) -> Result<PathBuf, EnricherError> {
```
gastonfournier, line 144: Maybe this can be called outside the for loop and then clone the result (if clone is needed, otherwise shared reference might be ok)

gastonfournier (approved):
LGTM! Just minor comments

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
sighphyre: Fair point, I originally called this "lazy_worker" but maybe this rename didn't improve it much :sweat_smile:

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
sighphyre: Probably can, didn't sweat about it too much because cloning a 20B path isn't going to be significant when the very next thing is spawning a 10MB node instance.

Probably good to set a proper example though, I'll take another look

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
sighphyre: We don't really need it. It's solving a very particular problem (that might not need solving) - if we don't do this and the pool is mostly idle then all requests will always route to the same worker. This gives us round robin behavior on a mostly idle pool. Do we need that? Dunno, but it felt wrong to spawn a pool and then not use any of the other workers

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
```diff
@@ -0,0 +1,394 @@
...
+            )
+            .await
+    }
+
+    async fn select_next_worker(&self) -> Result<usize, EnricherError> {
+        // This is a bit subtle and I don't know that I've made the right choice here, only that one needed to be made
+        // This algorithm spreads load equally across workers by always rotating out the next worker index on selection.
+        // Under high load this doesn't matter because we'll pick the lowest load worker, but under low load this means
+        // that we spread the jobs around the whole pool rather than ping ponging between worker 1 and worker 2. Does this
+        // matter? Dunno, might be net negative for cache locality, might be net positive for GC pressure
```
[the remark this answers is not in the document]
sighphyre: Co-pilot, with all due respect, bugger off, you're one of the worst sources of difficult to maintain code, don't lecture me about comments

crates/enterprise/unleash-edge-context-enrichers/src/pool.rs
sighphyre: Yeah okay, I did what you suggested - no idea why I didn't do that originally, it reads much nicer, thanks!

PR #1826, opened 2026-08-31
chore: initial wire up of context enrichers - still fully disabled

sighphyre:
This wires up context enrichers into the the frontend API. It explicitly does not provide a path to actually use context enrichers, only the wire up.

I've chosen to use a null object pattern here to deal with the enterprise feature gating so that we don't have to leak the check for context enrichers into the surrounding code or deal with a thousand unused variables.


crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
```diff
@@ -0,0 +1,51 @@
+use std::{collections::HashMap, num::NonZeroU32, path::PathBuf, sync::Arc, time::Duration};
```
sighphyre, line 1: Just a very basic proxy that stands in front of the WorkerPool - primary purpose is to be able to expose a `disabled()` function that yields a Null Object. Doing that purely because I don't want to have to cfg gate this off in a million places for OSS, the Null Object here lets calling code not have to reason about that

crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
```diff
@@ -0,0 +1,51 @@
...
+        script_path: PathBuf,
+    ) -> Result<Self, EnricherError> {
+        Ok(Self {
+            worker_pool: Some(Arc::new(
+                WorkerPool::start(worker_count, script_path).await?,
+            )),
+        })
+    }
+
+    pub async fn enrich_or_original(
```
sighphyre: Do we hate the name? I kinda hate the name but I can't think of a better one

crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
```diff
@@ -0,0 +1,51 @@
...
+        context: Context,
+        headers: HashMap<String, String>,
+        timeout: Duration,
+    ) -> Context {
+        let Some(worker_pool) = self.worker_pool.as_ref() else {
+            return context;
+        };
+
+        match worker_pool
+            .request_enrichment(context.clone(), headers, timeout)
```
sighphyre, line 39: Dumb to clone here, needs a second pass to clean up the ownership. Different PR - this one's job is to isolate the wire up chaos, no more chaos for now

crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
```diff
@@ -0,0 +1,60 @@
...
+                match error {
+                    EnricherError::Timeout(message) => {
+                        info!(
+                            "Context enrichment timed out, falling back to original context: {message}"
+                        );
+                    }
+                    _ => {
+                        warn!(
+                            "Failed to enrich frontend context, falling back to original context: {error}"
+                        );
```
sighphyre, line 53: Timeouts are an annoyance, everything else is Bad

crates/enterprise/unleash-edge-context-enrichers/worker_script.js
```diff
@@ -48,7 +48,7 @@ function captureProcessStdout() {
 
 async function handle(message, enrich) {
     try {
-        const result = await enrich(message.context);
+        const result = await enrich(message.context, message.headers || {});
```
sighphyre, line 51: I goofed here. This should have been a destructure based on previous feedback. I think that got lost in one of the rebases. I'll put that back in a different PR - this is the smallest shim for now that lets me test properly

crates/oss/unleash-edge-appstate/src/context_enricher.rs
```diff
@@ -0,0 +1,29 @@
...
+            _: HashMap<String, String>,
+            _: Duration,
+        ) -> Context {
+            context
+        }
+    }
+}
+
+#[cfg(not(feature = "enterprise"))]
+pub use disabled::ContextEnricher;
```
sighphyre, line 29: Stuff in this file is the smallest possible Null Object for a context enricher - this is the case where we're OSS, the enterprise code doesn't exist for us to depend on. We could fix that by forcing a dependency on Context Enrichers even when we don't need it but that seems like a greater evil than this tiny shim

Neat thing about this is that this is the only place we need the cfg gates, rest of Edge doesn't know or care that Context Enrichers don't exist; we don't need a thousand compiler directives, we don't need to ignore unused warnings, we don't have a bunch of if statements

crates/oss/unleash-edge-appstate/src/lib.rs
```diff
@@ -12,6 +12,8 @@ use unleash_edge_types::metrics::instance_data::EdgeInstanceData;
 use unleash_edge_types::{EngineCache, TokenCache};
 use unleash_types::client_metrics::ConnectVia;
 
+pub use context_enricher::ContextEnricher;
```
sighphyre, line 15: We re-export the type here so callers don't need to depend on the Context Enricher crate directly when they actually don't care about that

crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -22,6 +24,8 @@ use unleash_types::client_features::Context;
 use unleash_types::client_metrics::{ClientApplication, ClientMetrics, ConnectVia};
 use unleash_types::frontend::{EvaluatedToggle, EvaluatedVariant, FrontendResult};
 
+const CONTEXT_ENRICHER_TIMEOUT: Duration = Duration::from_millis(100);
```
sighphyre, line 27: Yes, wants to be a CLI arg. Later PR when we fully expose this. It's still 100% disabled

crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -247,6 +263,33 @@ pub async fn frontend_post_feature(
...
+    headers
+        .iter()
+        .filter_map(|(name, value)| {
+            value
+                .to_str()
+                .ok()
+                .map(|value| (name.as_str().to_string(), value.to_string()))
+        })
+        .collect()
+}
```
sighphyre, line 287: This needs a second pass to deal with ownership. This adds a bunch of header clones to the paths that don't use Context Enrichers (probaby, sometimes Rustc is magic). Different PR

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -578,6 +578,7 @@ pub async fn build_edge_state(
         allow_list: args.http_allow_list.unwrap_or_default(),
         trust_proxy: args.trust_proxy,
         proxy_trusted_servers: args.proxy_trusted_servers,
+        context_enricher: ContextEnricher::disabled(),
```
sighphyre, line 581: Right now, path to set this - that's intentional. One PR so the messy wire up can be understood and tested. Another later to turn this on

crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -1096,4 +1143,117 @@ mod tests {
         let reader = BufReader::new(file);
         serde_json::from_reader(reader).unwrap()
     }
+
+    #[cfg(feature = "enterprise")]
+    mod context_enricher_tests {
```
sighphyre, line 1144: Adding a submodule here so we don't litter the parent test module with cfg enables

crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
gastonfournier: try_enrich? I don't hate the current name, but try_enrich hints that it may enrich or not and you'll get a response anyway...

crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -247,6 +263,33 @@ pub async fn frontend_post_feature(
     .map(Json)
 }
 
+async fn enrich_context_if_configured(
```
gastonfournier: Speaking of names, maybe `try_enrich` works well here as well

crates/oss/unleash-edge-frontend-api/src/frontend.rs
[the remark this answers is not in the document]
gastonfournier: I was actually considering this

gastonfournier (approved):
LGTM! nice work, "naming is hard" continues to be a thing

crates/oss/unleash-edge-frontend-api/src/frontend.rs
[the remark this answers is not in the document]
sighphyre: I mean yes but also that is the entire point of this whole project, Copilot. If we want people to extract Auth headers so they can unpack them and map them to user ids then this suggestion isn't going to help!

crates/oss/unleash-edge-frontend-api/src/frontend.rs
sighphyre: Yeah that's not bad actually, I'll do that. Definitely better than what I have in both places

crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
sighphyre: Ye! I like it!

crates/oss/unleash-edge-frontend-api/src/frontend.rs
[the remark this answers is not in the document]
sighphyre: I guess allow listing this is kinda reasonable at the CLI level maybe. I dunno. Drowning in poorly considered AI responses at the moment

crates/oss/unleash-edge-frontend-api/src/frontend.rs
```diff
@@ -247,6 +263,29 @@ pub async fn frontend_post_feature(
...
 }
 
+async fn try_enrich(app_state: &FrontendState, context: Context, headers: &HeaderMap) -> Context {
+    app_state
+        .context_enricher
+        .try_enrich(
+            context,
+            headers_for_context_enricher(headers),
+            CONTEXT_ENRICHER_TIMEOUT,
+        )
```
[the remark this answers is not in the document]
sighphyre: I KNOW, I POINTED THIS OUT IN A COMMENT ON THE PR. GO AWAY

PR #1827, opened 2026-08-31
chore: context enrichers no longer take deep clones of headers on eve…

sighphyre:
Adds a bit of song and dance so that Context Enrichers don't deep copy the entire header map unnecessarily - this is important for cases where Context Enrichers are not actually active. Before this, we took that perf hit on every request regardless

To make this possible, we include a thin carrier type that knows how to write itself to a JSON value. The public API to Context Enrichers accepts a borrowed HeaderMap and converts that to the internal type as soon as possible and then deals with that internally  

This doesn't address the same problem with the Context value - next piece of work

### DIFF
--- a/crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
@@ -1,8 +1,9 @@
-use std::{collections::HashMap, num::NonZeroU32, path::PathBuf, sync::Arc, time::Duration};
+use http::HeaderMap;
+use std::{num::NonZeroU32, path::PathBuf, sync::Arc, time::Duration};
 use tracing::{info, warn};
 use unleash_types::client_features::Context;
 
-use crate::{EnricherError, WorkerPool};
+use crate::{EnricherError, WorkerPool, serializable_header::SerializableHeaders};
 
 #[derive(Clone)]
 pub struct ContextEnricher {
@@ -28,7 +29,7 @@ impl ContextEnricher {
     pub async fn try_enrich(
         &self,
         context: Context,
-        headers: HashMap<String, String>,
+        headers: &HeaderMap,
         timeout: Duration,
     ) -> Context {
         let Some(worker_pool) = self.worker_pool.as_ref() else {
@@ -36,7 +37,7 @@ impl ContextEnricher {
         };
 
         match worker_pool
-            .request_enrichment(context.clone(), headers, timeout)
+            .request_enrichment(context.clone(), SerializableHeaders(headers), timeout)
             .await
         {
             Ok(enriched_context) => enriched_context,
--- a/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
@@ -1,12 +1,31 @@
 use serde::{Deserialize, Deserializer, Serialize};
-use std::collections::HashMap;
 use unleash_types::client_features::Context;
 
+use crate::serializable_header::SerializableHeaders;
+
 #[derive(Serialize)]
-pub(crate) struct EnrichmentRequest {
+pub(crate) struct EnrichmentRequest<'a> {
     pub(crate) id: u64,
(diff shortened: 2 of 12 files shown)


crates/enterprise/unleash-edge-context-enrichers/src/context_enricher.rs
```diff
@@ -28,15 +29,15 @@ impl ContextEnricher {
...
+        headers: &HeaderMap,
         timeout: Duration,
     ) -> Context {
         let Some(worker_pool) = self.worker_pool.as_ref() else {
             return context;
         };
 
         match worker_pool
-            .request_enrichment(context.clone(), headers, timeout)
+            .request_enrichment(context.clone(), SerializableHeaders(headers), timeout)
```
sighphyre, line 40: Public surface takes the HTTP type and converts that to the type we actually care about for processing

crates/enterprise/unleash-edge-context-enrichers/src/protocol.rs
```diff
@@ -1,12 +1,31 @@
...
+}
+
+impl TryFrom<EnrichmentRequest<'_>> for SerializedEnrichmentRequest {
+    type Error = serde_json::Error;
+
+    fn try_from(request: EnrichmentRequest<'_>) -> Result<Self, Self::Error> {
+        let mut serialized = serde_json::to_vec(&request)?;
+        serialized.push(b'\n');
+        Ok(Self(serialized))
+    }
```
sighphyre, line 28: Reasonable to ask wtf I'm doing here. 

We now deal in a borrowed type. Previously we passed our owned type over a channel. Can't do that with a borrowed type so we need to serialize before we send it over the channel. However, that means dealing in raw bytes and not concrete types. I don't trust anyone to do that, especially not myself. So instead, we introduce a SerializedEnrichmentRequest type. Internally that's just a wrapper around a byte array. But - this is the clever part - you can't construct that manually, you can only create it from an EnrichmentRequest using the try_from implementation, which guarantees the backing data cannot be created incorrectly and forces the WorkerCommand to depend on that type. 

What does this buy you? The public surface and safety of strong, static typing, while being able to yeet around byte arrays like we're partying in a Python VM

Honestly I'm quite proud of this 😅

crates/enterprise/unleash-edge-context-enrichers/src/serializable_header.rs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+
+        for (name, value) in self.0 {
+            let value = value.to_str().map_err(S::Error::custom)?;
+
+            map.serialize_entry(name.as_str(), value)?;
+        }
+
+        map.end()
```
sighphyre: This is dumb, as in headers aren't necessarily a map behind the scenes, they can be a multimap (thank you HTTP spec, very nice). But we're not making this worse right now so I'm leaving it dumb

crates/enterprise/unleash-edge-context-enrichers/src/serializable_header.rs
```diff
@@ -0,0 +1,24 @@
...
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+
+        for (name, value) in self.0 {
+            let value = value.to_str().map_err(S::Error::custom)?;
+
+            map.serialize_entry(name.as_str(), value)?;
+        }
```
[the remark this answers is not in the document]
sighphyre: Fair, reasonable comment, accepted

crates/enterprise/unleash-edge-context-enrichers/src/serializable_header.rs
```diff
@@ -0,0 +1,26 @@
+use http::HeaderMap;
+use serde::{
+    Serialize, Serializer,
+    ser::{Error as _, SerializeMap},
+};
```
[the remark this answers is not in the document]
sighphyre: Yes, Copilot, you did that

gastonfournier (approved):
It looks `[115, 97, 110, 101]` (at least in utf8)

PR #1833, opened 2026-09-01
feat: expose context enrichers to public API

sighphyre:
Allows context enrichers to be configured at startup so they're technically usable at this point. However I've turned off their visibility from the help text so the only way anyone knows they're here is reading the source. Standard flow for dark releases in Edge really

There's a little more cfg feature gating than I would like here, however, this allows us to make sure downstream code never has to think about the differences between enterprise + context enrichers, enterprise - context enrichers and oss. And it's also impossible to get wrong because the invalid state simply doesn't exist at compile time (thank you rustc, I do appreciate you)

### DIFF
--- a/crates/oss/unleash-edge-cli/src/lib.rs
+++ b/crates/oss/unleash-edge-cli/src/lib.rs
@@ -7,6 +7,8 @@ use ipnet::IpNet;
 use reqwest::Client;
 use std::fmt::{Display, Formatter};
 use std::net::IpAddr;
+#[cfg(feature = "enterprise")]
+use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::time::Duration;
@@ -265,8 +267,22 @@ pub struct EdgeArgs {
     #[clap(long, env, hide = true)]
     pub ec2_instance_id: Option<String>,
 
+    #[cfg(feature = "enterprise")]
+    #[clap(flatten)]
+    pub context_enricher: ContextEnricherArgs,
+
     #[clap(flatten)]
     pub hmac_config: HmacConfig,
+}
+
+#[cfg(feature = "enterprise")]
+#[derive(Args, Debug, Clone, Default)]
+pub struct ContextEnricherArgs {
+    #[clap(long, env, hide = true)]
+    pub context_enricher_script: Option<PathBuf>,
+
+    #[clap(long, env, hide = true, requires = "context_enricher_script")]
+    pub context_enricher_workers: Option<NonZeroU32>,
 }
 
 pub fn string_to_header_tuple(s: &str) -> Result<(String, String), String> {
@@ -789,7 +805,7 @@ impl HttpServerArgs {
 mod tests {
     use super::{CliArgs, EdgeMode, NetworkAddr};
     use axum::http;
-    use clap::Parser;
+    use clap::{CommandFactory, Parser};
     use ipnet::IpNet;
     use std::net::IpAddr;
     use std::str::FromStr;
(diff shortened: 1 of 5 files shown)


crates/oss/unleash-edge-cli/src/lib.rs
```diff
@@ -7,6 +7,8 @@ use ipnet::IpNet;
 use reqwest::Client;
 use std::fmt::{Display, Formatter};
 use std::net::IpAddr;
+#[cfg(feature = "enterprise")]
+use std::num::NonZeroU32;
```
sighphyre, line 11: Only place we're using this right now. Honestly we should make use of this more so I'm fine with the feature gate for now - I intend to use widen the use here where it makes sense

crates/oss/unleash-edge-cli/src/lib.rs
```diff
@@ -265,10 +267,24 @@ pub struct EdgeArgs {
...
 }
 
+#[cfg(feature = "enterprise")]
+#[derive(Args, Debug, Clone, Default)]
+pub struct ContextEnricherArgs {
+    #[clap(long, env, hide = true)]
+    pub context_enricher_script: Option<PathBuf>,
+
+    #[clap(long, env, hide = true, requires = "context_enricher_script")]
+    pub context_enricher_workers: Option<NonZeroU32>,
```
sighphyre, line 285: The requires here means the entry point is setting the script. Nothing with work without that and everything else is likely going to be optional anyway

What this does mean is that the default value here needs to be set elsewhere. Think that tradeoff is worth it, it exchanges user facing clarity for a little bit of indirection in the code

crates/oss/unleash-edge-cli/src/lib.rs
```diff
@@ -1318,6 +1334,66 @@ mod tests {
         assert!(args.is_ok());
     }
 
+    #[cfg(feature = "enterprise")]
+    #[test]
+    pub fn can_parse_hidden_context_enricher_args() {
```
sighphyre, line 1339: Really not sure I care about these tests but everything else here has tests (why?) so it felt okay to just run with it

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -422,6 +424,39 @@ pub struct EdgeStateArgs {
...
+#[derive(Debug, Clone, Default)]
+pub enum ContextEnricherConfig {
+    #[default]
+    Disabled,
+    #[cfg(feature = "enterprise")]
+    Enabled {
+        script_path: PathBuf,
+        worker_count: NonZeroU32,
+    },
+}
```
sighphyre, line 439: Enabled state simply doesn't exist without the enterprise feature. Bit more cfg gating than I would like but making impossible states impossible on a compile level is just too nice to not tap into

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -590,6 +626,21 @@ pub async fn build_edge_state(
...
+) -> Result<ContextEnricher, EdgeError> {
+    match config {
+        ContextEnricherConfig::Disabled => Ok(ContextEnricher::disabled()),
+        #[cfg(feature = "enterprise")]
+        ContextEnricherConfig::Enabled {
+            script_path,
+            worker_count,
+        } => ContextEnricher::start(worker_count, script_path)
+            .await
+            .map_err(|error| EdgeError::ContextEnricherError(error.to_string())),
```
sighphyre, line 640: Gotta cfg gate this because the enum variant doesn't exist in OSS. Worth it I think to prevent people (me) from goofing

crates/oss/unleash-edge-cli/src/lib.rs
daveleek: So instead of having 0 workers, there's either None or 1+?

crates/oss/unleash-edge/src/edge_builder.rs
daveleek: :O

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -422,6 +424,39 @@ pub struct EdgeStateArgs {
...
+                        .context_enricher
+                        .context_enricher_workers
+                        .unwrap_or_else(|| NonZeroU32::new(1).unwrap()),
+                };
+            }
+        }
+
+        #[cfg(not(feature = "enterprise"))]
+        let _ = edge_args;
+        Self::Disabled
```
daveleek, line 458: So these are kind of like a more effective compiler directive/variable? Either picks one or the other

daveleek (approved):
Cool! Looks good

crates/oss/unleash-edge-cli/src/lib.rs
sighphyre: Exactly! It's a non zero numeric enforced at compile time, there's no way of inputting a 0 here (which would bork the pool). Sometimes Rust is very nice

crates/oss/unleash-edge/src/edge_builder.rs
sighphyre: Yep, you got it

PR #1837, opened 2026-09-02
chore: propagate env vars to context enrichers

sighphyre:
Propagates the env vars from Edge to the Node workers for Context Enrichers. For any meaningfully interesting context enricher, that needs to happen probably (currently blocked myself when making a JWKS validating example).

For hosted this would be a problem because of security concerns. But we're not hosting this. For self hosted - it's your script, your Edge instance and your env vars, I don't think it's an issue

### DIFF
--- a/crates/enterprise/unleash-edge-context-enrichers/src/child.rs
+++ b/crates/enterprise/unleash-edge-context-enrichers/src/child.rs
@@ -1,8 +1,4 @@
-use std::{
-    path::{Path, PathBuf},
-    process::Stdio,
-    time::Duration,
-};
+use std::{path::Path, process::Stdio, time::Duration};
 use tokio::{
     io::{AsyncBufReadExt, BufReader, Lines},
     process::{Child, ChildStderr, ChildStdin, ChildStdout, Command},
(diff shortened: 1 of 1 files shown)


crates/enterprise/unleash-edge-context-enrichers/src/child.rs
```diff
@@ -198,28 +194,10 @@ fn node_worker_command(enricher_script: &Path) -> Command {
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
-        .env_clear()
```
daveleek, line 201: All the things we care about in Edge internally are UNLEASH-prefixed?

gastonfournier (approved):
It's a spike to validate context enrichers. This gives the user flexibility to tweak NODE_OPTIONS if they want to... we can harden security later, but now we're trying to validate if this works

crates/enterprise/unleash-edge-context-enrichers/src/child.rs
```diff
@@ -198,28 +194,10 @@ fn node_worker_command(enricher_script: &Path) -> Command {
...
         .kill_on_drop(true);
```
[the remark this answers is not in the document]
sighphyre: Yes that's exactly the point. Hush you

crates/enterprise/unleash-edge-context-enrichers/src/child.rs
sighphyre: Nope, why?

daveleek: as mentioned in the slack thread I'm just raising the possibility that these env vars from edge could be confusing the node runtime

PR #1841, opened 2026-09-03
docs: add a context enricher example that details how to use libraries

sighphyre:
Last of the important Context Enricher examples. This one demos how to assemble a scriptlet that consumes external libraries and is rolled up into a single artifact.

The enricher itself is a bit dumb - it's just leveraging the semver library to check if this is stable or non stable release and enriching the context with that field. Probably don't want to do that for prod but I can't think of a simple example that actually needs an external library that makes more sense than this. Think it doesn't matter - the point is the build itself

### DIFF
--- a/examples/context-enrichers/semver-build/README.md
+++ b/examples/context-enrichers/semver-build/README.md
@@ -1,0 +1,69 @@
+# Semver Build Context Enricher
+
+This example shows how to use an npm library in a context enricher by bundling
+source code into a single CommonJS artifact and using that artifact as the
+enricher script.
+
+The enricher imports `semver`, reads the `X-Application-Version` request
+header, and sets `context.properties.releaseChannel` to:
+
+- `experimental` for prerelease versions such as `1.2.3-beta.1`
+- `stable` for normal versions such as `1.2.3`
+
+The enrichment logic is intentionally small. The point of the example is the
+build setup.
+
+## Build
+
+Install dependencies and build the bundled enricher:
+
+```sh
+cd examples/context-enrichers/semver-build
+npm install
+npm run build
+```
+
+This writes `dist/enricher.cjs`. The compose file mounts that built artifact
+into the Edge container and points `CONTEXT_ENRICHER_SCRIPT` at it.
+
+## Run
+
+This example does not start Unleash for you. Point it at either a local Unleash
+instance or a hosted Unleash instance. This requires an enterprise Unleash
+version with an Edge license.
+
+For local Unleash, use `host.docker.internal`, not `localhost`. From inside the
+Edge container, `localhost` points back at the Edge container, not at the host
+running Unleash.
+
+```sh
+UPSTREAM_URL=http://host.docker.internal:4242 \
+TOKENS=*:development.8400d1451101f5d05c0817c03ae5286371369311121bb114f4e268f3 \
+docker compose -f examples/context-enrichers/semver-build/compose.yml up --build
+```
+
+For hosted Unleash:
+
+```sh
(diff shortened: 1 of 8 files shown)


gastonfournier (approved):
Exciting stuff, just needs a web server 🤣 (if you reading this, are even thinking about it... no, it's not a good idea!)

examples/context-enrichers/semver-build/src/enricher.cjs
```diff
@@ -0,0 +1,18 @@
+const semver = require("semver");
+
+const versionHeader = (process.env.VERSION_HEADER || "x-application-version").toLowerCase();
+
+module.exports = async function enrich(context, headers) {
+  const version = semver.parse(headers[versionHeader]);
```
daveleek, line 6: Nice one, didn't recall that we're passing in headers!

examples/context-enrichers/semver-build/README.md
```diff
@@ -0,0 +1,69 @@
...
+instance or a hosted Unleash instance. This requires an enterprise Unleash
+version with an Edge license.
+
+For local Unleash, use `host.docker.internal`, not `localhost`. From inside the
+Edge container, `localhost` points back at the Edge container, not at the host
+running Unleash.
+
+```sh
+UPSTREAM_URL=http://host.docker.internal:4242 \
+TOKENS=*:development.8400d1451101f5d05c0817c03ae5286371369311121bb114f4e268f3 \
```
daveleek, line 41: Oooooh what's this

examples/context-enrichers/semver-build/README.md
```diff
@@ -0,0 +1,69 @@
...
+docker compose -f examples/context-enrichers/semver-build/compose.yml up --build
+```
+
+For hosted Unleash:
+
+```sh
+UPSTREAM_URL=https://app.unleash-hosted.com/<your-instance> \
+TOKENS=<your-client-api-token> \
```
daveleek, line 49: I'm suspecting the token above was supposed to look like this, even if that token is a "dud" as far as public instances are concerned (guessing it belongs to a local setup)

examples/context-enrichers/semver-build/compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      REDIS_URL: redis://redis:6379
+      CONTEXT_ENRICHER_SCRIPT: /app/context-enrichers/enricher.cjs
+      CONTEXT_ENRICHER_WORKERS: ${CONTEXT_ENRICHER_WORKERS:-1}
+      VERSION_HEADER: x-application-version
+    ports:
+      - "3063:3063"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    volumes:
+      - ./dist/enricher.cjs:/app/context-enrichers/enricher.cjs:ro
```
daveleek, line 18: what does :ro do here?

daveleek (approved):
I think this looks good, couple of questions that you can ignore after you've seen them

examples/context-enrichers/semver-build/compose.yml
daveleek: Cool, it sets it as read-only. TIL

sighphyre: It's a read-only file mount

examples/context-enrichers/semver-build/README.md
sighphyre: Local token, I've done this in all the examples, so I'm not changing in here. Can open another PR to patch this though. Not sure it's a big deal

PR #1875, opened 2026-09-21
fix: make sure we allow startup even if Unleash can't be reached

chriswk:
Currently, if HMAC is used and Unleash/Upstream is down, Edge will just crash with a message saying that we require at least one token. But we might still have validated tokens in our backup. This PR makes sure that we can recover from this case, where we have valid tokens in backup, but Unleash is not responding.

If we don't have tokens in backup and no HMAC settings (clientkey) is given nor a token at the CLI we'll still crash, because we have no reasonable expectation of being able to spin up.

Future PR could add a background task in HMAC mode, checking tokens tagged for refresh against environments requested at startup to make sure that if tokens are invalidated, Edge is able to make a new HMAC request for new tokens for the requested environments.

This is part 2 of solving https://linear.app/unleash/issue/EG-4348/edge-can-resurrect-stale-backend-tokens-from-persistence-after. It doesn't actually solve that we recover stale tokens even if we already had fresh tokens. It does however add to our stability story. We can now successfully startup with no Unleash even if we only have a backup.


crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -71,6 +60,7 @@ async fn send_instance_data(
     instance_data_sender: &Arc<InstanceDataSending>,
     our_instance_data: &Arc<EdgeInstanceData>,
     downstream_instance_data: &Arc<RwLock<Vec<EdgeInstanceData>>>,
+    token: &str,
 ) -> Result<(), InstanceDataSendError> {
     match instance_data_sender.as_ref() {
         InstanceDataSending::SendNothing => {
```
daveleek, line 66: Has this been orphaned? I don't see it used anywhere

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -227,11 +227,12 @@ pub async fn build_edge(
...
-        return Err(EdgeError::NoTokens(
-            "No tokens provided. Tokens must be specified".into(),
-        ));
-    }
+    /*    if tokens.is_empty() {
+            return Err(EdgeError::NoTokens(
+                "No tokens provided. Tokens must be specified".into(),
+            ));
+        }
+    */
```
ppicom: Have we considered deleting the commented out code? Is it here in case a quick revert is required?

crates/oss/unleash-edge/src/edge_builder.rs
```diff
@@ -227,11 +227,12 @@ pub async fn build_edge(
         features_refresh_interval,
     }: EdgeBuilderArgs,
 ) -> EdgeResult<EdgeInfo> {
-    if tokens.is_empty() {
-        return Err(EdgeError::NoTokens(
-            "No tokens provided. Tokens must be specified".into(),
-        ));
-    }
+    /*    if tokens.is_empty() {
```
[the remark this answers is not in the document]
chriswk: This one is actually a fair callout. Though, since the backed up tokens most likely have been filtered earlier at startup of whatever session performed the persistence, it feels like it's invalid here

crates/oss/unleash-edge/src/lib.rs
```diff
@@ -145,7 +145,8 @@ pub async fn configure_server(args: CliArgs) -> EdgeResult<(Router, Vec<Backgrou
                 edge_args.tokens.clone(),
                 edge_args.hmac_config.clone(),
             )
-            .await?;
+            .await
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
chriswk: Oh, nice catch, yeah, we should probably use token_cache for resolving license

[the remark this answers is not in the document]
chriswk: Though, the license might already have been hydrated from persistence. I'll try to write tests here confirming both paths work.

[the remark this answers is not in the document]
irstavr: yeah sounds like the bot is not wrong here 🤔 

maybe worths to log any errors here too? and fallback to the configured tokens as we do in LN 145 `edge_args.tokens.clone(),` 
as if any failure here, we fall back to empty vec instead of the static toekns the user has configured

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -155,30 +153,45 @@ pub fn create_send_instance_data_task(
...
+            } else {
+                debug!("No valid token found, skipping instance data send");
             }
         }
     })
 }
+
+fn find_first_valid_token(token_cache: &TokenCache) -> Option<String> {
+    token_cache.iter().find_map(|t| {
+        if t.status == TokenValidationStatus::Validated && t.token_type == Some(TokenType::Backend)
```
ppicom: Is there a reason not to take `TokenValidationStatus::Trusted` into consideration here? Probably a noob question, it just sounds like `Trusted` is a subset of `Validated` conceptually.

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -123,29 +113,37 @@ pub fn create_once_off_send_instance_data(
...
-    })
+            if let Err(err) = result {
+                warn!("Failed to send last set of instance data during graceful exit: {err:?}");
+            }
+        })
+    } else {
+        Box::pin(async move {
+            warn!("No valid token found, skipping sending instance data during graceful exit");
+        })
+    }
```
irstavr: nit:  moving the lookup inside the move() so we remove the duplicated box::pin branches 

```suggestion
        Box::pin(async move {
            let Some(our_token) = find_first_valid_token(&token_cache) {

}
			if let Err(err) =  send_instance_data(
                &instance_data_sender,
                &our_instance_data,
                &downstream_instance_data,
                &our_token,
            )
            .await {
                warn!("Failed to send last set of instance data during graceful exit: {err:?}");
            }
        })
    
```

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -155,30 +153,45 @@ pub fn create_send_instance_data_task(
...
+                }
+            } else {
+                debug!("No valid token found, skipping instance data send");
             }
         }
     })
 }
+
+fn find_first_valid_token(token_cache: &TokenCache) -> Option<String> {
+    token_cache.iter().find_map(|t| {
```
ppicom: We've gone from deterministically iterating a list, to iterating a map where the order of the elements of its iterator is not guaranteed. Probably OK, just raising my hand in case it was not intended. (this bit me in the ass in the past)

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -123,29 +113,37 @@ pub fn create_once_off_send_instance_data(
     instance_data_sender: Arc<InstanceDataSending>,
     our_instance_data: Arc<EdgeInstanceData>,
     downstream_instance_data: Arc<RwLock<Vec<EdgeInstanceData>>>,
+    token_cache: Arc<TokenCache>,
 ) -> BackgroundTask {
     let instance_data_sender = instance_data_sender.clone();
     let our_instance_data = our_instance_data.clone();
     let downstream_instance_data = downstream_instance_data.clone();
+    if let Some(our_token) = find_first_valid_token(&token_cache) {
```
ppicom: I think this is checked at the time the task is created, but Unleash could come back to life at a later date and we'd still print "no valid token found" at the time of shutdown.

Maybe the check needs to be moved inside the task itself?

crates/oss/unleash-edge-http-client/src/instance_data.rs
irstavr: 👀 actually this might save us also from a scenario when none token was validated at boot or that it got captured at boot but got invalidated later
as the find_first_valid_token() lookup atm runs when task is built ->token at boot time

crates/oss/unleash-edge/src/lib.rs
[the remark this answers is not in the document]
chriswk: Yeah, I'll have a look, currently working down the path of using token_cache, since the resolve work is done.

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -155,30 +153,45 @@ pub fn create_send_instance_data_task(
...
+                            warn!(message);
+                            errors += 1;
+                        }
+                        InstanceDataSendError::Unexpected(message) => {
+                            warn!(message);
+                        }
+                    },
+                }
+            } else {
+                debug!("No valid token found, skipping instance data send");
```
irstavr, line 180: how ofter will this be running? 🤔 
if it's per some seconds, won't be that visible in prod after all ?! warn instead?

crates/oss/unleash-edge/src/lib.rs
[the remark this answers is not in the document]
ppicom: It also treats all errors the same, maybe distinguish "upstream unreachable" (retry-worthy) from "signature rejected"?

crates/oss/unleash-edge-http-client/src/instance_data.rs
irstavr: nit: maybe a `matches!`reads a bit more easily than `find_map` with `if-else`

irstavr: also how about a test on this one? so it skips the Invalid frontend tokens and picks the validated backend ones

ppicom:
I asked this on slack, stamping it here for traceability purposes (even though it was answered in the description of the PR)

> Have we considered running build_tokens in a loop as well? That would get us fresh tokens eventually, innit? :thinking:

crates/oss/unleash-edge/src/lib.rs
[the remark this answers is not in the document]
chriswk: Yeah, updated and added a test that verifies that we keep running even if we only have tokens/license from persistence

crates/oss/unleash-edge-http-client/src/instance_data.rs
```diff
@@ -84,7 +74,7 @@ async fn send_instance_data(
             );
             let status = instance_data_sender
                 .unleash_client
-                .post_edge_observability_data(observed_data, &instance_data_sender.token)
+                .post_edge_observability_data(observed_data, token)
```
chriswk, line 77: isn't it used right here?

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: You mean this enum value? Looks like it. Should be cleaned. Different PR though

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: I'll quote Simon
"Don't you come here with your facts"

I believe you're right, we should move getting the token into the task. Thanks 👍

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: Yeah, Pere commented the same. Two good developers catching the same mistake. Updated :)

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: So, Trusted should never be validated, it's a thing only for Edge, to work similar to how our proxy worked, where the user was able to spin it up with pre-approved tokens that was supposed to be allowed to access the proxy. It does not work against Unleash, in fact Unleash does not know of it at all, so it's important we don't use it upstream

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: Yeah, it's a fair point, I think here we don't really care, if it's a validated token, we don't use the token for anything on the upstream side when posting instance data / metrics. Edge has already collated the data for multiple envs/tokens.

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: I think we send instance_data once per minute. and I don't think it's an important detail. You've probably noticed no valid token in another manner, no data for your SDKs connected to Edge :)

crates/oss/unleash-edge-http-client/src/instance_data.rs
chriswk: yeah. agreed, will add

chriswk: TIL matches! macro :)

chriswk: > nit: maybe a matches! reads a bit more easily than find_map with if-else

hmm, matches! still only returns a bool right? What were you thinking here? changing to filters and using the matches! macro and then a first?

chriswk: > also how about a test on this one? so it skips the Invalid frontend tokens and picks the validated backend ones

Yes, adding tests.

ppicom (approved):
we gathered to go over the PRs and we are happy with the way it looks

some comments from the robot are worth going over one more time, but the changes make sense!

